PR #13, opened 2018-05-21
Registries

littlebeer2100:
add registries


littlebeer2100:
add registries auth function

PR #20, opened 2018-05-24
Add the API of listing all PVC of the specific storageclass

wnxn:
uri: /api/v1alpha/storage/storageclasses/{name}/persistentvolumeclaims

### DIFF
--- a/pkg/models/storageutil.go
+++ b/pkg/models/storageutil.go
@@ -1,0 +1,64 @@
+package models
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/golang/glog"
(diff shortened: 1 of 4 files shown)


pkg/models/storageutil.go
```diff
@@ -0,0 +1,64 @@
...
+	return res
+}
+
+func (m jsonRawMessage) ToString() string {
+	res := strings.Replace(string(m[:]), "\"", "", -1)
+	return res
+}
+
+func GetApiserver(uri string) string {
+	url := "http://139.198.6.45:8001" + uri
```
richardxz: hard code?

PR #35, opened 2018-05-31
add response struct

wansir:


pkg/constants/common.go
```diff
@@ -16,13 +16,15 @@ limitations under the License.
 
 package constants
 
-type ResultMessage struct {
-
-	Data	interface{} `json:"data"`
+type MessageResponse struct {
+	Message interface{} `json:"message"`
```
wnxn: Is the difference between ResultMessage and MessageResponse just the key of response data?

PR #54, opened 2018-06-06
Com

littlebeer2100:
add istio and kubesphere components scanner

### DIFF
--- a/pkg/models/components.go
+++ b/pkg/models/components.go
@@ -28,6 +28,7 @@ import (
 
 const KUBESYSTEM = "kube-system"
 const OPENPITRIX = "openpitrix-system"
+const ISTIO  = "istio-system"
 
 type Components struct {
 	Name         string      `json:"name"`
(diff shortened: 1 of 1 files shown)


pkg/models/components.go
```diff
@@ -228,9 +230,9 @@ func GetComponents() (result []Components, err error) {
 
 	}
 
-	templates = []string{"kube-dns", "heapster", "monitoring-influxdb", "iam", "openpitrix"}
+	templates = []string{"kube-dns", "heapster", "monitoring-influxdb", "iam", "openpitrix","istio","kubesphere"}
 
-	namespaces := []string{KUBESYSTEM, OPENPITRIX}
+	namespaces := []string{KUBESYSTEM, OPENPITRIX,ISTIO,KUBESPHERE}
```
zryfish: define as a constant variable in constants.go would be better

pkg/models/components.go
```diff
@@ -228,9 +230,9 @@ func GetComponents() (result []Components, err error) {
 
 	}
 
-	templates = []string{"kube-dns", "heapster", "monitoring-influxdb", "iam", "openpitrix"}
+	templates = []string{"kube-dns", "heapster", "monitoring-influxdb", "iam", "openpitrix","istio","kubesphere"}
```
zryfish: same here

zryfish: Also need to format code before check in

pkg/models/components.go
littlebeer2100: constant will be handle together,for now there are in self model

PR #56, opened 2018-06-06
fix router bugs

zryfish:
Fix router bugs


pkg/apis/v1alpha/routes/routes_handler.go
```diff
@@ -64,6 +63,11 @@ func Register(ws *restful.WebService) {
 		Produces(restful.MIME_JSON))
 }
 
+type Router struct {
```
littlebeer2100: Router struct show be define in dir models,not in handler

zryfish: This only used in handler, and only for parameter parsing purpose

PR #71, opened 2018-06-09
add registries create update delete query function

littlebeer2100:


pkg/models/nodes.go
```diff
@@ -216,3 +217,9 @@ func getNodeFileSystemStatus(node *v1.Node) (string, string, string) {
 	}
 	return "", "", ""
 }
+
+func DrainNode(nodename string) {
```
calvinyv: As no logic is added in this PR for nodes model, I think you should undo this change

littlebeer2100: ok，i will remove this funcion

PR #180, opened 2018-09-27
add hpa api

richardxz:


pkg/apis/v1alpha/hpa/hpa.go
```diff
@@ -0,0 +1,57 @@
...
+func getHpa(req *restful.Request, resp *restful.Response) {
+	hpa := req.PathParameter("horizontalpodautoscaler")
+	namespace := req.PathParameter("namespace")
+	client := client.NewK8sClient()
+
+	res, err := client.AutoscalingV1().HorizontalPodAutoscalers(namespace).Get(hpa, metaV1.GetOptions{})
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			resp.WriteHeaderAndEntity(http.StatusOK, nil)
```
wnxn: The response status might be set as `not found` instead of `ok`.

richardxz: this is frontend's need,  if return 404, the web page will not work properly

pkg/apis/v1alpha/hpa/hpa.go
```diff
@@ -0,0 +1,57 @@
...
+		} else {
+			resp.WriteHeaderAndEntity(http.StatusInternalServerError, nil)
```
wnxn: Shall we return error message ?

pkg/apis/v1alpha/hpa/hpa.go
wnxn: Ok, I understand what you mean.

pkg/apis/v1alpha/hpa/hpa.go
richardxz: don't need,  nobody care this

PR #193, opened 2018-10-16
update job's "rerun" function

richardxz:


pkg/models/controllers/jobs.go
```diff
@@ -229,46 +234,80 @@ func deleteJob(namespace, job string) error {
...
 	}
 
-	newJob.Annotations["revisions"] = string(revisionsByte)
-
-	err = deleteJob(job.Namespace, job.Name)
-	if err != nil {
-		glog.Errorf("failed to rerun job %s, reason: %s", err, err)
-		return "", fmt.Errorf("failed to rerun job %s", jobName)
+	for i := 0; i < 3; i++ {
+		_, err = k8sClient.BatchV1().Jobs(namespace).Create(&newJob)
```
wnxn: It's better to add sleeping time before retrying and set retry times as a constant.

richardxz: The code has been modified according to your suggestion

PR #196, opened 2018-10-24
ignore the role/clusterrole which don't have "creator" annotation

richardxz:


pkg/models/controllers/types.go
```diff
@@ -49,6 +49,7 @@ const (
 	Warning      = "warning"
 	Error        = "error"
 	DisplayName  = "displayName"
+	creator      = "creator"
```
wnxn: Constant names usually start with an uppercase.

richardxz: i don‘t think so ，and start with uppercase in go usually mean it can be visit outside the package， it's not necessary in current scenario

PR #211, opened 2018-11-22
fix bug: rolebinding cannot delete

wansir:
fix bug: rolebinding cannot delete

### DIFF
--- a/pkg/models/controllers/namespaces.go
+++ b/pkg/models/controllers/namespaces.go
@@ -46,7 +46,7 @@ import (
 const (
 	provider                     = "kubernetes"
 	admin                        = "admin"
-	editor                       = "operator"
+	operator                     = "operator"
 	viewer                       = "viewer"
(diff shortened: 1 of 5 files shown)


pkg/models/controllers/namespaces.go
```diff
@@ -197,7 +190,7 @@ func (ctl *NamespaceCtl) createDefaultRoleBinding(namespace *v1.Namespace) error
 	viewerBinding.Subjects = make([]rbac.Subject, 0)
 
 	if workspace != "" {
-		workspaceViewer, err := ctl.K8sClient.RbacV1().ClusterRoleBindings().Get(fmt.Sprintf("system:%s:viewer", workspace), metaV1.GetOptions{})
+		workspaceViewer, err := ctl.K8sClient.RbacV1().ClusterRoleBindings().Get(fmt.Sprintf("system:%s:%s", workspace, "ws-viewer"), metaV1.GetOptions{})
```
zryfish: why not use your const variable WSViewer

PR #322, opened 2019-04-01
Refine tenant api

wansir:


pkg/models/iam/am.go
```diff
@@ -385,135 +501,139 @@ func NamespaceUsers(namespaceName string) ([]*models.User, error) {
...
-			for _, rule := range v.Rules {
-				rule.ResourceNames = nil
-				policyRules = append(policyRules, rule)
-			}
+	return workspaceRules
+}
 
-			rules := make([]models.SimpleRule, 0)
+//func merge(clusterRules, rules []models.SimpleRule) []models.SimpleRule {
+//	for _, clusterRule := range clusterRules {
```
zryfish: remove commented code

pkg/models/resources/pods.go
```diff
@@ -30,16 +31,155 @@ import (
...
+	}
+	return false
+}
+
+func podBelongToService(item *v1.Pod, serviceName string) bool {
+	service, err := informers.SharedInformerFactory().Core().V1().Services().Lister().Services(item.Namespace).Get(serviceName)
+	if err != nil {
+		return false
+	}
+	for k, v := range service.Spec.Selector {
```
zryfish: client-go has a more elegant way to determine if service selects a pod
 https://github.com/kubernetes/client-go/blob/89226ff90625425e6c7a87fe219dcefd75d28e53/listers/core/v1/service_expansion.go#L36
```
		selector := labels.Set(service.Spec.Selector).AsSelectorPreValidated()
		if selector.Matches(labels.Set(pod.Labels)) {
			services = append(services, service) // means service selected this pod
		}
```

pkg/models/resources/pods.go
```diff
@@ -30,16 +31,155 @@ import (
...
+		}
+	}
+	return true
+}
+
 // exactly Match
 func (*podSearcher) match(match map[string]string, item *v1.Pod) bool {
 	for k, v := range match {
 		switch k {
+		case "ownerKind":
```
zryfish: not a good habit to hard field name here

pkg/simple/client/kubesphere/kubesphereclient.go
```diff
@@ -0,0 +1,285 @@
+/*
```
zryfish: why we need this ?

PR #360, opened 2019-04-24
fix: applications api

wansir:


pkg/controller/job/job_controller.go
```diff
@@ -0,0 +1,297 @@
...
+
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		return err
+	}
+
+	job, err := v.jobLister.Jobs(namespace).Get(name)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return err
```
zryfish: if err is IsNotFound, normally, means job already been deleted, could just skip this one, no need to return err

pkg/controller/job/job_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil {
+		glog.Error(err)
+	}
+
+	if job.Annotations == nil {
+		job.Annotations = make(map[string]string)
+	}
+
+	job.Annotations[revisionsAnnotationKey] = string(revisionsByte)
+	_, err = v.client.BatchV1().Jobs(job.Namespace).Update(job)
```
zryfish: have you go through k8s jobController, what will happen when you update a job, I think it trigger k8s jobController to run. better double check source code, in case of job rerun when you update

pkg/controller/job/job_controller.go
```diff
@@ -0,0 +1,297 @@
...
+func (v *JobController) getCurrentRevision(item *batchv1.Job) JobRevision {
+	var revision JobRevision
+	for _, condition := range item.Status.Conditions {
+		if condition.Type == "Failed" && condition.Status == "True" {
+			revision.Status = Failed
+			revision.Reasons = append(revision.Reasons, condition.Reason)
+			revision.Messages = append(revision.Messages, condition.Message)
+		}
+
+		if condition.Type == "Complete" && condition.Status == "True" {
```
zryfish: is that possible item.Status.Conditions contains both Failed condtion.Type and Complete condition.Type? If true, here will be a problem, otherwise, use a elif to join these two `if` blocks to make it more clear

pkg/controller/job/types.go
```diff
@@ -0,0 +1,42 @@
...
+	Failed     = "failed"
+	Unfinished = "unfinished"
+	Completed  = "completed"
+	Pause      = "pause"
+)
+
+type JobRevisions map[int]JobRevision
+
+type JobRevision struct {
+	Status         string    `json:"status"`
```
zryfish: add `omitempty` to skip empty filed, will reduce response size

PR #371, opened 2019-04-28
request devops api with auth info

runzexia:


pkg/apigateway/caddy-plugin/authenticate/authenticate.go
```diff
@@ -130,7 +130,9 @@ func (h Auth) InjectContext(req *http.Request, token *jwt.Token) (*http.Request,
 	}
 
 	// hard code, support jenkins auth plugin
-	if httpserver.Path(req.URL.Path).Matches("/kapis/jenkins.kubesphere.io") || httpserver.Path(req.URL.Path).Matches("job") {
+	if httpserver.Path(req.URL.Path).Matches("/kapis/jenkins.kubesphere.io") ||
+		httpserver.Path(req.URL.Path).Matches("job") ||
```
zryfish: `Matches` does a prefix matching, so which kind of path will match keyword `job` ?

runzexia:
```
      proxy /kapis/jenkins.kubesphere.io http://ks-jenkins.kubesphere-devops-system.svc {
        without /kapis/jenkins.kubesphere.io
        transparent
      }
      
      proxy /job http://ks-jenkins.kubesphere-devops-system.svc {
        transparent
      }
```
In the old version, some non-rest api, like /job/xxx/job, was proxyed.

zryfish: ok, not a good pattern, need to make it deprecated soon

PR #379, opened 2019-04-30
add extra annotations

wansir:

### DIFF
--- a/pkg/models/iam/im.go
+++ b/pkg/models/iam/im.go
@@ -352,11 +352,11 @@ func ListUsers(conditions *params.Conditions, orderBy string, reverse bool, limi
 		}
 		switch orderBy {
 		case "username":
-			fallthrough
-		case "createTime":
-			return users[i].CreateTime.Before(users[j].CreateTime)
-		default:
 			return strings.Compare(users[i].Username, users[j].Username) <= 0
+		case "createTime":
+			fallthrough
+		default:
+			return users[i].CreateTime.Before(users[j].CreateTime)
 		}
 	})
 
--- a/pkg/models/resources/extraannotations.go
+++ b/pkg/models/resources/extraannotations.go
@@ -1,0 +1,74 @@
+/*
+
+ Copyright 2019 The KubeSphere Authors.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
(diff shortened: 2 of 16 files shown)


pkg/models/resources/extraannotations.go
```diff
@@ -0,0 +1,74 @@
...
+	return item
+}
+
+func isPvcInUse(pods []*v1.Pod, pvcName string) bool {
+	for _, pod := range pods {
+		volumes := pod.Spec.Volumes
+		for _, volume := range volumes {
+			pvc := volume.PersistentVolumeClaim
+			if pvc != nil && pvc.ClaimName == pvcName {
+				return true
```
zryfish: a pvc can be mounted by multiple pods, `kubectl describe pvc` can show pvc mounted by which pods. please have a look at this pr https://github.com/kubernetes/kubernetes/pull/65837/files

pkg/models/resources/resources.go
```diff
@@ -142,9 +143,9 @@ func ListResources(namespace, resource string, conditions *params.Conditions, or
 		return nil, err
 	}
 
-	for i, d := range result {
+	for i, item := range result {
 		if i >= offset && (limit == -1 || len(items) < limit) {
-			items = append(items, d)
+			items = append(items, injector.addExtraAnnotations(item))
```
zryfish: I think it's better to move add annotations part to `persistentvolumeclaims.go`, for now, only pvc needed to be dealt with differently

pkg/models/resources/extraannotations.go
wansir: just a status flag, this API is enough
https://github.com/kubesphere/kubesphere/blob/3e8035dc2b2848afaf896bc01626ecd185302e1a/pkg/models/resources/pods.go#L165-L168

pkg/models/iam/im.go
```diff
@@ -351,11 +351,11 @@ func ListUsers(conditions *params.Conditions, orderBy string, reverse bool, limi
 			j = tmp
 		}
 		switch orderBy {
-		case "username":
-			fallthrough
 		case "createTime":
 			return users[i].CreateTime.Before(users[j].CreateTime)
 		default:
+			fallthrough
+		case "username":
```
zryfish: so `username` is the default case, 
```
default:
        return strings.Compare(users[i].Username, users[j].Username) <= 0
```

pkg/models/status/status.go
```diff
@@ -34,13 +35,19 @@ func GetNamespacesResourceStatus(namespace string) (*workLoadStatus, error) {
...
+			notReadyStatus = resources.StatusFailed
+		default:
+			notReadyStatus = resources.StatusUpdating
 		}
 
-		notReadyList, err = resources.ListResources(namespace, resource, &params.Conditions{Match: map[string]string{"status": notReadyStatus}}, "", false, -1, 0)
+		notReadyList, err = resources.ListResources(namespace, resource, &params.Conditions{Match: map[string]string{resources.Status: notReadyStatus}}, "", false, -1, 0)
 
 		if err != nil {
 			return nil, err
```
zryfish: add log

PR #388, opened 2019-05-07
fix: log query range starts at namespace creation time

huanggze:


pkg/models/log/logcollector.go
```diff
@@ -181,6 +183,39 @@ func MatchNamespace(namespaceMatch string, namespaceFilled bool, namespaces []st
...
+		}
+
+		for _, ns := range nsList {
+			namespaceWithCreationTime[ns.Name] = strconv.FormatInt(ns.CreationTimestamp.UnixNano()/int64(time.Millisecond), 10)
+		}
+
+		return true, namespaceWithCreationTime
+	}
+
+	for _, item := range namespaces {
```
wansir: duplicate code

PR #405, opened 2019-05-11
wrap panic

runzexia:

log panic stack info & return error


cmd/ks-apiserver/app/server.go
```diff
@@ -37,6 +38,7 @@ import (
 	"kubesphere.io/kubesphere/pkg/simple/client/devops_mysql"
 	"log"
 	"net/http"
+	goRuntime "runtime"
```
zryfish: `goruntime` will be just fine

cmd/ks-iam/app/server.go
```diff
@@ -133,3 +136,18 @@ func initializeAdminJenkins() {
 func initializeDevOpsDatabase() {
 	devops_mysql.OpenDatabase()
 }
+
+func logStackOnRecover(panicReason interface{}, httpWriter http.ResponseWriter) {
```
zryfish: don't repeat yourself, this function is exactly same with upper one

PR #410, opened 2019-05-11
fix: filter out old namespace data from workspace cpu and memory metrics

huanggze:


benjaminhuo:
This fix requires additional container rules to be added to prometheus which will cost lots of memory. So maybe we can simply limit the ns level metrics range query start time

huanggze:
> This fix requires additional container rules to be added to prometheus which will cost lots of memory. So maybe we can simply limit the ns level metrics range query start time

I reshaped that commit [#11](https://github.com/kubesphere/kubernetes-mixin/pull/11). So I guess the fix should be good to merge now

benjaminhuo:
It's much better now. Clever move👍

PR #419, opened 2019-05-13
fix: log statitiscs

huanggze:

### DIFF
--- a/pkg/simple/client/elasticsearch/esclient.go
+++ b/pkg/simple/client/elasticsearch/esclient.go
@@ -114,7 +114,8 @@ type QueryWord struct {
 }
 
 type MainHighLight struct {
-	Fields []interface{} `json:"fields,omitempty"`
+	Fields       []interface{} `json:"fields,omitempty"`
+	FragmentSize int           `json:"fragment_size"`
 }
 
 type LogHighLightField struct {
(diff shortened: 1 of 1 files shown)


pkg/simple/client/elasticsearch/esclient.go
```diff
@@ -137,25 +137,33 @@ type EmptyField struct {
 }
 
 type StatisticsAggs struct {
```
zryfish: I believe Agg means Aggregation? don't use initial here, isn't easy to understand

huanggze: since the `aggs` is the field name used by es search body. Should we keep it in line with es?
image

huanggze: `aggregations` has been used in response. So you can find type `StatisticsAggregations` for response in code to distinguish these two naming

image

zryfish: OK, better to add comments to clarify

pkg/simple/client/elasticsearch/esclient.go
```diff
@@ -501,28 +495,10 @@ func parseQueryResult(operation int, param QueryParameters, body []byte, query [
...
-				containerResult.Count = container.Count
-				namespaceResult.Containers = append(namespaceResult.Containers, containerResult)
-			}
-
-			statisticsResult.Namespaces = append(statisticsResult.Namespaces, namespaceResult)
-		}
-
-		queryResult.Statistics = &statisticsResult
+		var statisticsResponse StatisticsResponseAggregations
+		jsonIter.Unmarshal(response.Aggregations, &statisticsResponse)
```
zryfish: what if unmarshal failed , content in response can be anything

huanggze: thx, see the updated commit

PR #424, opened 2019-05-14
fix: application api

wansir:


pkg/models/iam/policy/policy.go
```diff
@@ -358,18 +358,9 @@ var (
...
-							Verbs:     []string{"get"},
-							APIGroups: []string{"openpitrix.io"},
-							Resources: []string{"app_version/*"},
-						},
-						{
-							Verbs:     []string{"*"},
+							Verbs:     []string{"get", "list"},
 							APIGroups: []string{"openpitrix.io"},
-							Resources: []string{"apps", "clusters"},
+							Resources: []string{"apps", "clusters", "repos", "app_versions", "app_version/*"},
```
zryfish: consider refactor this section into a configuration file loaded mounted to application, so you don't need to change this frequently

rayzhou2017: Yes, I think so. But since GA is coming, we can hard code right now. Let's use configuration file in the next version.

PR #451, opened 2019-05-27
 filter jenkins header

soulseen:


soulseen:
@runzexia @zryfish @wansir

pkg/models/devops/devops.go
```diff
@@ -698,6 +698,7 @@ func jenkinsClient(baseUrl string, req *http.Request) ([]byte, http.Header, erro
 	defer resp.Body.Close()
 
 	if resp.StatusCode >= http.StatusBadRequest {
+		log.Errorf("%+v", string(resBody))
```
runzexia: also return resBody to frontend ?

runzexia: And do you need to log some context information such as the request URL?

soulseen: no，just log resBody. return error mssage to frontend.
already log context information in previous func.

runzexia: At least need to expose http status code to the front end, always 500 is not a good choice

soulseen: yes，will return resp.StatusCode same as Jenkins returns if StatusCode > 200, not always 500.

pkg/models/devops/devops.go
```diff
@@ -544,8 +545,8 @@ func ToJson(req *http.Request) ([]byte, error) {
 }
 
 func GetNotifyCommit(req *http.Request) ([]byte, error) {
-	baseUrl := fmt.Sprintf(jenkins.Server + GetNotifyCommitUrl + req.URL.RawQuery)
-	log.Infof("Jenkins-url: " + baseUrl)
+	baseUrl := fmt.Sprint(jenkins.Server + GetNotifyCommitUrl + req.URL.RawQuery)
```
runzexia: why use fmt.Sprint here?

soulseen: here the var req.URL.RawQuery mabey have some special characters such as "%s", but not should be formart.

runzexia: Is there a problem with using "+" directly?

soulseen: It's ok in test.

pkg/models/devops/devops.go
```diff
@@ -68,10 +68,11 @@ func SearchPipelines(req *http.Request) ([]byte, error) {
 }
 
 func SearchPipelineRuns(projectName, pipelineName string, req *http.Request) ([]byte, error) {
-	baseUrl := fmt.Sprintf(jenkins.Server+SearchPipelineRunUrl+req.URL.RawQuery, projectName, pipelineName)
+	baseUrl := fmt.Sprintf(jenkins.Server+SearchPipelineRunUrl, projectName, pipelineName)
+
 	log.Infof("Jenkins-url: " + baseUrl)
```
runzexia: this log maybe format error

PR #466, opened 2019-06-04
fix: update log statistics response, remove sorting by time

huanggze:


huanggze:
No need to return the time of the first record, I remove  relevant code snippets then

PR #480, opened 2019-06-12
api: change HTTP method of fluentbit output updating from POST to PUT

huanggze:


rayzhou2017:
Is the console changed too according to this commit?

huanggze:
Yep

huanggze:
Yep. The change has been verified on dev environment.

PR #490, opened 2019-06-20
 update api docs

wansir:


pkg/models/applications/applications.go
```diff
@@ -46,7 +46,7 @@ type Application struct {
 	VersionId   string            `json:"version_id" description:"application version id"`
 	Status      string            `json:"status" description:"application status"`
 	UpdateTime  time.Time         `json:"updateTime" description:"update time"`
-	CreateTime  time.Time         `json:"createTime" description:"create name"`
+	CreateTime  time.Time         `json:"createTime" description:"create time"`
```
rayzhou2017: Should it be UpdatedTime, CreatedTime?

wansir: the field is also used in openpitrix, should be different?

PR #496, opened 2019-06-23
logging: fix no result displayed problem when searching deleted pods' logs by workload name

huanggze:
There is a problem when searching logs of deleted pods, such as completed jobs, by workload name, issue [#1001](https://git.internal.yunify.com/KubeSphere/kubesphere-console/issues/1001). It is because we currently leverage SharedInformerFactory to list pods. Deleted pods may be removed out from the list dynamically.

The fix uses regexp to filter pods by constructing pod full name based on workload name instead. However, it would be better if fluent bit can gather pod's owner information as collecting logs, then we just filter workload name directly rather than wrting regexp. But we cannot do much now until the upstream project makes a change.


zryfish:
target release 2.1

PR #535, opened 2019-07-01
rm unused devops api

runzexia:


zryfish:
shouldn't the commit message be "rm unused api parameter"

PR #567, opened 2019-07-05
refine components API and drain node API docs

FeynmanZhou:
Signed-off-by: 周鹏飞@yunify <pengfeizhou@yunify.com>


FeynmanZhou:
@rayzhou2017 @wansir

PR #572, opened 2019-07-06
fix: recreate kubectl pod

wansir:

### DIFF
--- a/pkg/models/iam/am.go
+++ b/pkg/models/iam/am.go
@@ -628,6 +628,18 @@ func CreateClusterRoleBinding(username string, clusterRoleName string) error {
 		return err
 	}
 
+	// create kubectl pod if cluster role is cluster-admin
+	if clusterRoleName == constants.ClusterAdmin {
+		if err := kubectl.CreateKubectlDeploy(username); err != nil {
+			glog.Error("create user terminal pod failed", username, err)
+		}
+		// delete kubectl pod if cluster role is not cluster-admin, whether it exists or not
+	} else {
+		if err := kubectl.DelKubectlDeploy(username); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/models/iam/am.go
```diff
@@ -636,6 +636,16 @@ func CreateClusterRoleBinding(username string, clusterRoleName string) error {
 	clusterRoleBindingLister := informers.SharedInformerFactory().Rbac().V1().ClusterRoleBindings().Lister()
 	found, err := clusterRoleBindingLister.Get(username)
```
runzexia: handle error

pkg/models/iam/im.go
```diff
@@ -797,10 +802,6 @@ func CreateUser(user *models.User) (*models.User, error) {
 		setAvatar(user.Username, user.AvatarUrl)
 	}
 
-	if err := kubeconfig.CreateKubeConfig(user.Username); err != nil {
```
runzexia: if found.RoleRef.Name != constants.ClusterAdmin  => delete kubectl deploy?

pkg/models/iam/am.go
```diff
@@ -657,11 +669,6 @@ func CreateClusterRoleBinding(username string, clusterRoleName string) error {
 			glog.Errorln("delete cluster role binding", err)
 			return err
 		}
-		if found.RoleRef.Name == constants.ClusterAdmin {
```
runzexia: if found.RoleRef.Name != constants.ClusterAdmin => delete kubectl deploy?

wansir: delete kubectl pod if cluster role is not cluster-admin, whether it exists or not

PR #624, opened 2019-08-05
add ns networkpolicy, based on calicov3 api

magicsong:
this pr has done following things:

1. added a nsnetworkpolicy api based on calico.  
2. added related controller and unit-test
3. added an e2e-test (until we have a stable jenkins)


hack/network-test.sh
```diff
@@ -0,0 +1,57 @@
...
+#this script must invoked in the root directory of this repo
+
+
+tag=`git rev-parse --short HEAD`
+IMG=magicsong/ks-network:$tag
+DEST=/tmp/manager.yaml
+SKIP_BUILD=no
+
+echo "try to delete old yaml"
+kubectl delete -f $DEST
```
zheng1: Should check whether `$DEST` exist or not

PR #628, opened 2019-08-07
refine generate tools & add CRD s2i binary

runzexia:
Use `controller-tool` v0.0.12 to generate `crd yaml`& `RBAC`
Use `go generate` to generate `deepcopy` & `openapi struct`
Use `hack/genereate_client.sh` to generate `clientset`


runzexia:
This PR mainly fixes the problem of code generation tools after migrating go mod.
Although this code has a new CRD structure, it does not have a controller/api implementation and therefore does not affect the operation of the code.
I think this code can be merged quickly so that everyone can use the same code generation tool.

PR #656, opened 2019-08-21
logging: use elastic client for go

huanggze:

**What type of PR is this?**

/kind cleanup

using official elastic client-go brings convenience for adding new features in the feature

Fixes #

```
verified with elastic v5.6.16, v6.7.0, v7.1.0
```

```docs

```


pkg/simple/client/elasticsearch/esclient.go
```diff
@@ -30,28 +28,31 @@ import (
...
-	mu        sync.RWMutex
-	esConfigs *ESConfigs
-)
-
-type ESConfigs struct {
-	Host  string
-	Port  string
-	Index string
-}
+	mu     sync.Mutex
```
Ma-Dan: Why use Mutex instead of RWMutex?

pkg/simple/client/elasticsearch/esclient.go
```diff
@@ -476,24 +478,24 @@ func parseQueryResult(operation int, param QueryParameters, body []byte, query [
 	case OperationStatistics:
 		var statisticsResponse StatisticsResponseAggregations
 		err := jsonIter.Unmarshal(response.Aggregations, &statisticsResponse)
-		if err != nil {
+		if err != nil && response.Aggregations != nil {
```
Ma-Dan: Note: response with null aggregation from server should not be treated as server error.

pkg/simple/client/elasticsearch/esclient.go
huanggze: Because now we use elastic client, no need to frequently read config for structing http request as we once did: [L573](https://github.com/kubesphere/kubesphere/pull/656/files#diff-28320d9ea55aa094d44988b9161bbe6aL573)

huanggze: index is still read frequently. will fix and resubmitted

PR #666, opened 2019-08-26
modify gen

magicsong:
**What type of PR is this?**
 /kind bug

do not need to put repo in $GOPATH


runzexia:
/hold 

We should use a version management tool to manage the tools of the development process. This PR destroys this.

runzexia:
https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away

Mentioned in the official documentation for the go language. Vendor is not recommended in vgo, but it is not recommended in go mod.
So we can use the vendor to keep some dependent programs, before we find the way to ensure the program dependencies without using the vendor.

PR #668, opened 2019-08-27
refactor: openpitrix module

wansir:
**What type of PR is this?**

/kind api-change

1. replace openpitrix HTTP client with gRPC client.
2. support workspace level repo.
3. openpitrix decouple


pkg/simple/client/factory.go
```diff
@@ -302,7 +302,12 @@ func (cs *ClientSet) S3() (*s2is3.S3Client, error) {
 func (cs *ClientSet) OpenPitrix() (*openpitrix.OpenPitrixClient, error) {
 	var err error
 
-	if cs.csoptions.openPitrixOptions == nil || cs.csoptions.openPitrixOptions.APIServer == "" {
+	if cs.csoptions.openPitrixOptions.RepoManagerEndpoint == "" ||
```
zryfish: why skip nil check ? I think you missed RepoIndexerEndpoint?

PR #686, opened 2019-08-30
fix: email validation

### LINKED ISSUE #683: 无法配置用户邮箱
安装模式：单机模式和集群模式

问题：
安装成功后更改admin用户邮箱
image

将默认邮箱地址改为zhoul@asto-inc.com，点击保存，系统报错：invalid email
image

新建用户也无法配置邮箱

wansir:
**What type of PR is this?**

/kind bug
/area iam

Use the RFC5322 regular verification email address

Fixes #683

### DIFF
--- a/pkg/apiserver/iam/im.go
+++ b/pkg/apiserver/iam/im.go
@@ -19,10 +19,10 @@ package iam
 
 import (
 	"fmt"
-	"github.com/golang/glog"
+	"k8s.io/klog"
 	"kubesphere.io/kubesphere/pkg/params"
 	"net/http"
-	"regexp"
+	"net/mail"
 	"strings"
 
 	"github.com/emicklei/go-restful"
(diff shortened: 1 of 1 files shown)


pkg/apiserver/iam/im.go
```diff
@@ -35,7 +35,8 @@ import (
 )
 
 const (
-	emailRegex = "^[a-z0-9]+([._\\-]*[a-z0-9])*@([a-z0-9]+[-a-z0-9]*[a-z0-9]+.){1,63}[a-z0-9]+$"
+	// RFC 5322 regex, https://tools.ietf.org/html/rfc5322
+	emailRegex = "^[a-zA-Z0-9!#$%&'*+/=?^_`{|}~-]+(?:\\.[a-zA-Z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?\\.)+([a-zA-Z]{2,})$"
```
zryfish: why we change to this ? Also I find this expression can't really match email address. I tested it on https://www.regextester.com/, the older expression do match some though. Please double check.

wansir: An email address may contain upper-case letters as defined in RFC5322 (http://tools.ietf.org/html/rfc5322#section-3.4.1).

pkg/apiserver/iam/im.go
```diff
@@ -22,7 +22,7 @@ import (
 	"github.com/golang/glog"
```
zryfish: use klog instead of glog

zryfish: Other than is good

PR #692, opened 2019-09-02
✨support calico with k8s storage

magicsong:
**What type of PR is this?**
networkpolicy now supports calico with k8s storage

/kind feature
/area networking


kustomize/network/calico-etcd/patch_image_name.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: magicsong/ks-network:90fa3885
+      - image: magicsong/ks-network:f22e8ea9
```
zryfish: push image to kubespheredev/ks-network, I've already created that repo

kustomize/network/calico-etcd/kustomization.yaml
```diff
@@ -19,4 +20,4 @@ secretGenerator:
 patchesStrategicMerge:
   - patch_image_name.yaml
 
-namespace: network-test-90fa3885
+namespace: network-test-f22e8ea9
```
zryfish: if it's a temporary name, shouldn't we generate it during testing? hardcode the value means you will change it every time when creating a pr .

magicsong: it is a patch and will change each time. it is not hardcoded and it will be used by test or publish

kustomize/network/calico-etcd/patch_image_name.yaml
magicsong: done. but editing here is useless, refer to <https://github.com/kubesphere/kubesphere/pull/692/commits/d3e694c6f088331ae876ffa40b3a0cecdfca8681>

PR #694, opened 2019-09-02
search image from dockerhub

soulseen:
Signed-off-by: soulseen <sunzhu@yunify.com>

**What type of PR is this?**
/kind documentation
/kind feature

search image from dockerhub

### DIFF
--- a/pkg/apis/resources/v1alpha2/register.go
+++ b/pkg/apis/resources/v1alpha2/register.go
@@ -234,6 +234,23 @@ func addWebService(c *restful.Container) error {
 		Returns(http.StatusOK, ok, registriesmodel.ImageDetails{}),
 	)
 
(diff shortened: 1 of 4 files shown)


runzexia:
I think there needs to be some way to determine if it is an offline environment. I think the current implementation will have some problems in the offline environment.

Is the environment offline or is it judged from the code or from the global configuration? Which method do you think is better? 
/cc @zryfish

zryfish:
Why we need this code? IMO, dockerhub api can be called directly from console, we don't have to implement api again.

soulseen:
same effect with call dockerhub api directly from console, but can not generate api documentation.

runzexia:
I think this PR can be closed because the front end will use dockerhub api.

/close

PR #702, opened 2019-09-05
refactor dependent service client creation

zryfish:
**What type of PR is this?**
/kind feature

1.  refactor kubesphere dependency service client creation, we can
disable dependency by config
2. dependencies can be configured by configuration file
3. refactor cmd package using cobra.Command, so we can use hypersphere
to invoke command sepearately. Later we only need to build one image to
contains all kubesphere core components. One command to rule them all!

```
```

```docs
```


pkg/simple/client/factory.go
```diff
@@ -0,0 +1,143 @@
...
+func ClientSets() *ClientSet {
+	return sharedClientSet
+}
+
+func NewClientSetFactory(c *ClientSetOptions, stopCh <-chan struct{}) *ClientSet {
+	sharedClientSet = &ClientSet{}
+
+	// create clients one by one
+	if c.mySQLOptions != nil && c.mySQLOptions.Host != "" {
+		sharedClientSet.mySQLClient = mysql.NewMySQLClient(c.mySQLOptions, stopCh)
```
runzexia: Initializing here requires that other services in the cluster be started first.
Should we add some logic to ClientSet.XXXClient to perform another initialization?

pkg/simple/client/factory.go
```diff
@@ -0,0 +1,143 @@
...
+
+	if c.s3Options != nil && c.s3Options.Endpoint != "" {
+		sharedClientSet.s3Client = s2is3.NewS3Client(c.s3Options)
+	}
+
+	return sharedClientSet
+}
+
+// no guarantee to return a valid connection
+func (cs *ClientSet) MySQL() *mysql.Database {
```
runzexia: The idea now is to use the Client is nil to determine whether the function is enabled?
In the wrong configuration, Client will also be nil.

vendor/modules.txt
```diff
@@ -880,7 +880,7 @@ k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1
 k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset
 k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/scheme
 k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1
-# k8s.io/apimachinery v0.0.0-20190221213512-86fb29eff628 => k8s.io/apimachinery v0.0.0-20190221213512-86fb29eff628
+# k8s.io/apimachinery v0.0.0-20190831074630-461753078381 => k8s.io/apimachinery v0.0.0-20190221213512-86fb29eff628
```
runzexia: Dependency changes need to use hack script to operate, and it is best to test whether crd-doc-gen can run

pkg/simple/client/mysql/mysql.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+type MySQLClient struct {
+	database *Database
+}
+
+func NewMySQLClient(option *MySQLOptions, stopCh <-chan struct{}) *MySQLClient {
+	m := &MySQLClient{}
+
+	conn, err := dbr.Open("mysql", fmt.Sprintf("%s:%s@tcp(%s)/devops?parseTime=1&multiStatements=1&charset=utf8mb4&collation=utf8mb4_unicode_ci", option.Username, option.Password, option.Host), nil)
```
runzexia: The specific SCHEMA used by the module is currently included in the link information. Do we need to move the mysql configuration to a different module?

PR #736, opened 2019-09-12
update devops flags default value

runzexia:
/cc @zryfish 
In minio, the region name is `us-east-1`, so modify the default value.

### DIFF
--- a/pkg/simple/client/s2is3/options.go
+++ b/pkg/simple/client/s2is3/options.go
@@ -44,7 +44,7 @@ func (s *S3Options) AddFlags(fs *pflag.FlagSet) {
 		"Endpoint to access to s3 object storage service, if left blank, the following options "+
 		"will be ignored.")
 
-	fs.StringVar(&s.Region, "s3-region", s.Region, ""+
+	fs.StringVar(&s.Region, "s3-region", "us-east-1", ""+
 		"Region of s3 that will access to, like us-east-1.")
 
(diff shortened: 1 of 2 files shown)


pkg/simple/client/s2is3/options.go
```diff
@@ -44,7 +44,7 @@ func (s *S3Options) AddFlags(fs *pflag.FlagSet) {
 		"Endpoint to access to s3 object storage service, if left blank, the following options "+
 		"will be ignored.")
 
-	fs.StringVar(&s.Region, "s3-region", s.Region, ""+
+	fs.StringVar(&s.Region, "s3-region", "us-east-1", ""+
```
zryfish: change NewS3Options() to set default value
```

func NewS3Options() *S3Options {
	return &S3Options{
		Endpoint:        "",
		Region:          "us-east-1",
		DisableSSL:      false,
		ForcePathStyle:  true,
		AccessKeyID:     "",
		SecretAccessKey: "",
		SessionToken:    "",
		Bucket:          "",
	}
}
```

pkg/simple/client/s2is3/options.go
```diff
@@ -18,14 +18,14 @@ type S3Options struct {
 
 func NewS3Options() *S3Options {
 	return &S3Options{
-		Endpoint:        "",
-		Region:          "",
+		Endpoint:        "http://openpitrix-minio.openpitrix-system.svc:9000",
```
zryfish: Enpoint/host/apiserver fields are left empty for purpose, since we treat components as disabled if these fields was empty. If we set default values for these fields, means enable them by default

runzexia:
fix @zryfish

PR #749, opened 2019-09-16
Change glog to klog 

zryfish:
**What type of PR is this?**
/kind bug
/kind feature

1. We are switching glog to klog, so this pr changes all glog reference to klog
2. create new package api,  which used to put module types. Better organized, and to avoid cyclic import.
3. move filter/params/options to server pkg

Fixes #

```
```

```docs

```


pkg/api/devops/v1alpha2/types.go
```diff
@@ -0,0 +1,20 @@
+package v1alpha2
+
+import "time"
+
+type PageableDevOpsProject struct {
```
runzexia: What is the purpose of this change?

zryfish: currently, we create a client instance through client factory,  so import path will become
```
kubesphere.io/kubesphere/pkg/simple/client/devops -> kubesphere.io/kubesphere/pkg/simple/client
kubesphere.io/kubesphere/pkg/simple/client/kubesphere -> kubesphere.io/kubesphere/pkg/simple/client
```
In kubesphere client, https://github.com/kubesphere/kubesphere/blob/master/pkg/simple/client/kubesphere/kubesphereclient.go#L48 , this will cause cyclic import. 

Plus, it's better organized if we place types definition in a independent package, and we can avoid potential cyclic import.

pkg/apis/tenant/v1alpha2/register.go
```diff
@@ -22,16 +22,16 @@ import (
 	"github.com/emicklei/go-restful-openapi"
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	v1alpha22 "kubesphere.io/kubesphere/pkg/api/devops/v1alpha2"
```
runzexia: devopsv1alpha2?

zryfish: yeah, sure

PR #755, opened 2019-09-17
add github actions

runzexia:
/cc @zryfish


runzexia:
test in https://github.com/runzexia/kubesphere/pull/1
But I don’t know why there is no trigger in kubesphere.

PR #768, opened 2019-09-19
feat: allow to export logs

huanggze:

**What type of PR is this?**

/kind feature
/kind cleanup
/kind api-change

Fixes #

```
Search API has performance issue, use Scroll API instead
```

```docs

```

### DIFF
--- a/pkg/apis/logging/v1alpha2/register.go
+++ b/pkg/apis/logging/v1alpha2/register.go
@@ -47,7 +47,7 @@ func addWebService(c *restful.Container) error {
 
 	ws.Route(ws.GET("/cluster").To(logging.LoggingQueryCluster).
 		Doc("Query logs against the cluster.").
-		Param(ws.QueryParameter("operation", "Query type. This can be one of three types: query (for querying logs), statistics (for retrieving statistical data), and histogram (for displaying log count by time interval). Defaults to query.").DefaultValue("query").DataType("string").Required(false)).
+		Param(ws.QueryParameter("operation", "Operation type. This can be one of four types: query (for querying logs), statistics (for retrieving statistical data), histogram (for displaying log count by time interval) and export (for exporting logs). Defaults to query.").DefaultValue("query").DataType("string").Required(false)).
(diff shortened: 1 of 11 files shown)


pkg/apis/logging/v1alpha2/register.go
```diff
@@ -71,6 +71,28 @@ func addWebService(c *restful.Container) error {
 		Consumes(restful.MIME_JSON, restful.MIME_XML).
 		Produces(restful.MIME_JSON)
 
+	ws.Route(ws.GET("/cluster/export").To(logging.LoggingClusterExport).
```
zryfish: export is a verb, not suitable. I think logs would be good

pkg/apiserver/logging/logging.go
```diff
@@ -166,13 +166,14 @@ func logExport(param v1alpha2.QueryParameters, response *restful.Response) {
...
-	// retrieve 1000 records each time
+	// export 1000 records in every iteration
 	param.Size = 1000
 	// from is not allowed in a scroll context
 	param.From = 0
 
 	var scrollId string
-	for {
+	// limit to retrieve max 100k records
+	for i := 0; i < 100; i++ {
```
benjaminhuo: 这个 100 是不是单独作为一个 const 变量好点？

huanggze: 暂时只有这里有这样的限制，直接写限制 100 次循环应该也可以

pkg/apiserver/logging/logging.go
```diff
@@ -188,22 +139,86 @@ func LoggingDeleteFluentbitOutput(request *restful.Request, response *restful.Re
...
+	// export 1000 records in every iteration
+	param.Size = 1000
+	// from is not allowed in a scroll context
+	param.From = 0
+
+	var scrollId string
+	// limit to retrieve max 100k records
+	for i := 0; i < 100; i++ {
+		var res *v1alpha2.QueryResult
+		if scrollId == "" {
```
zryfish: What if user cancelled request in the middle of downloading? I believe this loop will continue. Need to break loop quickly if user cancelled downloading. You can try this https://gist.github.com/zryfish/353dddb2f7eb4dd99e195b0270d52e9b

huanggze: we interrupt downloading by capture error from io.Copy

zryfish: no, that's not right. Even request is cancelled, response is still there, keep writing. I modified the example, you can try

PR #771, opened 2019-09-20
fix: ldap connection pool leak

wansir:

**What type of PR is this?**

/kind bug

fix ldap connection pool leak

### DIFF
--- a/pkg/simple/client/ldap/ldap.go
+++ b/pkg/simple/client/ldap/ldap.go
@@ -18,6 +18,7 @@
 package ldap
 
 import (
+	"fmt"
 	"github.com/go-ldap/ldap"
 	"k8s.io/klog"
 )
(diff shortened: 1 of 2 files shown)


pkg/simple/client/ldap/ldap.go
```diff
@@ -58,24 +58,24 @@ func NewLdapClient(options *LdapOptions, stopCh <-chan struct{}) (*LdapClient, e
...
-		err = conn.Bind(l.options.ManagerDN, l.options.ManagerPassword)
-		if err != nil {
-			conn.Close()
-			klog.Error(err)
-			return nil
-		}
-		return conn
+func (l *LdapClient) NewConn() (ldap.Client, error) {
+	if l.pool == nil {
+		klog.Fatalln("ldap connection pool is not initialized")
```
zryfish: not appropriate to do fatal logging, since we will check return values, I think it's better to return error instead

PR #811, opened 2019-09-27
fix: filter app template in built-in repo

wansir:

**What type of PR is this?**

/kind bug

filter app template in built-in repo

### DIFF
--- a/pkg/models/openpitrix/apps.go
+++ b/pkg/models/openpitrix/apps.go
@@ -32,6 +32,11 @@ import (
 	"strings"
 )
 
+const (
+	BuiltinRepoId = "repo-helm"
+	StatusActive  = "active"
+)
+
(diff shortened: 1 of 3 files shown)


pkg/models/openpitrix/apps.go
```diff
@@ -53,7 +53,12 @@ func ListApps(conditions *params.Conditions, orderBy string, reverse bool, limit
 		describeAppsRequest.CategoryId = strings.Split(categoryId, "|")
 	}
 	if repoId := conditions.Match["repo"]; repoId != "" {
-		describeAppsRequest.RepoId = strings.Split(repoId, "|")
+		// hard code, app template in built-in repo has no repo_id attribute
+		if repoId == "repo-helm" {
```
zryfish: need to use a constant variable here

Issue #898, opened 2019-10-10
DevOps部署到测试集群阶段，报连不上k8s集群的错误


jinmin527:
参考https://kubesphere.io/docs/v2.0/zh-CN/quick-start/devops-online/测试的时候在deploy to dev阶段保持。日志如下：
deploy to dev?
Proceed or Abort
Approved by project-regular
[Pipeline] kubernetesDeploy
Starting Kubernetes deployment
ERROR: ERROR: io.fabric8.kubernetes.client.KubernetesClientException: An error has occurred.
hudson.remoting.ProxyException: io.fabric8.kubernetes.client.KubernetesClientException: An error has occurred.
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:62)
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:53)
	at io.fabric8.kubernetes.client.BaseClient.<init>(BaseClient.java:71)
	at io.fabric8.kubernetes.client.BaseClient.<init>(BaseClient.java:55)
	at io.fabric8.kubernetes.client.DefaultKubernetesClient.<init>(DefaultKubernetesClient.java:61)
	at com.microsoft.jenkins.kubernetes.KubernetesClientWrapper.<init>(KubernetesClientWrapper.java:76)
	at com.microsoft.jenkins.kubernetes.KubernetesDeployContext$ClientWrapperFactoryImpl.buildClient(KubernetesDeployContext.java:466)
	at com.microsoft.jenkins.kubernetes.command.DeploymentCommand$DeploymentTask.doCall(DeploymentCommand.java:141)
	at com.microsoft.jenkins.kubernetes.command.DeploymentCommand$DeploymentTask.call(DeploymentCommand.java:122)
	at com.microsoft.jenkins.kubernetes.command.DeploymentCommand$DeploymentTask.call(DeploymentCommand.java:105)
	at hudson.remoting.UserRequest.perform(UserRequest.java:212)
	at hudson.remoting.UserRequest.perform(UserRequest.java:54)
	at hudson.remoting.Request$2.run(Request.java:369)
	at hudson.remoting.InterceptingExecutorService$1.call(InterceptingExecutorService.java:72)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[...]

Issue #917, opened 2019-10-11
generate kubeconfig serverURL should include scheme


runzexia:
In some env, ks-apiserver generated kubeconfig don't include scheme.
This causes clients in some languages to fail to parse correctly when using this config.
We should keep kubesphere able to generate standard kubeconfig in any environment.

https://github.com/kubesphere/kubesphere/issues/898

PR #919, opened 2019-10-11
fix missing scheme in kubeconfig

runzexia:

/kind bug
/cc @zryfish @wansir 
/area apiserver

Fixes https://github.com/kubesphere/kubesphere/issues/898 https://github.com/kubesphere/kubesphere/issues/917


pkg/simple/client/k8s/kubernetes.go
```diff
@@ -35,13 +36,19 @@ func NewKubernetesClientOrDie(options *KubernetesOptions) *KubernetesClient {
 
 	config.QPS = options.QPS
 	config.Burst = options.Burst
-
+	var master string
```
zryfish: Agreed. But I think it's better to move this section to the end of this function, since we assign options.Master to instance `k`, also need to check options.Master missing scheme or not

PR #953, opened 2019-10-12
feat: waiting for rolebinding created

wansir:

**What type of PR is this?**

/kind feature

return after rolebinding created when create namespace. Make sure that namespace creator have sufficient permissions.

### DIFF
--- a/pkg/models/tenant/tenant.go
+++ b/pkg/models/tenant/tenant.go
@@ -18,7 +18,10 @@
 package tenant
 
 import (
+	"fmt"
 	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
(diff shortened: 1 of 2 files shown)


pkg/models/tenant/tenant.go
```diff
@@ -43,8 +46,29 @@ func CreateNamespace(workspaceName string, namespace *v1.Namespace, username str
...
 
 	namespace.Labels[constants.WorkspaceLabelKey] = workspaceName
+	created, err := client.ClientSets().K8s().Kubernetes().CoreV1().Namespaces().Create(namespace)
+	if err != nil {
+		klog.Error(err)
+		return nil, err
+	}
+
+	timeout := int64(10)
+	w, err := client.ClientSets().K8s().Kubernetes().RbacV1().RoleBindings(namespace.Name).Watch(metav1.ListOptions{TimeoutSeconds: &timeout})
```
zryfish: apiserver is only responsible for query, watching a object creation should be placed in controller

PR #991, opened 2019-10-15
fix: add missing parameter

wansir:

**What type of PR is this?**

/kind bug

1. add missing parameters
2. record the correct operator
3. set default categories when creating application template.

#976


wansir:
/assign @pengcong06  

When creating an application template,  we need to add  category_id parameter or set the category_id  to ctg-uncategorized  by default.

PR #1038, opened 2019-10-16
fix: sort deployment by last update time

wansir:

**What type of PR is this?**

/kind bug

1. sort deployment by last update time.
2. sort pod by start time.

### DIFF
--- a/pkg/models/resources/pods.go
+++ b/pkg/models/resources/pods.go
@@ -221,6 +221,14 @@ func (*podSearcher) fuzzy(fuzzy map[string]string, item *v1.Pod) bool {
 
 func (*podSearcher) compare(a, b *v1.Pod, orderBy string) bool {
 	switch orderBy {
+	case StartTime:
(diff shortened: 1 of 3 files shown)


pkg/models/resources/pods.go
```diff
@@ -221,6 +221,14 @@ func (*podSearcher) fuzzy(fuzzy map[string]string, item *v1.Pod) bool {
...
 func (*podSearcher) compare(a, b *v1.Pod, orderBy string) bool {
 	switch orderBy {
+	case StartTime:
+		if a.Status.StartTime == nil {
+			return true
+		}
+		if b.Status.StartTime == nil {
+			return false
+		}
+		return a.Status.StartTime.Before(b.Status.StartTime)
```
runzexia: if a is older than b. `true`
if a is not started . `true`
if a is started and b is not started `false` / / If b is not started, it should be false？

wansir: fixed

PR #1104, opened 2019-10-21
fix: null pointer exception

wansir:

**What type of PR is this?**

/kind bug

ensure redis is connected  when startup


pkg/apigateway/caddy-plugin/authenticate/auto_load.go
```diff
@@ -68,57 +69,57 @@ func parse(c *caddy.Controller) (Rule, error) {
 				switch c.Val() {
 				case "path":
 					if !c.NextArg() {
-						return rule, c.ArgErr()
+						return nil, c.ArgErr()
```
runzexia: Why return nil here?

PR #1180, opened 2019-10-24
support more redis client options

wansir:

**What type of PR is this?**

/kind feature

support maxRetries options, fix EOF errors when network is unreliable


zryfish:
Why do we need this maxRetries option? If really so, why we just set it to a default value instead of reading from options?

zryfish:
OK, but I'll suggest switch back to old style options, this one line redis-url is getting more and more complicated, not easy to understand and configure

PR #1290, opened 2019-10-31
fix: privilege escalation

wansir:

**What type of PR is this?**

/kind bug

change cluster role by self is not permitted

### DIFF
--- a/pkg/apiserver/iam/im.go
+++ b/pkg/apiserver/iam/im.go
@@ -160,6 +160,11 @@ func UpdateUser(req *restful.Request, resp *restful.Response) {
 		}
 	}
 
+	if usernameInHeader == user.Username {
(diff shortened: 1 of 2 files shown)


pkg/apiserver/iam/im.go
```diff
@@ -160,6 +160,11 @@ func UpdateUser(req *restful.Request, resp *restful.Response) {
 		}
 	}
 
+	if usernameInHeader == user.Username {
+		// change cluster role by self is not permitted
```
rayzhou2017: how about admin changes itself? Still not allowed? If so, it requires a different admin to change it?

wansir: Yes, cluster role can only be modified by other admin or user manager.

PR #1624, opened 2019-12-02
kubesphere-fix-GetDeployRevision

zehuaiWANG:
**What type of PR is this?**

/kind bug

不属于deployment的rs也可能label相同，我觉得通过label选择器来挑选是不完全对的，最好的方法是使用`metaV1.IsControlledBy(rs, deploy)`来校验。

### DIFF
--- a/pkg/models/revisions/revisions.go
+++ b/pkg/models/revisions/revisions.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/klog"
 
 	"k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


wansir:
This API will be deprecated, revision id is too difficult to understand.

zehuaiWANG:
@wansir More than one interface has been modified in this PR, which is not only because revision id is too difficult to understand a hard problem, but also it's a bug which will case error.

wansir:
Use `metav1.IsControlledBy` to determine that the relationship is correct,  we will refactor this API instead of fixing it.

zehuaiWANG:
@wansir okay, I agree, please close this pr if it's unnecessary

zryfish:
closing per previous comments

PR #1625, opened 2019-12-03
kubesphere-add-default-param

zehuaiWANG:
**What type of PR is this?**
>
/kind cleanup

获取请求参数的地方，可以封装方法来处理默认值的问题，其他地方也可以对应修改。


wansir:
`reverse` parameter can also be obtained through GetBoolValueWithDefault

zehuaiWANG:
@wansir Yes, I'm just give an example. It should be handled like this. There are many code such like this  in the project, but I'm not familiar with this part of the code, so I don't know their specific places.

wansir:
You can commit the code, let me complete it.

zehuaiWANG:
@wansir Yes, you can merge this part of the code and then continue to modify it

Issue #1627, opened 2019-12-03
can't search image from QingCloud Docker Hub


soulseen:
**Describe the bug(描述下问题)**
push an image to QingCloud Docker Hub, but not found this image by kubesphere image search

![image](https://user-images.githubusercontent.com/16031013/70028783-86675d00-15e0-11ea-8daa-0d055bd3ebfc.png)

**Versions used(KubeSphere/Kubernetes的版本)**
KubeSphere: v2.1.0

**To Reproduce(复现步骤)**
1. create a secret 
![image](https://user-images.githubusercontent.com/16031013/70028860-adbe2a00-15e0-11ea-83fc-3f844366c96f.png)

2. creating deployment
3. search image
![image](https://user-images.githubusercontent.com/16031013/70028783-86675d00-15e0-11ea-8daa-0d055bd3ebfc.png)

**Expected behavior(预期行为)**
search image successful.

soulseen:
is not used SSL in search image by default, if need use SSL, when creating secret should add "https://" in registry address.

PR #1628, opened 2019-12-03
remove schema if required and add check SSL

soulseen:
Signed-off-by: zhuxiaoyang <sunzhu@yunify.com>

**What type of PR is this?**

/kind improve

add check SSL, improve parse Image, update unit test

improve https://github.com/kubesphere/kubesphere/issues/1627

### DIFF
--- a/pkg/apiserver/registries/registries.go
+++ b/pkg/apiserver/registries/registries.go
@@ -21,6 +21,7 @@ package registries
 import (
 	"github.com/emicklei/go-restful"
 	"net/http"
+	"strings"
 
 	"kubesphere.io/kubesphere/pkg/models/registries"
 	"kubesphere.io/kubesphere/pkg/server/errors"
@@ -68,6 +69,25 @@ func RegistryImageBlob(request *restful.Request, response *restful.Response) {
 		return
 	}
 
+	// default use ssl
+	checkSSl := func(serverAddress string) bool {
+		if strings.HasPrefix(serverAddress, "http://") {
+			return false
+		} else {
+			return true
+		}
+	}
+
+	if strings.HasPrefix(imageName, "http") {
+		dockerurl, err := registries.ParseDockerURL(imageName)
+		if err != nil {
+			log.Errorf("%+v", err)
+			errors.ParseSvcErr(restful.NewError(http.StatusBadRequest, err.Error()), response)
+			return
+		}
+		imageName = dockerurl.StringWithoutScheme()
+	}
+
 	// parse image
 	image, err := registries.ParseImage(imageName)
 	if err != nil {
@@ -76,8 +96,10 @@ func RegistryImageBlob(request *restful.Request, response *restful.Response) {
 		return
 	}
 
+	useSSL := checkSSl(entry.ServerAddress)
+
 	// Create the registry client.
-	r, err := registries.CreateRegistryClient(entry.Username, entry.Password, image.Domain)
+	r, err := registries.CreateRegistryClient(entry.Username, entry.Password, image.Domain, useSSL)
 	if err != nil {
 		log.Errorf("%+v", err)
 		response.WriteAsJson(&registries.ImageDetails{Status: registries.StatusFailed, Message: err.Error()})
--- a/pkg/models/registries/image.go
+++ b/pkg/models/registries/image.go
@@ -2,9 +2,11 @@ package registries
 
 import (
 	"fmt"
-
 	"github.com/docker/distribution/reference"
-	digest "github.com/opencontainers/go-digest"
+	"github.com/opencontainers/go-digest"
+	log "k8s.io/klog"
+	"net/url"
+	"strings"
 )
 
 // Image holds information about an image.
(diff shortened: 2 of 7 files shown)


pkg/models/registries/image.go
```diff
@@ -30,9 +30,45 @@ func (i *Image) Reference() string {
 	return i.Tag
 }
 
+// Maybe rawurl is of the form scheme:path.
+// (Scheme must be [a-zA-Z][a-zA-Z0-9+-.]*)
+// If so, return scheme, path; else return "", rawurl.
+func Getscheme(rawurl string) (scheme, path string, err error) {
```
runzexia: use https://golang.org/pkg/net/url/#URL ?

soulseen: `Getscheme`  is  a privately function in `net/url` package.
image name is not an URL, here just want to get scheme, and remove it

runzexia: I would prefer to use the standard library implementation, the code now looks complicated.
like:
```
type DockerURL struct {
	*url.URL
}

func (u *DockerURL) StringWithoutScheme() string {
	u.Scheme = ""
	s := u.String()
	return s[2:]
}

func ParseDockerURL(rawurl string) (*DockerURL, error) {
	url, _ := url.Parse(rawurl)
	return &DockerURL{URL: url}, nil
}
```

soulseen: this is a good way, but `rawurl` must contain "http://" or "https://". otherwise it will get a wrong result, such as `rawurl = nginx:latest`

pkg/models/registries/image.go
```diff
@@ -30,9 +31,24 @@ func (i *Image) Reference() string {
...
+
+func (u *DockerURL) StringWithoutScheme() string {
+	u.Scheme = ""
+	s := u.String()
+	return strings.Trim(s, "//")
+}
+
+func ParseDockerURL(rawurl string) (*DockerURL, error) {
+	url, _ := url.Parse(rawurl)
+	return &DockerURL{URL: url}, nil
```
zryfish: so, no error will ever return, why the function signature has error as return value? It's not good.

pkg/models/registries/image_test.go
```diff
@@ -0,0 +1,67 @@
...
+func TestParseImage(t *testing.T) {
+	type image struct {
+		ImageName string
+		ExDomain  string
+		ExTag     string
+		ExPath    string
+	}
+
+	testImages := []image{
+		{ImageName: "dockerhub.qingcloud.com/kubesphere/test:v1", ExDomain: "dockerhub.qingcloud.com", ExTag: "v1", ExPath: "kubesphere/test"},
```
zryfish: need more test cases, what if user inputs  invalid image addresses?

pkg/models/registries/registry_client_test.go
```diff
@@ -37,7 +39,7 @@ func TestCreateRegistryClient(t *testing.T) {
 	}
 
 	testImage := Image{Domain: "docker.io", Path: "library/alpine", Tag: "latest"}
-	r, err := CreateRegistryClient("", "", "docker.io")
+	r, err := CreateRegistryClient("", "", "docker.io", true)
```
zryfish: `docker.io` is default image address, use a constant variable

pkg/apiserver/registries/registries.go
```diff
@@ -68,6 +69,25 @@ func RegistryImageBlob(request *restful.Request, response *restful.Response) {
 		return
 	}
 
+	// default use ssl
+	checkSSl := func(serverAddress string) bool {
```
zryfish: check the schema here, and do another check on schema. I think the first thing here is to properly parse image address, then continue to work on this parse result, this will save lots of schema checking. We can do this refactoring later on version 3.0, but keep in mind.

soulseen: before parse image, should remove protocol. if parse image address first with protocol,  will get an error.

pkg/models/registries/image_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+func TestStringWithoutScheme(t *testing.T) {
+	type testRawurl struct {
+		Rawurl string
+		ExUrl  string
+	}
+	testRawurls := []testRawurl{
+		{"http://dockerhub.qingcloud.com/kubesphere/nginx:v1", "dockerhub.qingcloud.com/kubesphere/nginx:v1"},
+		{"https://dockerhub.qingcloud.com/kubesphere/nginx:v1", "dockerhub.qingcloud.com/kubesphere/nginx:v1"},
+		{"dockerhub.qingcloud.com/kubesphere/nginx:v1", "dockerhub.qingcloud.com/kubesphere/nginx:v1"},
```
runzexia: add test case like 'nginx:latest'

pkg/apiserver/registries/registries.go
```diff
@@ -68,6 +69,25 @@ func RegistryImageBlob(request *restful.Request, response *restful.Response) {
 		return
 	}
 
+	// default use ssl
```
runzexia: Is there a protocol check process in the creation？
Validation during creation easily show error messages

soulseen: do you mean validation during creating secret? but here also will use SSL by default if not have protocol.

runzexia: Yes.
Because the validation of the secret creating process is now based on the docker daemon.
If the user configures the insecure registry in daemon, it may cause addresses without a protocol to be converted to no-ssl requests during creating validation.

soulseen: maybe we could add some descriptions to prompt users to use SSL by default.
@minghaoye

pkg/models/registries/image_test.go
runzexia: why marked this?

pkg/apiserver/registries/registries.go
runzexia: @soulseen Not all users will see the prompt, and it may cause user confusion. This issue should be addressed at the API level.

pkg/models/registries/image_test.go
soulseen: only an image name start with "http" will be used this func.

pkg/models/registries/image_test.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"testing"
+)
+
+func TestParseImage(t *testing.T) {
+	type image struct {
+		ImageName string
+		ExDomain  string
+		ExTag     string
+		ExPath    string
```
zryfish: this is not a good unit test example, you should define what expected result would like, and compare expected result with what you got using reflect

PR #1637, opened 2019-12-05
review node.go

zehuaiWANG:
**What type of PR is this?**

/kind cleanup
/kind bug


wansir:
We can use `node.Spec.Unschedulable` and taints with NoExecute effect instead of this API, all pods will be evicted.
This API is currently not used, we will remove it.

PR #1667, opened 2019-12-13
fix license header

zehuaiWANG:
**What type of PR is this?**
/kind cleanup

fix license header


zehuaiWANG:
修改了一部分，之后我会写一个脚本来批量修改它。

PR #1699, opened 2019-12-22
add golangci-lint script and fix golangci-lint error

zehuaiWANG:
add golangci-lint script and fix golangci-lint error,
do as following:
`go get -u github.com/golangci/golangci-lint/cmd/golangci-lint`
`golangci-lint run -c .golangci.yml --fix cmd/...`
`golangci-lint run -c .golangci.yml --fix pkg/...`


runzexia:
Thanks a lot for contributing to KubeSphere.
I don't recommend introducing formatting tools other than `go tools`.
The introduction of third-party tools requires that project contributors adhere to the specifications brought by this tool, which causes the cost of contributing code to the project to be higher.
So I suggest using only official tools like `go fmt`、` go vet`

zehuaiWANG:
The function of gofmt is very limited. You can check kubernetes, istio and other projects. They all use golangci lint instead of gofmt. @runzexia

PR #1737, opened 2020-01-03
[WIP] API refactor

wansir:

**What type of PR is this?**

/kind api-change
/kind cleanup

1. Refactoring API under openpitrix.io API Group.
2. Refactoring API under resources.kubesphere.io API Group


pkg/kapis/openpitrix/v1/application.go
```diff
@@ -16,112 +16,90 @@
...
-	namespaceName := req.PathParameter("namespace")
-	conditions, err := params.ParseConditions(req.QueryParameter(params.ConditionsParam))
-	orderBy := req.QueryParameter(params.OrderByParam)
-	reverse := params.ParseReverse(req)
-
-	if orderBy == "" {
-		orderBy = "create_time"
-		reverse = true
-	}
+func (h *Handler) ListApplications(request *restful.Request, response *restful.Response) {
```
zryfish: we are moving restful handlers to kapis/openpitrix/v1/handler.go, which is more clear.

pkg/controller/namespace/namespace_controller.go
```diff
@@ -355,13 +355,24 @@ func (r *ReconcileNamespace) checkAndCreateRoleBindings(namespace *corev1.Namesp
 // Create openpitrix runtime
 func (r *ReconcileNamespace) checkAndCreateRuntime(namespace *corev1.Namespace) error {
 
+	op, err := cs.ClientSets().OpenPitrix()
```
zryfish: we shouldn't use clienset any more, it's anti pattern, bad for unit testing. Instead, we should pass op dependency to controller.

pkg/kapis/openpitrix/v1/handler.go
```diff
@@ -1,76 +1,967 @@
...
+
+	app, err := h.openpitrix.DescribeApplication(namespace, clusterId)
+
+	if status.Code(err) == codes.NotFound {
+		resp.WriteHeaderAndEntity(http.StatusNotFound, errors.Wrap(err))
+		return
+	}
+
+	if err != nil {
+		resp.WriteHeaderAndEntity(http.StatusInternalServerError, errors.Wrap(err))
```
zryfish: use api.HandleInternalError instead

pkg/kapis/openpitrix/v1/handler.go
```diff
@@ -1,76 +1,967 @@
...
+	resp.WriteEntity(app)
+	return
+}
+
+func (h *openpitrixHandler) CreateApplication(req *restful.Request, resp *restful.Response) {
+	namespace := req.PathParameter("namespace")
+	var createClusterRequest openpitrix.CreateClusterRequest
+	err := req.ReadEntity(&createClusterRequest)
+	if err != nil {
+		resp.WriteHeaderAndEntity(http.StatusBadRequest, errors.Wrap(err))
```
zryfish: use api.HandleBadRequest instead

pkg/kapis/openpitrix/v1/handler.go
```diff
@@ -1,76 +1,967 @@
...
+			Type:       createRepoRequest.Type,
+			Url:        createRepoRequest.URL,
+			Credential: createRepoRequest.Credential,
+		}
+		result, err = h.openpitrix.ValidateRepo(validateRepoRequest)
+	} else {
+		result, err = h.openpitrix.CreateRepo(createRepoRequest)
+	}
+
+	if status.Code(err) == codes.InvalidArgument {
```
zryfish: what will happen to status.Code(err) if err == nil ? I think it's better to be rewritten like following:
```
if err != nil {
  if status.Code(err) == codes.InvalidArgument {
     ...
     return
  }
  api.HandleInternalError(resp, err)
  return
}
```

pkg/kapis/openpitrix/v1/register.go
```diff
@@ -21,34 +21,36 @@ import (
...
 var GroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1"}
 
-var (
-	WebServiceBuilder = runtime.NewContainerBuilder(addWebService)
-	AddToContainer    = WebServiceBuilder.AddToContainer
-)
+func AddToContainer(c *restful.Container, k8s k8s.Client, op op.Client) error {
 
-func addWebService(c *restful.Container) error {
+	if k8s == nil || op == nil {
```
zryfish: since k8s and op is required for openpitrix, no need to nullability here, dependencies is guaranteed to not be null

pkg/controller/namespace/namespace_controller.go
wansir: I will create a new PR to refactor namespace controller.

PR #1750, opened 2020-01-09
refactor resources API

### LINKED ISSUE #1722: Resource API returns duplicate data
**Describe the bug(描述下问题)**
Resource API returns duplicate data when having the same creation time.

**Versions used(KubeSphere/Kubernetes的版本)**
KubeSphere: v2.1.0

image

image

https://kubesphere.com.cn/forum/d/383-bug

wansir:

**What type of PR is this?**

/kind bug

Sort resources by name when the creation time is the same.

Fixes #1722

### DIFF
--- a/pkg/models/resources/resources.go
+++ b/pkg/models/resources/resources.go
@@ -19,7 +19,9 @@ package resources
 
 import (
 	"fmt"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
+	"kubesphere.io/kubesphere/pkg/constants"
 	"kubesphere.io/kubesphere/pkg/models"
 	"kubesphere.io/kubesphere/pkg/server/params"
 	"kubesphere.io/kubesphere/pkg/utils/sliceutil"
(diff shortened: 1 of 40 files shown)


pkg/models/resources/resources.go
```diff
@@ -160,13 +162,18 @@ func ListResources(namespace, resource string, conditions *params.Conditions, or
...
-		if i >= offset && (limit == -1 || len(items) < limit) {
-			items = append(items, injector.addExtraAnnotations(item))
-		}
+	totalCount := len(result)
+
+	if limit == -1 || limit+offset > totalCount {
+		limit = totalCount - offset
+	}
+
+	result = result[offset : offset+limit]
```
zryfish: what if offset > totalCount?

pkg/server/params/params.go
```diff
@@ -45,7 +44,9 @@ func ParsePaging(paging string) (limit, offset int) {
 	return
```
zryfish: why ignore error here? what would happen if there is an error during parsing?

wansir: Use default values when parsing failed, limit and page must be defined at the same time.

zryfish: I believe atoi will return 0 when there is an error

wansir: Regular expression `\d+` defines that only numbers are allowed here, i think this error can be ignored.

PR #756, opened 2019-09-17
refactor monitor module

huanggze:
Signed-off-by: huanggze <loganhuang@yunify.com>

**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>

/kind api-change
> /kind bug

/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:
cleanup monitor module

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for reviewers**:
```
1. remove: /nodes/{node}/pods/{pod}/containers
2. add: /workspaces/{workspace}/namespaces
3. deprecate workload metrics:

  workload_pod_cpu_usage             > workload_cpu_usage
  workload_pod_memory_usage_wo_cache > workload_memory_usage_wo_cache
  workload_pod_net_bytes_transmitted > workload_net_bytes_transmitted
  workload_pod_net_bytes_received    > workload_net_bytes_received

4. deprecate workspace metrics for statistics: INCOMPATIBLE

  workspace_all_organization_count   > cluster_workspace_count
  workspace_all_account_count        > cluster_account_count
  workspace_all_project_count        > cluster_namespace_count
  workspace_all_devops_project_count > cluster_devops_project_count

5. remove namespace resource quota metrics (except for cpu.limit and mem.limit) 
  Note: Either unused or incorrect, waiting for redesign.

6. refine response for node metrics: INCOMPATIBLE

  GET /nodes/{node}
  {
    ...
    {
      "metric": {
        "address": [
         {
           "type": "InternalIP",
           "address": "192.168.2.2"
         },
         {
           "type": "Hostname",
           "address": "ks-allinone"
         }
        ],
        "resource_name": "ks-allinone"
      },
      "value": [
        1568703384.343,
        "16"
      ]
    }
  }

====>

  GET /nodes/{node}
  {
    ...
    {
      "metric": {
        "node_ip": "192.168.2.2",
        "node_role": "master",
        "resource_name": "ks-allinone"
      },
      "value": [
        1568703384.343,
        "16"
      ]
    }
  }

7. better error handling
```

**Additional documentation, usage docs, etc.**:

```docs

```

PR #1751, opened 2020-01-09
[WIP] monitoring refactor

huanggze:
**What type of PR is this?**
>

/kind api-change
> /kind bug

 /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

- Move to KubeSphere 3.0
- Finish decoupling work: the monitoring module won't denpenden on Prometheus implement
- Expose KubeSphere's self merics in Prometheus format (TODO ?): We still manually construct the following metrics `cluster_account_count`, `workspace_member_count`, `workspace_role_count`, etc.
- **[API Change]** Remove deprecated metrics since 2.1.0 https://github.com/kubesphere/kubesphere/pull/756
- Add test cases (TODO)
- Move metrics [promql templates](https://github.com/kubesphere/kubesphere/blob/15cfa2010c27ad8aaceb17b7f975b7915f93b9ec/pkg/simple/client/monitoring/prometheus/promql_templates.go#L28) to a ConfigMap as suggested by @zryfish  (TODO)
- **[API Change]** Don't append extra label `resouce_name` to Prometheus metric result:
We once hard-code `resouce_name` to each metric result like below
```json
{
    "result":[
        {
            "metric":{
                "node":"ks-allinone",
                "resource_name":"ks-allinone"
            },
            "value":[
                1583313133.145,
                "8"
            ]
        }
    ]
}
```
`resouce_name` is copied from existing label (e.g. node, workload) and used as a unique identifier for sorting. Since the SortMetrics function has been refactored, we may remove out this manual modification.

Fixes #

```
```

```docs

```


pkg/models/metrics/metrics.go
```diff
@@ -19,785 +19,16 @@
...
-	return &Response{
-		MetricsLevel: MonitorLevelContainer,
-		Results:      apiResponse,
-	}
-}
-
-func GetPVCMetrics(params RequestParams) *Response {
-	client, err := cs.ClientSets().Prometheus()
+func GetMetrics(opt monitoring.QueryOptions) (*monitoring.MetricsGroup, error) {
+	client, err := cs.ClientSets().MonitoringClient()
```
zryfish: dependency should be passed from caller, we shouldn't use client factory any more

pkg/models/metrics/namespaces.go
```diff
@@ -18,8 +18,11 @@
 package metrics
 
 import (
-	"net/url"
+	"k8s.io/klog"
```
zryfish: need unit test code on this module

pkg/simple/client/monitoring/interface.go
```diff
@@ -1,39 +1,82 @@
...
 
+type MetricsValue struct {
+	Metadata map[string]string `json:"metric,omitempty" description:"time series labels"`
+	Sample   Point             `json:"value,omitempty" description:"time series, values of vector type"`
+	Series   []Point           `json:"values,omitempty" description:"time series, values of matrix type"`
 }
 
 // Interface defines all the abstract behaviors of monitoring
 type Interface interface {
+	GetMetrics(QueryOptions) *MetricsGroup
```
zryfish: Personally, I think `GetMetrics` is too general, now well designed,  and `QueryOptions` contains too much information. For example, if I want to query metrics of level `MetricsLevelWorkspace` , which fields of `QueryOptions` should I have to set, and how to set ? It's not obvious now.

pkg/simple/client/monitoring/prometheus/metrics_rules.go
```diff
@@ -0,0 +1,418 @@
...
+
+const (
+	StatefulSet = "StatefulSet"
+	DaemonSet   = "DaemonSet"
+	Deployment  = "Deployment"
+)
+
+var metricsPromqlMap = map[monitoring.MetricsLevel]map[string]string{
+
+	monitoring.MetricsLevelCluster: {
```
zryfish: Can we use a configuration file and load them at startup, since these are static rules, using configuration would make it easy to change.

pkg/simple/client/monitoring/prometheus/prometheus.go
```diff
@@ -0,0 +1,147 @@
...
+		queryType = "query_range"
+		params.Set("start", q.start)
+		params.Set("end", q.end)
+	}
+	params.Set("step", q.interval)
+	params.Set("query", q.promql)
+
+	u := fmt.Sprintf("%s/api/v1/%s?%s", q.endpoint, queryType, params.Encode())
+
+	var m monitoring.Metrics
```
zryfish: It's better to return both metrics and error, not one big struct.

pkg/simple/client/monitoring/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,90 @@
...
+						Data:        nil,
+						ErrorType:   monitoring.ErrorTypeUnavailable,
+					},
+				},
+			},
+		},
+	}
+
+	opt := &Options{
+		Endpoint:          "http://localhost:9090",
```
zryfish: Isn't this piece of unit test code depending on Prometheus? Then your test code can't run standalone, I think the solution here is to using fake data, or using a mocking Prometheus.

benjaminhuo:
/lgtm
The coredns metrics is not used anywhere for now, I'm thinking maybe we can remove it if not used in 3.0 later.

PR #1757, opened 2020-01-13
Add API for setting default storage class

### LINKED ISSUE #1656: The default storage class should not be set more than 1
I guess KubeSphere should not allow users to set 2 default StorageClass in UI, when you set one as default, the other one should be unset automaticlly.
image

wnxn:
Signed-off-by: Xin Wang <wileywang@yunify.com>

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Fixes #1656

```
```

```docs
Usage:
PATCH /kapis/resources.kubesphere.io/v1alpha2/storageclasses/{storageclassname}
Body {"metadata":{"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}
If succeed, return code 200 and return default storage class definition.
```

### DIFF
--- a/pkg/apiserver/resources/storage.go
+++ b/pkg/apiserver/resources/storage.go
@@ -20,8 +20,10 @@ package resources
 import (
 	"github.com/emicklei/go-restful"
 	"k8s.io/api/core/v1"
+	"k8s.io/klog"
 	"net/http"
 
+	storagev1 "k8s.io/api/storage/v1"
 	"kubesphere.io/kubesphere/pkg/models/storage"
 	"kubesphere.io/kubesphere/pkg/server/errors"
 )
@@ -68,4 +70,27 @@ func GetPvcListBySc(request *restful.Request, response *restful.Response) {
 	}
 
 	response.WriteAsJson(result)
+}
+
+func PatchStorageClass(request *restful.Request, response *restful.Response) {
+	scObj := &storagev1.StorageClass{}
+	err := request.ReadEntity(scObj)
+	if err != nil {
+		klog.Errorf("read entity error: %s", err.Error())
+		response.WriteHeaderAndEntity(http.StatusBadRequest, errors.Wrap(err))
+		return
+	}
+	klog.V(5).Infof("succeed to read entity %v", scObj)
+	scName := request.PathParameter("storageclass")
+	if scObj.Annotations[storage.IsDefaultStorageClassAnnotation] == "true" {
+		// Set default storage class
+		sc, err := storage.SetDefaultStorageClass(scName)
+		if err != nil {
+			response.WriteHeaderAndEntity(http.StatusInternalServerError, errors.Wrap(err))
+			return
+		}
+		response.WriteEntity(sc)
+		return
+	}
+	response.WriteEntity(errors.None)
 }
--- a/pkg/kapis/resources/v1alpha2/register.go
+++ b/pkg/kapis/resources/v1alpha2/register.go
@@ -22,6 +22,7 @@ import (
 	"github.com/emicklei/go-restful-openapi"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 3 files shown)


pkg/models/storage/storage.go
```diff
@@ -27,6 +28,11 @@ import (
 	"kubesphere.io/kubesphere/pkg/informers"
 )
 
+const (
+	IsDefaultStorageClassAnnotation     = "storageclass.kubernetes.io/is-default-class"
```
zryfish: Unnecessary exported constants

pkg/models/storage/storage.go
```diff
@@ -27,6 +28,11 @@ import (
...
+	BetaIsDefaultStorageClassAnnotation = "storageclass.beta.kubernetes.io/is-default-class"
+)
+
 type ScMetrics struct {
```
zryfish: same here, unnecessary exported

zryfish: remember to remove unused function `init()` down below

pkg/models/storage/storage.go
```diff
@@ -102,3 +108,32 @@ func GetScList() ([]*storageV1.StorageClass, error) {
...
+	k8sClient := client.ClientSets().K8s().Kubernetes()
+	var defaultSc *storageV1.StorageClass
+	for _, sc := range scList {
+		delete(sc.Annotations, IsDefaultStorageClassAnnotation)
+		delete(sc.Annotations, BetaIsDefaultStorageClassAnnotation)
+		if sc.Name == defaultScName {
+			sc.Annotations[IsDefaultStorageClassAnnotation] = "true"
+			defaultSc = sc
+		}
+		_, err := k8sClient.StorageV1().StorageClasses().Update(sc)
```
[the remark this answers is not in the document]
wnxn: > you are updating every single storageclass, some of them may don't have default storage class annotation. Although, this is probably harmless, we better only update those we need to.

We need to ensure all storage classes not set the default storage class annotation except the one we want to set.

pkg/kapis/resources/v1alpha2/register.go
```diff
@@ -257,7 +258,13 @@ func addWebService(c *restful.Container) error {
 		Returns(http.StatusOK, ok, status.WorkLoadStatus{}).
 		To(workloadstatuses.GetNamespacedAbnormalWorkloads))
 
-	c.Add(webservice)
+	webservice.Route(webservice.PUT("/defaultstorageclasses/{storageclass}").
```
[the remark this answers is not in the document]
wnxn: The API not only for patching one storage class but also ensuring only one default storage class in one Kubernetes cluster.

pkg/models/storage/storage.go
wnxn: OK, I will fix it.

pkg/models/storage/storage.go
wnxn: > remember to remove unused function `init()` down below

Sure

pkg/models/storage/storage.go
wnxn: IsDefaultStorageClassAnnotation is used at PatchStorageClass.

pkg/apiserver/resources/storage.go
```diff
@@ -69,3 +71,26 @@ func GetPvcListBySc(request *restful.Request, response *restful.Response) {
 
 	response.WriteAsJson(result)
 }
+
+func PatchStorageClass(request *restful.Request, response *restful.Response) {
+	scObj := &storagev1.StorageClass{}
+	err := request.ReadEntity(scObj)
+	if err != nil {
+		klog.V(5).Infof("read entity error: %s", err.Error())
```
zryfish: print error log all the time

wnxn: > print error log all the time

Done

Issue #1501, opened 2019-11-14
Refine docker secret check


runzexia:
ref https://kubesphere.io/forum/d/158-source-to-image/5

should allow scheme in docker secret
good address:
`harbor.xx.xxx/xx/xxx`
`http://harbor.xx.xxx/xx/xxx`
`https://harbor.xx.xxx/xx/xxx`

/area api
/assign 
/kind bug
/milestone 2.1.1

PR #1831, opened 2020-02-10
verify secret

soulseen:
Signed-off-by: zhuxiaoyang <sunzhu@yunify.com>

fix https://github.com/kubesphere/kubesphere/issues/1501

### DIFF
--- a/pkg/models/registries/registries.go
+++ b/pkg/models/registries/registries.go
@@ -28,6 +28,8 @@ import (
 	"k8s.io/klog"
 	log "k8s.io/klog"
 	"kubesphere.io/kubesphere/pkg/informers"
+	"net/http"
+	"strings"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/models/registries/registries.go
```diff
@@ -68,6 +70,23 @@ func RegistryVerify(authInfo AuthInfo) error {
 		ServerAddress: authInfo.ServerHost,
 	}
 
+	// Sometimes request DockerHub V2 API will timeout in China. Skipped.
```
runzexia: It would be nice if you could add some test cases.
like 
`docker.io "" ""`
`dockerhub.qingcloud.com guest guest`

soulseen: added some test registry that can login with guest.

PR #1836, opened 2020-02-11
fix: remove owner reference when unbind workspace

wansir:

**What type of PR is this?**

/kind bug

Remove owner reference when delete workspace label, prevent  cascading delete.

### DIFF
--- a/pkg/controller/namespace/namespace_controller.go
+++ b/pkg/controller/namespace/namespace_controller.go
@@ -170,9 +170,15 @@ func (r *ReconcileNamespace) Reconcile(request reconcile.Request) (reconcile.Res
 
 	if !controlledByWorkspace {
 
-		err = r.deleteRoleBindings(instance)
+		if err = r.unbindWorkspace(instance); err != nil {
+			return reconcile.Result{}, err
+		}
 
-		return reconcile.Result{}, err
+		if err = r.deleteRoleBindings(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		return reconcile.Result{}, nil
 	}
 
 	if err = r.checkAndBindWorkspace(instance); err != nil {
@@ -220,6 +226,7 @@ func (r *ReconcileNamespace) checkAndCreateRoles(namespace *corev1.Namespace) er
 					klog.Error(err)
 					return err
 				}
+				found = role
 			} else {
 				klog.Error(err)
 				return err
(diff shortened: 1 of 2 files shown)


pkg/controller/namespace/namespace_controller.go
```diff
@@ -169,10 +169,19 @@ func (r *ReconcileNamespace) Reconcile(request reconcile.Request) (reconcile.Res
 	}
 
 	if !controlledByWorkspace {
+		err = r.unbindWorkspace(instance)
```
zryfish: To make it shorter and more clear, I suggest
```suggestion
		if err := r.unbindWorkspace(instance); err != nil {
		   return reconcile.Result{}, err
		}
```

pkg/controller/namespace/namespace_controller.go
```diff
@@ -169,10 +169,19 @@ func (r *ReconcileNamespace) Reconcile(request reconcile.Request) (reconcile.Res
...
+
+		if err != nil {
+			return reconcile.Result{}, err
+		}
 
 		err = r.deleteRoleBindings(instance)
```
zryfish:
```suggestion
		if err := r.deleteRoleBindings(instance); err != nil {
		    return reconcile.Result{}, err
		}
```

pkg/controller/namespace/namespace_controller.go
```diff
@@ -451,6 +461,21 @@ func (r *ReconcileNamespace) deleteRuntime(namespace *corev1.Namespace) error {
 
 	return nil
 }
+func (r *ReconcileNamespace) unbindWorkspace(instance *corev1.Namespace) error {
+	changed := false
+	for i := 0; i < len(instance.OwnerReferences); i++ {
```
zryfish: this is not the straightforward way, not easy to read, I suggest make a copy without unwanted references, and then assign back to instance

pkg/models/workspaces/workspaces.go
```diff
@@ -156,8 +156,9 @@ func DeleteWorkspaceRoleBinding(workspace, username string, role string) error {
 	}
 	workspaceRoleBinding = workspaceRoleBinding.DeepCopy()
 
-	for i, v := range workspaceRoleBinding.Subjects {
-		if v.Kind == v1.UserKind && v.Name == username {
+	for i := 0; i < len(workspaceRoleBinding.Subjects); i++ {
```
zryfish: same here

PR #1879, opened 2020-02-14
fix: devops project sorted in ascending order by default

wansir:

**What type of PR is this?**

/kind bug

devops project sorted in ascending order by default


runzexia:
Change this api will cause api change.
And there is no reverse option for the list of devops projects on the UI.

PR #1960, opened 2020-03-18
implement authorizer filter

wansir:

**What type of PR is this?**

> /kind api-change
> /kind cleanup

implement authorizer filter.


pkg/apiserver/authorization/authorizer/interfaces.go
```diff
@@ -0,0 +1,190 @@
...
+	GetCluster() string
+
+	// The workspace of the object, if a request is for a REST object.
+	GetWorkspace() string
+
+	// The namespace of the object, if a request is for a REST object.
+	GetNamespace() string
+
+	// The devops project of the object, if a request is for a REST object.
+	GetDevopsProject() string
```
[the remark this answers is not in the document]
wansir: Devops project is the same level as namespace in system architecture, even if it is defined through CRD, the devops project is still a level under the cluster. How to distinguish it from namespace? For example, pipline is not a resource in namespace. We need to know the level of the resource when authenticating. I think the API of list pipeline should be defined as follows, `/kapis/devops.kubesphere.io/v1alpha3/devops/{devops}/piplines`. Even though pipeline CRD is still a cluster level resource, we need to define the Devops project level. I think we can use a special (fake) namespace to represent the devops project.

pkg/apiserver/authorization/path/path_test.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+func TestNewAuthorizer(t *testing.T) {
+	tests := []struct {
+		name                       string
+		excludedPaths              []string
+		allowed, denied, noOpinion []string
+		wantErr                    bool
+	}{
+		{"inner star", []string{"/foo*bar"}, nil, nil, nil, true},
```
zryfish: Some real request examples will be better

pkg/apiserver/request/requestinfo_test.go
```diff
@@ -0,0 +1,84 @@
...
+		url                       string
+		method                    string
+		expectedErr               error
+		expectedVerb              string
+		expectedResource          string
+		expectedIsResourceRequest bool
+		expectedCluster           string
+	}{
+		{
+			name:                      "login",
```
zryfish: More test cases

pkg/apiserver/authorization/authorizerfactory/opa.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	// check platform role policy rules
+	if a, r, e := roleAuthorize(platformRole, a); a == authorizer.DecisionAllow {
+		return a, r, e
+	}
+
+	// it's not in cluster resource, permission denied
+	// TODO declare implicit cluster info in request Info
+	if a.GetCluster() == "" {
```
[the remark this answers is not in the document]
wansir: Even if there is no clear cluster name in the request path, we can also supplement the implicit cluster information when parsing the request info. If the cluster name is empty, means the host cluster, there is no difference between multi cluster and single cluster.

pkg/apiserver/authorization/authorizerfactory/opa.go
```diff
@@ -0,0 +1,128 @@
...
+		return authorizer.DecisionDeny, "permission undefined", nil
+	}
+
+	clusterRole, err := o.am.GetClusterRole(a.GetCluster(), a.GetUser().GetName())
```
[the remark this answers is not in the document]
wansir: It means that the user cannot obtain the resources in this cluster, some error will be returned,  and the request will be denied.

pkg/apiserver/authorization/authorizerfactory/opa.go
```diff
@@ -0,0 +1,128 @@
...
+		return authorizer.DecisionDeny, "", err
+	}
+
+	// check cluster role policy rules
+	if a, r, e := roleAuthorize(clusterRole, a); a == authorizer.DecisionAllow {
+		return a, r, e
+	}
+
+	// it's not in cluster resource, permission denied
+	if a.GetWorkspace() == "" && a.GetNamespace() == "" && a.GetDevopsProject() == "" {
```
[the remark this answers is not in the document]
wansir: In the previous step, cluster level permissions have been checked.If it is not  workspaced resource namespaced resource or  devops project resouce, this request can be denied.

pkg/apiserver/authorization/path/path.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"kubesphere.io/kubesphere/pkg/apiserver/authorization/authorizer"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// NewAuthorizer returns an authorizer which accepts a given set of paths.
+// Each path is either a fully matching path or it ends in * in case a prefix match is done. A leading / is optional.
+func NewAuthorizer(alwaysAllowPaths []string) (authorizer.Authorizer, error) {
```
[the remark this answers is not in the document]
wansir: The code is copied from k8s, alwaysAllowPaths means some request path  do not require authentication, just like login request, of course these are mostly non-resourced  requests.

pkg/api/auth/token/user.go
```diff
@@ -0,0 +1,30 @@
...
+	GetName() string
+
+	// UID
+	GetUID() string
+
+	// Groups
+	GetGroups() []string
+}
+
+type AuthUser struct {
```
zryfish: Why we have AuthUser here since we already have  `iam.User`

pkg/api/auth/token/jwt.go
```diff
@@ -13,8 +12,9 @@ const DefaultIssuerName = "kubesphere"
 var errInvalidToken = errors.New("invalid token")
 
 type claims struct {
-	Username string `json:"username"`
-	UID      string `json:"uid"`
+	Username string   `json:"username"`
+	UID      string   `json:"uid"`
+	Groups   []string `json:"groups"`
```
zryfish: If we issue a token include groups, what will happen to the issued token if user groups changed?

pkg/api/auth/token/user.go
```diff
@@ -0,0 +1,30 @@
+package token
+
+type User interface {
+	// Name
+	GetName() string
```
zryfish: No need to put explicit `Get` here, the method here behaves like an attribute to User, it's more go style

pkg/apiserver/apiserver.go
```diff
@@ -174,27 +177,27 @@ func (s *APIServer) buildHandlerChain() {
...
 	handler := s.Server.Handler
+
 	handler = filters.WithKubeAPIServer(handler, s.KubernetesClient.Config(), &errorResponder{})
 	handler = filters.WithMultipleClusterDispatcher(handler, dispatch.DefaultClusterDispatch)
-	handler = filters.WithAuthorization(handler, authorizerfactory.NewAlwaysAllowAuthorizer())
 
-	authn := union.New(&authenticationrequest.AnonymousAuthenticator{}, bearertoken.New(jwttoken.NewTokenAuthenticator(s.CacheClient, s.AuthenticateOptions.JwtSecret)))
-	handler = filters.WithAuthentication(handler, authn, failed)
-	handler = filters.WithRequestInfo(handler, requestInfoResolver)
+	excludedPaths := []string{"/oauth/authorize", "/oauth/token"}
```
zryfish: It's not a good a idea to hardcode excluded paths here, we need a better solution

pkg/apiserver/authentication/authenticators/jwttoken/jwt_token.go
```diff
@@ -35,23 +34,19 @@ func (t *tokenAuthenticator) AuthenticateToken(ctx context.Context, token string
 		return nil, false, err
 	}
 
-	_, err = t.cacheClient.Get(tokenKeyForUsername(providedUser.Name(), token))
```
zryfish: Why we don't need this piece of  code any more

pkg/apiserver/authorization/authorizerfactory/opa.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/open-policy-agent/opa/rego"
+	"kubesphere.io/kubesphere/pkg/apiserver/authorization/authorizer"
+	"kubesphere.io/kubesphere/pkg/models/iam/am"
+)
+
+type opaAuthorizer struct {
+	am am.AccessManagementInterface
+}
+
+func (o *opaAuthorizer) Authorize(attr authorizer.Attributes) (authorized authorizer.Decision, reason string, err error) {
```
zryfish: Please add more comments and examples here to explain how this will work

pkg/apiserver/filters/dispatch.go
```diff
@@ -23,9 +24,11 @@ func WithMultipleClusterDispatcher(handler http.Handler, dispatch dispatch.Dispa
 			return
 		}
 
-		if info.Cluster == "" {
+		if info.Cluster == "host-cluster" || info.Cluster == "" {
```
zryfish: empty cluster name implies no dispatching, "host-cluster" is unnecessary

pkg/apiserver/filters/kubeapiserver.go
```diff
@@ -33,6 +35,8 @@ func WithKubeAPIServer(handler http.Handler, config *rest.Config, failed proxy.E
 			s := *req.URL
 			s.Host = kubernetes.Host
 			s.Scheme = kubernetes.Scheme
+			// remove cluster path
+			s.Path = strings.Replace(s.Path, fmt.Sprintf("/clusters/%s", info.Cluster), "", 1)
```
zryfish: Multiple cluster requests are dispatched to designated clusters before sent to kube-apiserver

pkg/apiserver/request/requestinfo_test.go
```diff
@@ -0,0 +1,178 @@
...
+			t.Fatal(err)
+		}
+		requestInfo, err := requestInfoResolver.NewRequestInfo(req)
+
+		if err != nil {
+			if test.expectedErr != err {
+				t.Errorf("%s: expected error %v, actual %v", test.name, test.expectedErr, err)
+			}
+		} else {
+			if test.expectedVerb != "" && test.expectedVerb != requestInfo.Verb {
```
zryfish: `test.expectedVerb != ""` what's this for

pkg/kapis/iam/v1alpha2/register.go
```diff
@@ -41,135 +38,59 @@ const groupName = "iam.kubesphere.io"
...
-		Returns(http.StatusOK, api.StatusOK, iamv1alpha2.ListUserResponse{}).
-		Metadata(restfulspec.KeyOpenAPITags, []string{constants.IdentityManagementTag}))
-	ws.Route(ws.GET("/users/{user}/roles").
-		To(handler.ListUserRoles).
-		Doc("Retrieve all the roles that are assigned to the specified user.").
-		Param(ws.PathParameter("user", "username")).
-		Returns(http.StatusOK, api.StatusOK, []*rbacv1.Role{}).
-		Metadata(restfulspec.KeyOpenAPITags, []string{constants.AccessManagementTag}))
+	// implemented by create CRD object.
+	ws.Route(ws.POST("/users"))
```
zryfish: Unfinished?

pkg/api/auth/token/user.go
wansir: If the UID field is defined, the UID method cannot be defined, https://github.com/kubernetes/apiserver/blob/master/pkg/authentication/user/user.go#L53-L63

pkg/api/auth/token/jwt.go
wansir: TokenReview API needs to return group information, it's unnecessary.

pkg/apiserver/apiserver.go
wansir: These two API paths are fixed, if we need dynamic configuration, we need to add  some configurations like caddy.

pkg/apiserver/filters/dispatch.go
wansir: Implied cluster info will be completed when parsing request info, dispatching is not needed if cluster is empty or request host-cluster's resources .   https://github.com/kubesphere/kubesphere/pull/1960/files#diff-07d66560df8f80f564e18b05364c82d1R130

pkg/apiserver/filters/kubeapiserver.go
wansir: There is no difference between host cluster and other clusters. `/api/v1/clusters/host-cluster/nodes` Equivalent to `/api/v1/nodes`, even if it's not recommended.

pkg/apiserver/request/requestinfo_test.go
wansir: if expectedVerb is not defined, it's unnecessary to make a comparison.

pkg/kapis/iam/v1alpha2/register.go
wansir: This API has been deprecated, and then users are created through CRD.

PR #1961, opened 2020-03-19
add test for s2i controller & devops CRD

runzexia:
/kind api-change
/area devops
/kind cleanup


cmd/controller-manager/app/controllers.go
```diff
@@ -28,6 +28,7 @@ import (
 	"kubesphere.io/kubesphere/pkg/controller/virtualservice"
 	"kubesphere.io/kubesphere/pkg/informers"
 	"kubesphere.io/kubesphere/pkg/simple/client/k8s"
+	fakeS3 "kubesphere.io/kubesphere/pkg/simple/client/s3/fake"
```
zryfish: conventionally, we use all lower-cased letters as import alias

pkg/apis/devops/v1alpha3/devopsproject_types.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// DevOpsProject is the Schema for the devopsprojects API
+// +kubebuilder:resource:categories="devops",scope="Cluster"
+// +k8s:openapi-gen=true
+type DevOpsProject struct {
```
zryfish: Since DevOpsProject is in group devops.kubesphere.io, it seems a little bit unnecessary for extra devops in the name.  Or you are trying to make it more specific to distinguish  with other projects?

pkg/apis/devops/v1alpha3/devopsproject_types.go
```diff
@@ -0,0 +1,73 @@
...
+const DevOpsProjectFinalizerName = "devopsproject.finalizers.kubesphere.io"
+
+const (
+	ResourceKindDevOpsProject     = "DevOpsProject"
+	ResourceSingularDevOpsProject = "devopsproject"
+	ResourcePluralDevOpsProject   = "devopsprojects"
+)
+
+// DevOpsProjectSpec defines the desired state of DevOpsProject
+type DevOpsProjectSpec struct {
```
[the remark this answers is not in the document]
runzexia: Yes, there is no spec yet, because it is an isolation concept similar to ns.
We need some specs (creator, etc.) similar to the workspace, I want to deal with that part after more discussion.

pkg/apis/devops/v1alpha3/devopsproject_types.go
runzexia: I think that project is too generic. When using kubectl, it may have the same name as other CRDs (although they are in different apigroups).

pkg/controller/s2irun/s2irun_controller.go
```diff
@@ -24,6 +24,10 @@ import (
 	devopslisters "kubesphere.io/kubesphere/pkg/client/listers/devops/v1alpha1"
 )
 
+/**
+s2irun-controller used to handle s2irun's delete logic.
+s2irun creation and operation provided by s2ioperator
```
zryfish: so we need s2irun-controller and s2ioperator, that's wired. How about move s2ioperator into controller-manager, but also independent s2ioperator so people who need only s2irun can deploy without whole kubesphere package.

pkg/controller/s2irun/s2irun_controller.go
```diff
@@ -213,6 +218,11 @@ func (c Controller) syncHandler(key string) error {
 	return nil
 }
 
+/**
+	DeleteS2iBinary mainly cleans up two parts of S2iBinary
+	1. s2ibinary bound to s2irun
+    2. s2ibinary that has been created for more than 24 hours but has not been used
```
zryfish: formatting

pkg/simple/client/devops/fake/fakedevops.go
```diff
@@ -1,206 +1,290 @@
...
-	var fakeData FakeDevops
+func NewFakeDevops(data map[string]interface{}) *Devops {
+	var fakeData Devops
 	fakeData.Data = data
 	return &fakeData
 }
 
 // Pipelinne operator interface
-func (d *FakeDevops) GetPipeline(projectName, pipelineName string, httpParameters *devops.HttpParameters) (*devops.Pipeline, error) {
+func (d *Devops) GetPipeline(projectName, pipelineName string, httpParameters *devops.HttpParameters) (*devops.Pipeline, error) {
```
zryfish: Why FakeDevops has all these no-op functions? How can we use this in unit test?

pkg/apis/devops/v1alpha3/devopsproject_types.go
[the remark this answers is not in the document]
runzexia: DevOps project is a cluster-level resource.
Pipelines are namespace-level resources.
When we create a DevOps project, we will create an admin namespace for the project, and the pipeline will be stored in the admin namespace.

pkg/controller/s2irun/s2irun_controller.go
runzexia: fix

pkg/simple/client/devops/fake/fakedevops.go
runzexia: When we create a devops project, we will create an admin namespace for the project, and the pipeline will be stored in the admin namespace.

pkg/controller/s2irun/s2irun_controller.go
runzexia: Currently s2ioperator has a separate e2e test, but ks does not.
Migration also leads to the need to maintain two parts of the code.
There are two parts of the controller because for b2i, the ks API provides the function of uploading artifacts, while s2ioperator itself does not support uploading files, he will only use the link for file download.

cmd/ks-apiserver/app/options/options.go
```diff
@@ -17,10 +17,9 @@ import (
 	"kubesphere.io/kubesphere/pkg/simple/client/ldap"
 	esclient "kubesphere.io/kubesphere/pkg/simple/client/logging/elasticsearch"
 	"kubesphere.io/kubesphere/pkg/simple/client/monitoring/prometheus"
-	"kubesphere.io/kubesphere/pkg/simple/client/mysql"
 	"kubesphere.io/kubesphere/pkg/simple/client/openpitrix"
 	"kubesphere.io/kubesphere/pkg/simple/client/s3"
-	fakeS3 "kubesphere.io/kubesphere/pkg/simple/client/s3/fake"
+	fakes3 "kubesphere.io/kubesphere/pkg/simple/client/s3/fake"
```
zryfish, line 22: why use fake client?

pkg/apiserver/apiserver.go
```diff
@@ -135,18 +136,20 @@ func (s *APIServer) PrepareRun() error {
...
 	urlruntime.Must(monitoringv1alpha2.AddToContainer(s.container, s.KubernetesClient, s.MonitoringClient))
 	urlruntime.Must(openpitrixv1.AddToContainer(s.container, s.InformerFactory, s.OpenpitrixClient))
 	urlruntime.Must(operationsv1alpha2.AddToContainer(s.container, s.KubernetesClient.Kubernetes()))
 	urlruntime.Must(resourcesv1alpha2.AddToContainer(s.container, s.KubernetesClient.Kubernetes(), s.InformerFactory))
-	//urlruntime.Must(tenantv1alpha2.AddToContainer(s.container, s.KubernetesClient, s.InformerFactory, s.DBClient.Database()))
 	urlruntime.Must(terminalv1alpha2.AddToContainer(s.container, s.KubernetesClient.Kubernetes(), s.KubernetesClient.Config()))
 	urlruntime.Must(iamv1alpha2.AddToContainer(s.container, s.KubernetesClient, s.InformerFactory, s.LdapClient, s.CacheClient, s.Config.AuthenticationOptions))
 	urlruntime.Must(oauth.AddToContainer(s.container, token.NewJwtTokenIssuer(token.DefaultIssuerName, s.Config.AuthenticationOptions, s.CacheClient), s.Config.AuthenticationOptions))
 	urlruntime.Must(servicemeshv1alpha2.AddToContainer(s.container))
+	devopsv1alpha2Service := ksruntime.NewWebService(devopsv1alpha2.GroupVersion)
```
zryfish, line 148: Do we have to create new webservice here? why not in `AddToContainer` method to align with the reset

runzexia: Before this place was marked as needing to be refactored because there are too long parameters, so use webservice instead.
  If there are multiple addToContainer with the same prefix path, it will cause panic at startup

cmd/ks-apiserver/app/options/options.go
runzexia: fake endpoint

pkg/apiserver/apiserver.go
runzexia: https://github.com/emicklei/go-restful/blob/master/container.go#L100

cmd/ks-apiserver/app/options/options.go
runzexia: https://github.com/kubesphere/kubesphere/pull/1961/files#diff-a3fe81fb19317e98c13a364d1f3c091cR120

PR #1986, opened 2020-04-03
fix bug for create ingress controller

### LINKED ISSUE #1987: Bug for creating ingress
**Describe the Bug**
1. The configmap name for the ingrees controller created is not the same as the _--configmap_ in spec.containers.args
2. If repeatedly create the ingress controller, the parameters in spec.containers.args will repeat or error

**Versions Used**
KubeSphere: 2.1.1

klj890:
**What type of PR is this?**
>
> /kind bug

This pr fix this issues https://github.com/kubesphere/kubesphere/issues/1987

Fixes #1987


FeynmanZhou:
@klj890 

It would be better to change the title with a detailed description, and use English in your comments above.

zryfish:
Please fix conflicts

PR #1989, opened 2020-04-05
add iam crd

wansir:

**What type of PR is this?**
/kind api-change
/kind cleanup


cmd/controller-manager/app/server.go
```diff
@@ -151,6 +151,14 @@ func Run(s *options.KubeSphereControllerManagerOptions, stopCh <-chan struct{})
 		// Start cache data after all informer is registered
 		informerFactory.Start(stopCh)
 
+		// Setup webhooks TODO enable webhook
```
zryfish, line 154: Please remove unused code

pkg/apis/iam/v1alpha2/globalrole_types.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +genclient:nonNamespaced
+// +kubebuilder:resource:categories="iam",scope="Cluster"
+
+// GlobalRole is the Schema for the globalroles API
+// +k8s:openapi-gen=true
+type GlobalRole struct {
```
zryfish, line 31: What does this global role do? Is there any documentation about this?

pkg/apiserver/query/types.go
```diff
@@ -52,29 +33,30 @@ type Pagination struct {
 	// items per page
 	Limit int
 
-	// page number
-	Page int
+	// offset
+	Offset int
 }
```
zryfish, line 38: Why are we using **Offset** instead of **Page**?

pkg/apiserver/query/types.go
```diff
@@ -86,34 +68,34 @@ func (p *Pagination) GetPaginationSettings(total int) (startIndex, endIndex int)
...
+		SortBy:     "",
+		Ascending:  false,
+		Filters:    []Filter{},
 	}
 }
 
 type Filter struct {
 	Field Field
-	Value Comparable
+	Value Value
```
zryfish, line 80: What's the intention of changing Comparable to Value(string)?

pkg/kapis/resources/v1alpha3/handler.go
```diff
@@ -22,22 +22,8 @@ func New(factory informers.InformerFactory) *Handler {
...
-		return
-	}
-
-	response.WriteHeaderAndEntity(http.StatusOK, result)
-}
-
-// handleListNamedResource retrieves namespaced scope resources
-func (h Handler) handleListNamespacedResource(request *restful.Request, response *restful.Response) {
+// ListResource retrieves resources
+func (h Handler) ListResource(request *restful.Request, response *restful.Response) {
```
zryfish, line 26: Unnecessary exported function, and what's wrong with the function name of handleGetNamespacedResource or handleListNamespacedResource?

pkg/kapis/resources/v1alpha3/register.go
```diff
@@ -45,14 +45,26 @@ func AddToContainer(c *restful.Container, informerFactory informers.InformerFact
 	webservice := runtime.NewWebService(GroupVersion)
 	handler := New(informerFactory)
 
+	webservice.Route(webservice.GET("/{resources}").
+		To(handler.ListResource).
+		Metadata(restfulspec.KeyOpenAPITags, []string{tagNamespacedResource}).
+		Doc("Cluster level resource query").
+		Param(webservice.PathParameter("resources", "namespace level resource type, e.g. pods,jobs,configmaps,services.")).
```
zryfish, line 52: Is this parameter description right?

pkg/models/iam/im/im_operator.go
```diff
@@ -0,0 +1,85 @@
...
+func NewOperator(ksClient kubesphereclient.Interface, informer informers.SharedInformerFactory) IdentityManagementInterface {
+
+	return &crdOperator{
+		ksClient: ksClient,
+		informer: informer,
+	}
+
+}
+
+type crdOperator struct {
```
zryfish, line 38: crdOperator is unclear, maybe we should use another name.

pkg/kapis/iam/v1alpha2/register.go
```diff
@@ -44,7 +44,7 @@ func AddToContainer(c *restful.Container, k8sClient k8s.Client, factory informer
 	handler := newIAMHandler(k8sClient, factory, ldapClient, cacheClient, options)
 
 	// implemented by create CRD object.
-	//ws.Route(ws.POST("/users"))
+	//ws.Route(ws.POST("/users").To(handler.CreateUser))
```
zryfish, line 47: What's the intention of exposing users api on ks-apiserver, since we can do it directly through kube-apiserver?

pkg/models/resources/v1alpha3/interface.go
```diff
@@ -52,14 +47,80 @@ func DefaultList(objects []runtime.Object, query *query.Query, compareFunc Compa
...
+		// /namespaces?page=1&limit=10&ownerKind=Workspace
+	case query.FieldOwnerKind:
+		for _, ownerReference := range item.OwnerReferences {
+			if strings.Compare(ownerReference.Kind, string(filter.Value)) == 0 {
+				return true
+			}
+		}
+		return false
+		// /namespaces?page=1&limit=10&label=kubesphere.io/workspace:system-workspace
+	case query.FieldLabel:
```
zryfish, line 100: The following comparison is actually a prefix match. for example, we have a deployment with labels `app.kubernetes.io/name:bookinfo` will match to query `app.kubernetes.io/name:book`, but not `app.kubernetes.io/name:info`, because you are using `strings.Contains`.

pkg/models/resources/v1alpha3/interface.go
```diff
@@ -52,14 +47,80 @@ func DefaultList(objects []runtime.Object, query *query.Query, compareFunc Compa
...
+		return false
+		// /namespaces?page=1&limit=10&annotation=openpitrix_runtime
+	case query.FieldAnnotation:
+		for key, value := range item.Annotations {
+			if strings.Contains(fmt.Sprintf("%s:%s", key, value), string(filter.Value)) {
+				return true
+			}
+		}
+		return false
+	case query.FieldClusterName:
```
zryfish, line 115: ClusterName in objectMeta is not set by k8s now, we can ignore.
```
         // The name of the cluster which the object belongs to.
	// This is used to distinguish resources with same name and namespace in different clusters.
	// This field is not set anywhere right now and apiserver is going to ignore it if set in create or update request.
	// +optional
	ClusterName string `json:"clusterName,omitempty" protobuf:"bytes,15,opt,name=clusterName"`
```

cmd/controller-manager/app/server.go
wansir: I'm still debugging the webhook server.

pkg/apiserver/query/types.go
wansir: Page usually start at 1, offset start at 0.

pkg/apiserver/query/types.go
wansir: It is difficult to extend a compareable object, we need to define a lot of uncommon comparable objects under the `pkg/apiserver/query`. Usually we only need to compare based on string type and implement a special compare function for the resource. I implement the general compare function for metav1.ObjectMeta. https://github.com/kubesphere/kubesphere/pull/1989/commits/028d229c083648d24e045730d4e2f2c2061f050b#diff-ba2636a45a25c29c988c339b74061594R72-R120

pkg/kapis/resources/v1alpha3/handler.go
wansir: agree

PR #2007, opened 2020-04-13
feat: custom monitoring

huanggze:

**What type of PR is this?**
>

/kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

Tow new apis:
- /namespaces/{namespace}/targets/metadata
- /namespaces/{namespace}/targets/query?expr=\<required\>

The first is to retrive metric list with help and type. The second is to perform an adhoc query.

For dashboard data model, see [kubesphere/monitoring-dashboard](https://github.com/kubesphere/monitoring-dashboard) for more information. It is based on CRD `dashboard`.

It is worth noting that dashboards custom resources with label-value pair `dashboard.monitoring.kubesphere.io/template: true` are assumed to be dashboard templates. No need to duplicate them into a ConfigMap as we discussed before.

Fixes #

@zryfish We need to upgrade prometheus mod 

```docs

```


pkg/models/monitoring/expressions/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	for i, tt := range tests {
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
+			result, err := labelReplace(tt.expr, "default")
+			if err != nil {
+				if !tt.expectedErr {
+					t.Fatal(err)
+				}
+				return
```
benjaminhuo, line 43: return should be changed to continue?

pkg/models/monitoring/expressions/prometheus/prometheus.go
```diff
@@ -0,0 +1,99 @@
+package prometheus
```
benjaminhuo, line 1: this file is already in a folder called prometheus, is it better to change the file name prometheus.go to a more meaningful name like label_replace.go or others ?

pkg/models/monitoring/monitoring.go
```diff
@@ -38,14 +40,22 @@ func NewMonitoringOperator(client monitoring.Interface) MonitoringOperator {
 	return &monitoringOperator{client}
 }
 
-// TODO(huanggze): reserve for custom monitoring
-func (mo monitoringOperator) GetMetrics(stmts []string, time time.Time) Metrics {
-	panic("implement me")
+func (mo monitoringOperator) GetMetric(expr, namespace string, time time.Time) (monitoring.Metric, error) {
+	// Enforce namespace matcher
+	expr, err := expressions.ReplaceNamespaceFns["prometheus"](expr, namespace)
```
benjaminhuo, line 45: Is it better to check if the value of key "prometheus" exist first, and then call the label replace function?

pkg/models/monitoring/monitoring.go
```diff
@@ -38,14 +40,22 @@ func NewMonitoringOperator(client monitoring.Interface) MonitoringOperator {
...
+	}
+	return mo.c.GetMetric(expr, time), nil
 }
 
-// TODO(huanggze): reserve for custom monitoring
-func (mo monitoringOperator) GetMetricsOverTime(stmts []string, start, end time.Time, step time.Duration) Metrics {
-	panic("implement me")
+func (mo monitoringOperator) GetMetricOverTime(expr, namespace string, start, end time.Time, step time.Duration) (monitoring.Metric, error) {
+	// Enforce namespace matcher
+	expr, err := expressions.ReplaceNamespaceFns["prometheus"](expr, namespace)
```
benjaminhuo, line 54: Same here, it's better to check if the value of key "prometheus" exist first, and then call the label replace function?

pkg/simple/client/monitoring/prometheus/prometheus.go
```diff
@@ -24,14 +25,35 @@ func NewPrometheus(options *Options) (monitoring.Interface, error) {
 	return prometheus{client: apiv1.NewAPI(client)}, err
 }
 
-// TODO(huanggze): reserve for custom monitoring
-func (p prometheus) GetMetrics(stmts []string, time time.Time) []monitoring.Metric {
-	panic("implement me")
+func (p prometheus) GetMetric(expr string, ts time.Time) monitoring.Metric {
+	var parsedResp monitoring.Metric
+
+	value, err := p.client.Query(context.Background(), expr, ts)
```
benjaminhuo, line 31: context.Background or context.WithTimeout?
Is it reasonable to enforce the query timeout here?

pkg/models/monitoring/expressions/prometheus/prometheus_test.go
huanggze: It is to return the Run method, and it won't break the outter loop

pkg/models/monitoring/monitoring.go
huanggze: The idea is that we allow for various monitoring backend implements, and prometheus is just one of them. Different implements have different ways to enforce namespace isolation. No matter how, all of them should register themselves in `ReplaceNamespaceFns` during init. 

Below is how prometheus conducts. So the key "prometheus" always exists. We hard code here because we only have one monitoring backend imeplemets so far, that is prometheus. But the framework allows extension.
 
https://github.com/kubesphere/kubesphere/commit/b09541729a84d99edecf7e21e29965c54c76a152#diff-791d0597894092e1fe80cdaee06c798eR11-R13

pkg/models/monitoring/expressions/prometheus/prometheus.go
huanggze: yep, label_replace.go looks meaningful. done

pkg/models/monitoring/monitoring.go
benjaminhuo: I see. It's ok as long as no nil reference occurs.

pkg/models/monitoring/expressions/prometheus/prometheus_test.go
benjaminhuo: I see, the return is in another function

pkg/simple/client/monitoring/prometheus/prometheus.go
huanggze: I think the context puts an extra requirement on the underlying HTTP connection.  But maybe we don't need to do that.

To enable a client-side timeout, we could adjust the http client on initialization. By now, I think the default setting is good.
https://github.com/kubesphere/kubesphere/blob/31135c17aea1cd76df037f4acc402b37bd9542be/pkg/simple/client/monitoring/prometheus/prometheus.go#L19-L26

benjaminhuo: ok

pkg/models/monitoring/monitoring.go
huanggze: I put more comments on this issue for further clarification.

pkg/models/monitoring/monitoring.go
```diff
@@ -38,14 +40,28 @@ func NewMonitoringOperator(client monitoring.Interface) MonitoringOperator {
 	return &monitoringOperator{client}
 }
 
-// TODO(huanggze): reserve for custom monitoring
-func (mo monitoringOperator) GetMetrics(stmts []string, time time.Time) Metrics {
-	panic("implement me")
+func (mo monitoringOperator) GetMetric(expr, namespace string, time time.Time) (monitoring.Metric, error) {
+	// Different monitoring backend implements have different ways to enforce namespace isolation.
+	// Each implement should register itself to `ReplaceNamespaceFns` during init().
+	// We hard code "prometheus" here because we only support this implement so for.
```
benjaminhuo, line 46: so for => so far

pkg/models/monitoring/monitoring.go
```diff
@@ -38,14 +40,28 @@ func NewMonitoringOperator(client monitoring.Interface) MonitoringOperator {
 	return &monitoringOperator{client}
 }
 
-// TODO(huanggze): reserve for custom monitoring
-func (mo monitoringOperator) GetMetrics(stmts []string, time time.Time) Metrics {
-	panic("implement me")
+func (mo monitoringOperator) GetMetric(expr, namespace string, time time.Time) (monitoring.Metric, error) {
+	// Different monitoring backend implements have different ways to enforce namespace isolation.
```
benjaminhuo, line 44: implements => implementation ?

pkg/models/monitoring/monitoring.go
benjaminhuo: That's good.
Two minor syntax issues.

pkg/models/monitoring/monitoring.go
huanggze: done

pkg/models/monitoring/monitoring.go
huanggze: done

PR #2008, opened 2020-04-13
fix iam admission webhook

wansir:

### DIFF
--- a/pkg/controller/user/user_webhook.go
+++ b/pkg/controller/user/user_webhook.go
@@ -21,7 +21,6 @@ package user
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"golang.org/x/crypto/bcrypt"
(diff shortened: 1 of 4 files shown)


pkg/controller/user/user_webhook.go
```diff
@@ -51,28 +50,26 @@ func (a *EmailValidator) Handle(ctx context.Context, req admission.Request) admi
...
 		return admission.Errored(http.StatusInternalServerError, err)
 	}
 
-	found := emailAlreadyExist(allUsers, email)
+	alreadyExist := emailAlreadyExist(allUsers, user)
 
-	if !found {
-		return admission.Denied(fmt.Sprintf("email %s must be unique", email))
+	if alreadyExist {
+		return admission.Denied("user email must be unique")
```
zryfish, line 64: I think error message `user email already exists` is better

PR #2011, opened 2020-04-14
migrate legacy API

wansir:

### DIFF
--- a/pkg/models/tenant/tenant.go
+++ b/pkg/models/tenant/tenant.go
@@ -21,63 +21,91 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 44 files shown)


pkg/models/tenant/tenant.go
```diff
@@ -21,63 +21,91 @@ import (
...
-	ListWorkspaces(username string) (*api.ListResult, error)
-	ListNamespaces(username, workspace string) (*api.ListResult, error)
+	ListWorkspaces(user user.Info) (*api.ListResult, error)
+	ListNamespaces(user user.Info, workspace string) (*api.ListResult, error)
 }
 
 type tenantOperator struct {
-	informers informers.InformerFactory
-	am        am.AccessManagementInterface
+	informers  informers.InformerFactory
```
zryfish, line 42: if only kubesphereinformer needed,  it's unnecessary to import the whole InformerFactory

wansir: kubernetesinformer is also used

PR #2012, opened 2020-04-15
Network policy

duanjiong:
network: support network isolation

1.  Add new crds to convert kubesphere network policy to k8s network policy, and then other network plugins will do the rest work.
2.  Port cache.go from calico project's kube-controller,  it aims to  sync nsnp with k8s np, delete unused np, and relieve the pressure on k8s restful client.
3. If you want higher performance, you can implement interface NsNetworkPolicyProvider in pkg/controller/provider/namespace_np.go.


zryfish:
I don't think we need `ks-network`, we can build those into ks-controller-manager which holds all the controllers.

pkg/controller/network/provider/ns_k8s.go
```diff
@@ -0,0 +1,235 @@
...
+// syncToDatastore syncs the given update to the k8s datastore. The provided key can be used to
+// find the corresponding resource within the resource cache. If the resource for the provided key
+// exists in the cache, then the value should be written to the datastore. If it does not exist
+// in the cache, then it should be deleted from the datastore.
+func (c *k8sPolicyController) syncToDatastore(key string) error {
+	// Check if it exists in the controller's cache.
+	obj, exists := c.resourceCache.Get(key)
+	if !exists {
+		// The object no longer exists - delete from the datastore.
+		klog.Infof("Deleting NetworkPolicy %s from k8s datastore", key)
```
zheng1, line 125: Consistent terms in the log, `NetworkPolicy` or `network policy`

cmd/controller-manager/app/controllers.go
```diff
@@ -125,6 +127,16 @@ func AddControllers(
 		client.KubeSphere().ClusterV1alpha1().Agents(),
 		client.KubeSphere().ClusterV1alpha1().Clusters())
 
+	k8sProvider, err := provider.NewNsNetworkPolicyProvider(client.Kubernetes(),
```
zheng1, line 130: `k8sProvider  => nsnpProvider`

zryfish:
Why are we go back Namespace network policy instead of Workspace policy? What's the difference between them? And how the new namespace network policy behaves inside workspace? Can I access services from other projects but within the same workspace, still remain isolated with namespaces outside of the workspace?

zryfish:
And need to update vendor using script under directory `hack`,  using `hack/pin-dependency.sh` to pin a specific version, and then use `hack/update-vendor.sh` to update the vendor directory. 

PS. we need to update the development docs to clarify this.

duanjiong:
> Why are we go back Namespace network policy instead of Workspace policy? What's the difference between them? And how the new namespace network policy behaves inside workspace? Can I access services from other projects but within the same workspace, still remain isolated with namespaces outside of the workspace?

There two kind isolation, workspace and project. workspace use a new field "networkIsolation" in spec, and project use namespace annotation "kubesphere.io/network-isolate = enabled".  I think those two isolation, just different in k8s network policy, they cannot coexist, so we can just use one k8s network policy do this. 

Let's say I have two projects, A and B, they are in same workspace. If the A and B have not set annotation  "kubesphere.io/network-isolate = enabled", they can access each other,  the same goes for services.  If project A set annotation  "kubesphere.io/network-isolate = enabled",  A cannot access service in B, unless A allow egress for service in B. 

PS. now it just work correctly when proxy mode is iptables, because ipvs will use src ip which assigned by cni plugin, and we cannot obtain this ip.  we will support ipvs in future.

duanjiong:
> And need to update vendor using script under directory `hack`, using `hack/pin-dependency.sh` to pin a specific version, and then use `hack/update-vendor.sh` to update the vendor directory.
> 
> PS. we need to update the development docs to clarify this.

ok, i will do it soon

zryfish:
need to resolve conflict issues

PR #2051, opened 2020-05-07
fix list namespaces

wansir:

**What type of PR is this?**

/kind bug

List namespaces in workspace returned incorrect items.

### DIFF
--- a/pkg/apiserver/filters/kubeapiserver.go
+++ b/pkg/apiserver/filters/kubeapiserver.go
@@ -35,8 +35,8 @@ func WithKubeAPIServer(handler http.Handler, config *rest.Config, failed proxy.E
 
 			// Do not cover k8s client authorization header
 			req.Header.Del("Authorization")
-
 			httpProxy := proxy.NewUpgradeAwareHandler(&s, defaultTransport, true, false, failed)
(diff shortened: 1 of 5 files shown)


pkg/apiserver/filters/kubeapiserver.go
```diff
@@ -35,8 +35,8 @@ func WithKubeAPIServer(handler http.Handler, config *rest.Config, failed proxy.E
 
 			// Do not cover k8s client authorization header
 			req.Header.Del("Authorization")
-
 			httpProxy := proxy.NewUpgradeAwareHandler(&s, defaultTransport, true, false, failed)
+			httpProxy.UpgradeTransport = proxy.NewUpgradeRequestRoundTripper(defaultTransport, defaultTransport)
```
zryfish, line 39: What's this for? Please add more comments to clarify

wansir: NewUpgradeRequestRoundTripper takes two round trippers - one for the underlying TCP connection, and one that is able to write headers to an HTTP request. 

see also: https://github.com/kubernetes/kubernetes/pull/88781

zryfish: Please add more comments, this will disable http2 for all requests. I think it's a temporary solution, we'd better switch back in future once upgraded to v1.18

PR #2065, opened 2020-05-13
fix devops create and delete bug

shaowenchen:
**What type of PR is this?**
/kind bug
/area devops

**What this PR does / why we need it:**
This pr fixes issues which user can't create or delete devopsprojects, because
- Jenkin go client get the error code unsteadily，404 or 500.
- delete Jenkins parameters is not name，but AdminNamespace.

**Special notes for your reviewer:**
```
NONE
```


shaowenchen:
/cc @zryfish 

fix unit test for ci

PR #2068, opened 2020-05-14
init opClient

pengcong06:
**What type of PR is this?**

> /kind bug
> /area appstore

init opClient


pkg/kapis/openpitrix/v1/register.go
```diff
@@ -39,7 +39,12 @@ const (
 
 var GroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1"}
 
+var ErrorNilClient = errors.New("the opClient is nil")
+
 func AddToContainer(c *restful.Container, factory informers.InformerFactory, op op.Client) error {
+	if op == nil {
```
zryfish, line 45: if client is nil, it means OpenPitrix is not enabled, we should do nothing and return nil so we can continue registering other API groups.

PR #2069, opened 2020-05-14
add events search apis

junotx:
**What type of PR is this?**

/kind feature

Add apis to search events


pkg/models/tenant/tenant.go
```diff
@@ -200,6 +210,50 @@ func (t *tenantOperator) ListNamespaces(user user.Info, workspace string, queryP
...
+	}
+	decision, _, err := t.authorizer.Authorize(listEvts)
+	if err != nil {
+		klog.Error(err)
+		return nil, err
+	}
+
+	var workspaces []*tenantv1alpha1.Workspace
+	lw, err := t.ListWorkspaces(user, query.New())
+	if err != nil {
```
wansir, line 230: Need to filter out the workspace that current user have the right to list events.

pkg/models/tenant/tenant.go
```diff
@@ -200,6 +210,50 @@ func (t *tenantOperator) ListNamespaces(user user.Info, workspace string, queryP
...
+		ResourceRequest: true,
+	}
+	decision, _, err := t.authorizer.Authorize(listEvts)
+	if err != nil {
+		klog.Error(err)
+		return nil, err
+	}
+
+	var workspaces []*tenantv1alpha1.Workspace
+	lw, err := t.ListWorkspaces(user, query.New())
```
wansir, line 229: Need to filter out the workspace that current user have the right to list events.

junotx: Ok, let me add that

benjaminhuo: Looks like the filtering you mention is done in the IntersectedNamespaces function @wansir

wansir: ListWorkspaces only returns workspaces of a specific user, but the user may not have permission to list events in all the namespaces, even if they can view namespaces. Filtering logic is necessary.

benjaminhuo: I see.

benjaminhuo:
@junotx , there are conflicts that must be resolved

pkg/simple/client/events/interface.go
```diff
@@ -0,0 +1,44 @@
+package events
+
+import (
+	v1 "k8s.io/api/core/v1"
+	"time"
+)
+
+type Client interface {
+	SearchEvents(filter *Filter, from, size int64, sort string) (*Events, error)
```
zryfish, line 9: How to do unit testing without connecting to a real elasticsearch service? Please add more test code to this interface, you can refer to LDAP interface which implements a fake client that can be used to do simple unit testing. This is what we try to achieve, running standalone ks-apiserver or ks-controller-manager without all dependency.

junotx: Now it is added

PR #2086, opened 2020-05-18
monitoring: fix metric value NaN

huanggze:

**What type of PR is this?**
/kind bug

To address https://github.com/golang/go/issues/3480, implement custom json marshler for `type Point [2]float64`.

```
```

```docs

```


pkg/models/monitoring/sort_page.go
```diff
@@ -50,9 +50,22 @@ func (w wrapper) Less(i, j int) bool {
 	p := w.MetricValues[i]
 	q := w.MetricValues[j]
 
-	if p.Sample.Value() == q.Sample.Value() {
+	// Place Nil to the tail.
+	if p.Sample == nil && q.Sample != nil {
```
benjaminhuo, line 54: nil < not nil : false
not nil < nil : true
You mean the nil should be placed at the end no matter it's descending or ascending order?

pkg/models/monitoring/sort_page.go
```diff
@@ -50,9 +50,22 @@ func (w wrapper) Less(i, j int) bool {
...
+		return false
+	}
+	if p.Sample != nil && q.Sample == nil {
+		return true
+	}
+
+	// If both Samples are Nil or have the same metric value, sort by resource name
+	if p.Sample == q.Sample || p.Sample[1] == q.Sample[1] {
 		return p.Metadata[w.identifier] < q.Metadata[w.identifier]
```
benjaminhuo, line 63: this looks like for ascending order only?
what about descending?

pkg/models/monitoring/sort_page.go
```diff
@@ -50,9 +50,22 @@ func (w wrapper) Less(i, j int) bool {
...
 	}
+	// Place NaN to the tail (NaN takes precedence over Nil).
+	if math.IsNaN(p.Sample[1]) != math.IsNaN(q.Sample[1]) {
+		return !math.IsNaN(p.Sample[1])
```
benjaminhuo, line 67: NaN value is placed after non-NaN value no matter it's ascending or descending order?

pkg/models/monitoring/sort_page.go
huanggze: Yes. Actually for a long time, the nil pointer panic has never been triggered, because Prometheus server response is always reliable and flawless.

See manually constructed faulty test data:
https://github.com/kubesphere/kubesphere/blob/14069b95f231440a9254b8ec5c700d397d90adf3/pkg/models/monitoring/testdata/faulty-node-metrics-sorted-2.json#L223-L227

pkg/models/monitoring/sort_page.go
huanggze: The ranking rule is like this, because `p` and `q` has the same value (for example, cpu utilization for node `i-js38t` and node  `i-2p9s0` are both 35.6%), to prevent random ranking, sort `p` and `q` by their resource name in **ASCENDING** order. As a result, `i-2p9s0` take precedence over  `i-js38t`. 

I didn't change the ranking rule since ks 2.0. So if the equal value case happens, resource name sort order is always **ASCENDING**, and won't be affected by user-specified `order`.

pkg/models/monitoring/sort_page.go
huanggze: Yep, again, because we have well-designed builtin promqls. NaN (and +Inf, -Inf) kindof never occurs. But it's possible in custom monitoring.

See test data:
https://github.com/kubesphere/kubesphere/blob/14069b95f231440a9254b8ec5c700d397d90adf3/pkg/models/monitoring/testdata/faulty-node-metrics-sorted-3.json#L5-L81

pkg/models/monitoring/sort_page.go
benjaminhuo: ok, that's fine to me

PR #2112, opened 2020-05-25
fix SecretRef nil

shaowenchen:

**What type of PR is this?**
/kind bug

when the parameter is {“remoteUrl”:“https://github.com/xxx.git”}，credential.SecretRef is nil .
This pr fixs this situation.

### DIFF
--- a/pkg/kapis/resources/v1alpha2/handler.go
+++ b/pkg/kapis/resources/v1alpha2/handler.go
@@ -274,8 +274,12 @@ func (r *resourceHandler) handleVerifyGitCredential(request *restful.Request, re
 		api.HandleBadRequest(response, nil, err)
 		return
 	}
-
-	err = r.gitVerifier.VerifyGitCredential(credential.RemoteUrl, credential.SecretRef.Namespace, credential.SecretRef.Name)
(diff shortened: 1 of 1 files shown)


pkg/kapis/resources/v1alpha2/handler.go
```diff
@@ -274,8 +274,12 @@ func (r *resourceHandler) handleVerifyGitCredential(request *restful.Request, re
...
 		return
 	}
-
-	err = r.gitVerifier.VerifyGitCredential(credential.RemoteUrl, credential.SecretRef.Namespace, credential.SecretRef.Name)
+	var namespace, secretName string
+	if credential.SecretRef != nil {
+		namespace = credential.SecretRef.Namespace
+		secretName = credential.SecretRef.Name
+	}
+	err = r.gitVerifier.VerifyGitCredential(credential.RemoteUrl, namespace, secretName)
```
zryfish, line 282: Is it ok when namespace and secretName are empty?

shaowenchen: Yes， it‘s ok both values are empty and secret is empty

PR #2115, opened 2020-05-25
Fix query credential

shaowenchen:
**What type of PR is this?**
/kind bug

- fix query credential ，but include others secrets

### DIFF
--- a/pkg/apiserver/apiserver.go
+++ b/pkg/apiserver/apiserver.go
@@ -366,6 +366,8 @@ func (s *APIServer) waitForResourceSync(stopCh <-chan struct{}) error {
 		{Group: "devops.kubesphere.io", Version: "v1alpha1", Resource: "s2ibuildertemplates"},
 		{Group: "devops.kubesphere.io", Version: "v1alpha1", Resource: "s2iruns"},
 		{Group: "devops.kubesphere.io", Version: "v1alpha1", Resource: "s2ibuilders"},
+		{Group: "devops.kubesphere.io", Version: "v1alpha3", Resource: "devopsprojects"},
+		{Group: "devops.kubesphere.io", Version: "v1alpha3", Resource: "pipelines"},
 	}
 
 	servicemeshGVRs := []schema.GroupVersionResource{
--- a/pkg/models/devops/devops.go
+++ b/pkg/models/devops/devops.go
@@ -24,9 +24,11 @@ import (
 	"io/ioutil"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
+	"kubesphere.io/kubesphere/pkg/api"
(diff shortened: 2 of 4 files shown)


pkg/models/devops/devops.go
```diff
@@ -242,7 +242,25 @@ func (d devopsOperator) ListCredentialObj(workspace string, projectName string)
 	if err != nil {
 		return nil, err
 	}
-	return d.k8sclient.CoreV1().Secrets(projectObj.Status.AdminNamespace).List(metav1.ListOptions{})
+	credentialList, err := d.k8sclient.CoreV1().Secrets(projectObj.Status.AdminNamespace).List(metav1.ListOptions{})
```
zryfish, line 245: using lister instead will be faster, normally we use client only when update needed

shaowenchen: I will improve it in next pr，for devops/pipelines/credentials.

zryfish: please fix it now to keep a clean commit log

shaowenchen: fixed

pkg/apiserver/apiserver.go
```diff
@@ -365,6 +365,8 @@ func (s *APIServer) waitForResourceSync(stopCh <-chan struct{}) error {
 		{Group: "devops.kubesphere.io", Version: "v1alpha1", Resource: "s2ibuildertemplates"},
 		{Group: "devops.kubesphere.io", Version: "v1alpha1", Resource: "s2iruns"},
 		{Group: "devops.kubesphere.io", Version: "v1alpha1", Resource: "s2ibuilders"},
+		{Group: "devops.kubesphere.io", Version: "v1alpha3", Resource: "devopsprojects"},
```
zryfish, line 368: Why crd named `devopsproject`, and `devops` in restful API spec?

shaowenchen: The CRD 's name is devopsproject, but there are some old /kapis using /devops/{devops} path.
The consistent url path is good for the convenience of devops authentication. 
/cc @wansir

zryfish: Why we name the crd `devops` if for compatibility reason? It's more clear

shaowenchen: Devopsproject is used as the name of CRD during the refactoring phase. Here just keep consistense.

pkg/models/devops/devops.go
```diff
@@ -158,90 +160,145 @@ func (d devopsOperator) UpdateDevOpsProject(workspace string, project *v1alpha3.
 	return d.ksclient.DevopsV1alpha3().DevOpsProjects().Update(project)
 }
 
-func (d devopsOperator) ListDevOpsProject(workspace string) (*v1alpha3.DevOpsProjectList, error) {
-	return d.ksclient.DevopsV1alpha3().DevOpsProjects().List(metav1.ListOptions{})
+func (d devopsOperator) ListDevOpsProject(workspace string, limit, offset int) (models.PageableResponse, error) {
```
zryfish, line 163: we are switching to api.ListResult `pkg/api/types.go`, please use api.ListResult instead of models.PageableResponse

shaowenchen: fixed

PR #2117, opened 2020-05-25
modify devops api for tenant

shaowenchen:
**What type of PR is this?**
 /kind feature

normalize devops API for tenant 

/cc @wansir


pkg/kapis/devops/v1alpha3/register.go
```diff
@@ -51,128 +51,120 @@ func AddToContainer(container *restful.Container, devopsClient devopsClient.Inte
 		ws := runtime.NewWebService(GroupVersion)
 		handler := newDevOpsHandler(devopsClient, k8sclient, ksclient, ksInformers, k8sInformers)
 		// credential
-		ws.Route(ws.GET("/workspaces/{workspace}/devopsprojects/{projectName}/credential/").
+		ws.Route(ws.GET("/devops/{devops}/credentials/").
```
wansir, line 54: suffix `/`  is unnecessary

shaowenchen: fixed

PR #2126, opened 2020-05-27
fix events search

junotx:
**What type of PR is this?**

> Uncomment only one `/kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
> 
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:

Fix bug of query to events without `involvedObject.namespace` field
Remove redundant judgment code in `listIntersectedNamespaces` method

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for reviewers**:
```
```

**Additional documentation, usage docs, etc.**:

```docs

```

PR #2129, opened 2020-05-28
add auditing events search api

wanjunlei:
What type of PR is this?

/kind feature

Add auditing events search api

### DIFF
--- a/pkg/simple/client/auditing/elasticsearch/elasticsearch.go
+++ b/pkg/simple/client/auditing/elasticsearch/elasticsearch.go
@@ -1,0 +1,391 @@
+/*
+Copyright 2020 The KubeSphere Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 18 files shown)


pkg/simple/client/auditing/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,391 @@
...
+						"match_phrase": map[string]string{"ObjectRef.Namespace.keyword": k},
+					}, {
+						"range": map[string]interface{}{
+							"RequestReceivedTimestamp": map[string]interface{}{
+								"gte": v,
+							},
+						},
+					}},
+				},
+			})
```
junotx, line 266: The case where namespace is empty requires special handling, and you can refer to [this pr](https://github.com/kubesphere/kubesphere/pull/2126) @wanjunlei

wanjunlei: When namespace is empty, the value in es is "", so it dose not need special handling

junotx: So are those fields with empty value also stored in es audit indices ?

wanjunlei: yes

benjaminhuo: So there is no need to adjust this ? @junotx @wanjunlei

wanjunlei: yes

benjaminhuo:
@zryfish is this PR ok for you?

PR #2132, opened 2020-05-29
update user's role templates API

wansir:

**What type of PR is this?**

/kind api-change

Update user's role templates API.


pkg/apiserver/request/requestinfo.go
```diff
@@ -134,10 +134,20 @@ func (r *RequestInfoFactory) NewRequestInfo(req *http.Request) (*RequestInfo, er
 	requestInfo.APIPrefix = currentParts[0]
 	currentParts = currentParts[1:]
 
+	// Distinguish global resources and cluster resources with the same name
+	// such as /kapis/clusters/{cluster}/iam.kubesphere.io/v1alpha2/users and /kapis/iam.kubesphere.io/v1alpha2/users
+	// If the injected cluster info in the request header means that cluster resources is requested
+	if cluster := req.Header.Get("cluster"); cluster != "" {
```
zryfish, line 140: Still don't understand why we need this custom header ?

wansir: Distinguish global resources and cluster resources with the same name.

https://github.com/kubesphere/kubesphere/blob/2f650dbac551d6a66957acb519c60df76b761c9e/pkg/kapis/iam/v1alpha2/handler.go#L125-L130

PR #2138, opened 2020-05-30
fix: crash if config file not provide

wansir:

**What type of PR is this?**

/kind bug

fix: crash if config file not provide.


zryfish:
Please fix conflicts

PR #2139, opened 2020-05-30
fix: support redirect proxy authorization

wansir:

**What type of PR is this?**

/kind feature

Support redirect proxy authorization.

for example:

The cluster scope resource request(`/kapis/clusters/{cluster}/ openpitrix.io/v1/namespaces/{namespace}/applications`) need to authenticate through ks-apiserver in member cluster and then handle the request(`/kapis/ openpitrix.io/v1/namespaces/{namespace}/applications`) in host cluster, need to authenticate before redirection.

This PR fixes the wrong redirect path.


zryfish:
Fixed in another pr, closing

Issue #1769, opened 2020-01-14
Storage management


wnxn:
1. Get storage class capability: access mode, snapshot, resizing (online or offline), monitoring, cloning. #1581 #1427 #951
2. Install storage plugin: install from application market. #833 
3. Uninstall storage plugin

PR #2149, opened 2020-06-02
storage capability

min-zh:
**What type of PR is this?**
>
 /kind feature

controller of storage capability 

Fixes #

```
```

```docs

```


pkg/controller/storage/capability/capability_controller.go
```diff
@@ -0,0 +1,317 @@
...
+
+func (c *StorageCapabilityController) Start(stopCh <-chan struct{}) error {
+	return c.Run(5, stopCh)
+}
+
+func (c *StorageCapabilityController) Run(threadCnt int, stopCh <-chan struct{}) error {
+	defer utilruntime.HandleCrash()
+	defer c.workQueue.ShutDown()
+
+	isValid, err := c.IsValidKubernetesVersion()
```
zryfish, line 123: I think we should move this version check before controller initialization, it's not efficient after controller is running.

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -0,0 +1,317 @@
...
+		klog.V(4).Infof("CSI address of snapshot class: %s, driver:%s not exist", snapshotClass.Name, snapshotClass.Driver)
+		return
+	}
+	var key string
+	var err error
+	if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
+		utilruntime.HandleError(err)
+		return
+	}
+	c.workQueue.Add(key)
```
zryfish, line 173: enqueued storageClass key in function `enqueueStorageClass`, and here enqueued snapshotClass key, there are two different types of object,  why put them in one queue?

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	// If the resource exist, we can update it.
+	storageClassCapabilityUpdate := storageClassCapabilityExist.DeepCopy()
+	err = c.addSpec(&storageClassCapabilityUpdate.Spec, storageClass, snapshotClassCreated)
+	if err != nil {
+		return err
+	}
+	klog.V(4).Info("Update StorageClassCapability: ", storageClassCapabilityUpdate)
+	_, err = c.storageClassCapabilityClient.StorageV1alpha1().StorageClassCapabilities().Update(storageClassCapabilityUpdate)
```
zryfish, line 275: better to do a reflect.compare before updating, if they are identical, an update will change the object resource version which will lead to another object updated event

pkg/controller/storage/capability/csi_capability.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	u, err := url.Parse(address)
+	if err == nil && (!u.IsAbs() || u.Scheme == "unix") {
+		dialOptions = append(dialOptions,
+			grpc.WithDialer(
+				func(addr string, timeout time.Duration) (net.Conn, error) {
+					return net.DialTimeout("unix", u.Path, timeout)
+				}))
+	}
+	// This is necessary when connecting via TCP and does not hurt
```
zryfish, line 145: what does `hurt` mean here? Compatible?

zryfish:
And don't forget to run `hack/update-vendor.sh` after change `go.sum`

zheng1:
Related issues: https://github.com/kubesphere/console/issues/250 https://github.com/kubesphere/kubesphere/issues/1769
Related document: https://github.com/kubesphere/community/blob/master/sig-storage/concepts-and-designs/storage-capability-interface.md 

I saw "Storage Provisioner Capability CRD" and "Mutating Admission Webhook" describe in the document. Do these need to be included in this PR?

min-zh:
@zheng1 Yes,  "Storage Provisioner Capability CRD" and "Mutating Admission Webhook" are included in the initial design. It included two new CRDs: StorageClassCapability&ProvisionerCapability,  and three components: CRD-Controller, CSI-Sidecar, Mutating-Admission-Webhook. After reviewing with @zryfish，we think it was a little heavy and maybe difficult to realized and easily cause bugs. So we simplified the design, only one CRD StorageClassCapability and CRD controller are reserved

pkg/controller/storage/capability/capability_controller.go
min-zh: OK, I will fix it.

zheng1:
> @zheng1 Yes, "Storage Provisioner Capability CRD" and "Mutating Admission Webhook" are included in the initial design. It included two new CRDs: StorageClassCapability&ProvisionerCapability, and three components: CRD-Controller, CSI-Sidecar, Mutating-Admission-Webhook. After reviewing with @zryfish，we think it was a little heavy and maybe difficult to realized and easily cause bugs. So we simplified the design, only one CRD StorageClassCapability and CRD controller are reserved

OK, this is make sense.

pkg/controller/storage/capability/csi_capability.go
min-zh: I use the func from [csi-test](https://github.com/kubernetes-csi/csi-test/blob/master/utils/grpcutil.go#L31). I think 'does not hurt' means 'the dialOption has no influence' when using Unix domain socket. If this is not suitable，I will delete it.

pkg/controller/storage/capability/capability_controller.go
min-zh: If the snaposhtClass with the same name with the storageClass was created or deleted, I will handler the StorageClassCapability with same name. The handler logic is the same, so i put  them in one queue.

pkg/controller/storage/capability/capability_controller.go
min-zh: OK, I will fix it.

PR #2154, opened 2020-06-04
fix: clusterrolebinding controller bug

wansir:

**What type of PR is this?**

/kind bug

### DIFF
--- a/pkg/controller/clusterrolebinding/clusterrolebinding_controller.go
+++ b/pkg/controller/clusterrolebinding/clusterrolebinding_controller.go
@@ -22,7 +22,8 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
-	k8sinformers "k8s.io/client-go/informers"
+	appsv1informers "k8s.io/client-go/informers/apps/v1"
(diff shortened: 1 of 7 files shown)


pkg/controller/clusterrolebinding/clusterrolebinding_controller.go
```diff
@@ -168,7 +170,7 @@ func (c *Controller) processNextWorkItem() bool {
 			return nil
 		}
 		// Run the reconcile, passing it the namespace/name string of the
-		// Foo resource to be synced.
+		// Foo resource to be clusterRoleBindingSynced.
```
zryfish: Foo is example resource name from example code

pkg/controller/globalrolebinding/globalrolebinding_controller.go
```diff
@@ -213,9 +212,7 @@ func (c *Controller) reconcile(key string) error {
 		return err
 	}
 
-	isPlatformAdmin := globalRoleBinding.RoleRef.Name == iamv1alpha2.PlatformAdmin
-
-	if isPlatformAdmin {
+	if isPlatformAdmin := globalRoleBinding.RoleRef.Name == iamv1alpha2.PlatformAdmin; isPlatformAdmin {
```
zryfish, line 215: If `isPlatformAdmin` isn't used anywhere else, why bother to create one?
```suggestion
	if globalRoleBinding.RoleRef.Name == iamv1alpha2.PlatformAdmin {
```

PR #2161, opened 2020-06-05
fix jenkins basic auth

shaowenchen:

**What type of PR is this?**
/kind bug

The apiserver uses Bearer Token , but Jenkins needs Basic Token. This pr transfers the Auth Token.

### DIFF
--- a/pkg/simple/client/devops/jenkins/pure_request.go
+++ b/pkg/simple/client/devops/jenkins/pure_request.go
@@ -1,10 +1,15 @@
 package jenkins
 
 import (
+	"encoding/base64"
+	"fmt"
+	"github.com/dgrijalva/jwt-go"
 	"k8s.io/klog"
+	authtoken "kubesphere.io/kubesphere/pkg/apiserver/authentication/token"
 	"kubesphere.io/kubesphere/pkg/simple/client/devops"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/simple/client/devops/jenkins/pure_request.go
```diff
@@ -24,10 +30,41 @@ func (j *Jenkins) SendPureRequestWithHeaderResp(path string, httpParameters *dev
 
 	client := &http.Client{Timeout: 30 * time.Second}
 
+	header := httpParameters.Header
+	bearTokenArray := strings.Split(header.Get("Authorization"), " ")
+	bearFlag := bearTokenArray[0]
```
zryfish, line 35: What if there is no `Authorization` header in the request? Better to check bearTokenArray first.

pkg/simple/client/devops/jenkins/pure_request.go
```diff
@@ -24,10 +30,41 @@ func (j *Jenkins) SendPureRequestWithHeaderResp(path string, httpParameters *dev
...
+	if err == nil && bearFlag == "Bearer" {
```
zryfish, line 36: it's unnecessary to have `err == nil` since we returned if not.

zryfish: Also better to use case insensitive compare
```suggestion
	if strings.ToLower(bearFlag) == "bearer" {
```

pkg/simple/client/devops/jenkins/pure_request.go
```diff
@@ -24,10 +30,41 @@ func (j *Jenkins) SendPureRequestWithHeaderResp(path string, httpParameters *dev
...
+		bearToken := bearTokenArray[1]
+		if err != nil {
+			klog.Error(err)
+			return nil, nil, err
+		}
+		infoArray := strings.Split(bearToken, ".")
```
zryfish, line 42: Why not using a tokenIssuer `pkg/apiserver/authentication/token/jwt.go` to parse a token?

pkg/simple/client/devops/jenkins/pure_request.go
shaowenchen: Even if there is no Authorization header，bearTokenArray is assigned as [''] .

pkg/simple/client/devops/jenkins/pure_request.go
shaowenchen: fixed

PR #2178, opened 2020-06-08
add auditing

wanjunlei:
What type of PR is this?
/kind feature

This PR will add auditing in ks-apiserver. Each request will generate an auditing event, and the event will be sent to kube-auditing-webhook by ks-apiserver.  The auditing rule determines which event will be persisted and which event will triggering an alert.

### DIFF
--- a/pkg/apiserver/auditing/backend.go
+++ b/pkg/apiserver/auditing/backend.go
@@ -1,0 +1,117 @@
+package auditing
+
+import (
+	"bytes"
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"k8s.io/client-go/listers/auditregistration/v1alpha1"
+	"k8s.io/klog"
+	"net/http"
+	"time"
+)
+
+const (
+	WaitTimeout = time.Second
+)
+
+type Backend struct {
+	lister          v1alpha1.AuditSinkLister
+	channelCapacity int
+	semCh           chan interface{}
+	cache           chan *EventList
+	client          http.Client
+	sendTimeout     time.Duration
+	waitTimeout     time.Duration
+}
+
+func NewBackend(channelCapacity int, cache chan *EventList, sendTimeout time.Duration, lister v1alpha1.AuditSinkLister) *Backend {
+
+	b := Backend{
+		semCh:           make(chan interface{}, channelCapacity),
+		channelCapacity: channelCapacity,
+		waitTimeout:     WaitTimeout,
+		cache:           cache,
+		sendTimeout:     sendTimeout,
+		lister:          lister,
+	}
+
+	b.client = http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
+			},
+		},
+		Timeout: b.sendTimeout,
+	}
+
+	go b.worker()
+
+	return &b
+}
+
+func (b *Backend) worker() {
+
+	for {
+		event := <-b.cache
+		if event == nil {
+			break
+		}
+
+		send := func(event *EventList) {
+			ctx, cancel := context.WithTimeout(context.Background(), b.waitTimeout)
+			defer cancel()
+
+			select {
+			case <-ctx.Done():
+				klog.Errorf("get goroutine for audit(%s) timeout", event.Items[0].AuditID)
+				return
+			case b.semCh <- struct{}{}:
+			}
+
+			defer func() {
+				<-b.semCh
+			}()
+
+			bs, err := json.Marshal(event)
+			if err != nil {
+				klog.Errorf("json marshal error, %s", err)
+				return
+			}
+
+			response, err := b.client.Post(b.getWebhookUrl(), "application/json", bytes.NewBuffer(bs))
+			if err != nil {
+				klog.Errorf("send audit event[%s] error, %s", event.Items[0].AuditID, err)
+				return
+			}
+
+			if response.StatusCode != http.StatusOK {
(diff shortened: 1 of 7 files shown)


pkg/simple/client/auditing/elasticsearch/options.go
```diff
@@ -47,6 +48,7 @@ func (s *Options) Validate() []error {
 }
 
 func (s *Options) AddFlags(fs *pflag.FlagSet, c *Options) {
+	fs.BoolVar(&s.Enabled, "auditing-enabled", c.Enabled, "Dose auditing enabled")
```
zryfish, line 51: Typo `Does`,  I think we can following description.
```suggestion
	fs.BoolVar(&s.Enabled, "auditing-enabled", c.Enabled, "Enable auditing component or not.")
```

pkg/simple/client/auditing/elasticsearch/options.go
```diff
@@ -22,6 +22,7 @@ import (
 )
 
 type Options struct {
+	Enabled     bool   `json:"enabled" yaml:"enable"`
```
zryfish, line 25: Why json tag is `enabled` while yam tag is `enable`, this is misleading. I think `enable` will be just fine.

pkg/simple/client/auditing/elasticsearch/options.go
```diff
@@ -47,6 +48,7 @@ func (s *Options) Validate() []error {
 }
 
 func (s *Options) AddFlags(fs *pflag.FlagSet, c *Options) {
+	fs.BoolVar(&s.Enabled, "auditing-enabled", c.Enabled, "Dose auditing enabled")
 	fs.StringVar(&s.Host, "auditing-elasticsearch-host", c.Host, ""+
 		"Elasticsearch service host. KubeSphere is using elastic as auditing store, "+
 		"if this filed left blank, KubeSphere will use kubernetes builtin event API instead, and"+
```
zryfish, line 54: Have we already implemented this, use built-in event API when our auditing component not enabled?

pkg/apiserver/filters/auditing.go
```diff
@@ -0,0 +1,44 @@
...
+	"kubesphere.io/kubesphere/pkg/apiserver/auditing"
+	"kubesphere.io/kubesphere/pkg/apiserver/request"
+	"net/http"
+)
+
+func WithAuditing(handler http.Handler, a auditing.Auditing) http.Handler {
+
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+
+		if !a.Enable() {
```
zryfish, line 14: it's unnecessary to do a check here, just return `handler` when auditing is not enabled. Check out the code above

pkg/apiserver/filters/auditing.go
```diff
@@ -0,0 +1,44 @@
+package filters
+
+import (
+	"k8s.io/klog"
+	"kubesphere.io/kubesphere/pkg/apiserver/auditing"
+	"kubesphere.io/kubesphere/pkg/apiserver/request"
+	"net/http"
+)
+
+func WithAuditing(handler http.Handler, a auditing.Auditing) http.Handler {
```
zryfish, line 10:
```suggestion
func WithAuditing(handler http.Handler, a auditing.Auditing) http.Handler {
   if !a.Enable() {
      return handler
   }
}
```

pkg/apiserver/filters/auditing.go
```diff
@@ -0,0 +1,44 @@
...
+
+		if !a.Enable() {
+			handler.ServeHTTP(w, req)
+			return
+		}
+
+		info, ok := request.RequestInfoFrom(req.Context())
+		if !ok {
+			klog.Error("Unable to retrieve request info from request")
+			handler.ServeHTTP(w, req)
```
zryfish, line 22: If we can't retrieve request info from context, why just let the request pass, shouldn't we return error ?

pkg/apiserver/filters/auditing.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+
+		if info.IsKubernetesRequest {
+			handler.ServeHTTP(w, req)
+			return
+		}
+
+		e := a.LogRequestObject(info, req)
+		resp := auditing.NewResponseCapture(w)
+		go handler.ServeHTTP(resp, req)
```
zryfish, line 33: Why put handler.ServeHTTP in a goroutine, why we need to wait from resp.StopCh ?
You can refer to https://github.com/kubernetes/apiserver/blob/master/pkg/endpoints/filters/audit.go#L77 to make sure `audit LogResponseObject` is always called after finish serving requests?

pkg/apiserver/auditing/backend.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	go b.worker()
+
+	return &b
+}
+
+func (b *Backend) worker() {
+
+	for {
```
zryfish, line 56: how do we end this loop? There is only one condition to end this loop `event == nil`, but when program is gracefully terminated, how can we end this loop gracefully?

pkg/apiserver/auditing/config.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"gopkg.in/yaml.v3"
+	"io/ioutil"
+	"k8s.io/klog"
+	"os"
+	"regexp"
+)
+
+const (
+	ConfigFile = "/etc/kubesphere/auditing/config"
```
zryfish, line 12: Is there a default auditing config file?

wanjunlei:
There has a big change, i will resubumit this pr.

PR #2190, opened 2020-06-10
only accept validated CIDR in namespace networkpolicy

duanjiong:
**What type of PR is this?**

/kind bug

And fix the error handle when network-isotate is diabled.
And remove the useless crd.

Fixes #

```
```

```docs

```


zheng1:
Please change the title to the right description. Like this: `Bugfix: only accept validated CIDR in network policy `

PR #2193, opened 2020-06-11
fix apiresult

shaowenchen:

**What type of PR is this?**
 /kind bug

It returns items=nil，when there is no data for querying.

### DIFF
--- a/pkg/models/devops/devops.go
+++ b/pkg/models/devops/devops.go
@@ -175,6 +175,9 @@ func (d devopsOperator) ListDevOpsProject(workspace string, limit, offset int) (
 		limit = len(result) - offset
 	}
 	items = result[offset : offset+limit]
+	if items == nil {
+		items = []interface{}{}
+	}
 	return api.ListResult{TotalItems: len(result), Items: items}, nil
 }
 
@@ -218,7 +221,7 @@ func (d devopsOperator) ListPipelineObj(projectName string, limit, offset int) (
 	}
 	data, err := d.ksInformers.Devops().V1alpha3().Pipelines().Lister().Pipelines(projectObj.Status.AdminNamespace).List(labels.Everything())
 	if err != nil {
-		return api.ListResult{}, nil
+		return api.ListResult{}, err
 	}
 	items := make([]interface{}, 0)
 	var result []interface{}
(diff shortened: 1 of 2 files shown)


pkg/models/devops/devops.go
```diff
@@ -175,6 +175,9 @@ func (d devopsOperator) ListDevOpsProject(workspace string, limit, offset int) (
 		limit = len(result) - offset
 	}
 	items = result[offset : offset+limit]
+	if items == nil {
```
zryfish, line 178: why items is nil since you've already initialized in line https://github.com/kubesphere/kubesphere/pull/2193/files#diff-2c85fb739a144b30d0a086122b84ab75R168

shaowenchen: When return nil , It will work.  But , this function return {Items:nil, TotalItems:0}.

zryfish: What I mean is `items` can't be nil, since you've already initialized above,  statement `items = result[offset : offset+limit]` also can't lead to a nil `items`. So if this function returns a nil `items`, I think maybe somewhere else.

pkg/models/devops/devops.go
```diff
@@ -218,7 +221,7 @@ func (d devopsOperator) ListPipelineObj(projectName string, limit, offset int) (
 	}
 	data, err := d.ksInformers.Devops().V1alpha3().Pipelines().Lister().Pipelines(projectObj.Status.AdminNamespace).List(labels.Everything())
 	if err != nil {
-		return api.ListResult{}, nil
+		return api.ListResult{}, err
```
zryfish, line 224: How about this one? It probably returns a nil items result.

shaowenchen: This api.ListResult{} will not be used ，just for return value type check.

pkg/models/devops/devops.go
shaowenchen: when result is nil ，item wil get a nil value.

zryfish: If result is nil, what do you think will happen to this statement
```
items = result[offset: offset+limit]
```

shaowenchen: nill[0:0] is ok.

zryfish: OK，just tried, it still passes when result is nil, that's wired

PR #2195, opened 2020-06-11
Refactor with OpenPitrix

zheng1:
**What type of PR is this?**
>
/kind api-change
/kind cleanup
/kind feature

- Refactor OpenPitrix and multi-cluster
- Add OpenPitrix repo event apis
- Add OpenPitrix app upgrade apis

Fixes #

```
```

```docs

```


go.mod
```diff
@@ -56,30 +51,28 @@ require (
...
 	github.com/opencontainers/go-digest v1.0.0-rc1
-	github.com/opencontainers/image-spec v1.0.1 // indirect
 	github.com/openshift/api v0.0.0-20180801171038-322a19404e37 // indirect
 	github.com/opentracing/opentracing-go v1.1.0 // indirect
 	github.com/patrickmn/go-cache v2.1.0+incompatible // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/projectcalico/kube-controllers v3.8.8+incompatible
 	github.com/projectcalico/libcalico-go v1.7.2-0.20191104213956-8f81e1e344ce
-	github.com/prometheus/client_golang v1.0.0
+	github.com/prometheus/client_golang v1.2.1
```
zryfish, line 64: Why do we need to bump prometheus client_golang version, I may break kiali dependency @huanggze

pkg/controller/namespace/namespace_controller.go
```diff
@@ -153,13 +145,6 @@ func (r *ReconcileNamespace) Reconcile(request reconcile.Request) (reconcile.Res
 		return reconcile.Result{}, err
 	}
 
-	// skip if openpitrix is not enabled
-	if r.openpitrixClient != nil {
```
zryfish, line 157: Since openpitrix client is no longer needed in namespace controller, we'd better remove it from recocileNamespace fields list and initializer signature as well.

zheng1: Do you mean this commit? https://github.com/kubesphere/kubesphere/pull/2195/commits/3778abf767d416ea1e680d2c965de51a66cceec8

zryfish: yes, exactly

go.mod
zryfish: wired, don't know why update-vendor.sh bump this version and leave replace section.

PR #2205, opened 2020-06-16
ignore jenkins error

shaowenchen:

**What type of PR is this?**

> /kind bug

It displays CRD data ，not Jenkins project or credential in KubeSphere Console. It is unnecessary 
  to maintain strong consistency.


pkg/controller/devopscredential/devopscredential_controller.go
```diff
@@ -239,7 +239,6 @@ func (c *Controller) syncHandler(key string) error {
 		if sliceutil.HasString(copySecret.ObjectMeta.Finalizers, devopsv1alpha3.CredentialFinalizerName) {
 			if _, err := c.devopsClient.DeleteCredentialInProject(nsName, secret.Name); err != nil {
 				klog.V(8).Info(err, fmt.Sprintf("failed to delete secret %s in devops", key))
-				return err
```
zryfish, line 242: Is it ok to just leave the secrets sit in the Jenkins? If that's the case, please add more comments to explain why we do this

PR #2227, opened 2020-06-18
process audit information for resource creating requests

wanjunlei:

What type of PR is this?
> /kind feature

Get resource name from request body when request is a resource creating request and the resource is k8s resource or crd.

### DIFF
--- a/pkg/apiserver/auditing/backend.go
+++ b/pkg/apiserver/auditing/backend.go
@@ -112,24 +112,18 @@ func (b *Backend) worker() {
 
 func (b *Backend) eventToBytes(event *v1alpha1.EventList) ([]byte, error) {
 
-	if bs, err := json.Marshal(event); err == nil {
-		return bs, nil
-	}
-
-	// Normally, the serialization failure is caused by the failure of RequestObject or ResponseObject serialization.
-	// To ensure the integrity of the auditing event to the greatest extent,
-	// it is necessary to delete RequestObject or ResponseObject and and then try to serialize again.
-	if event.Items[0].RequestObject != nil {
-		if _, err := json.Marshal(event.Items[0].RequestObject); err != nil {
-			event.Items[0].RequestObject = nil
-		}
-	}
-
-	if event.Items[0].ResponseObject != nil {
-		if _, err := json.Marshal(event.Items[0].ResponseObject); err != nil {
+	bs, err := json.Marshal(event)
+	if err != nil {
+		// Normally, the serialization failure is caused by the failure of ResponseObject serialization.
+		// To ensure the integrity of the auditing event to the greatest extent,
+		// it is necessary to delete ResponseObject and and then try to serialize again.
+		if event.Items[0].ResponseObject != nil {
 			event.Items[0].ResponseObject = nil
+			return json.Marshal(event)
 		}
+
+		return nil, err
 	}
 
-	return json.Marshal(event)
+	return bs, err
 }
--- a/pkg/apiserver/auditing/backend.go
+++ b/pkg/apiserver/auditing/backend.go
@@ -86,11 +86,13 @@ func (b *Backend) worker() {
 				<-b.semCh
 			}()
 
-			bs, err := json.Marshal(event)
+			bs, err := b.eventToBytes(event)
 			if err != nil {
-				klog.Errorf("json marshal error, %s", err)
+				klog.V(6).Infof("json marshal error, %s", err)
 				return
 			}
+
+			klog.V(8).Infof("%s", string(bs))
 
 			response, err := b.client.Post(b.url, "application/json", bytes.NewBuffer(bs))
 			if err != nil {
@@ -106,4 +108,28 @@ func (b *Backend) worker() {
 
 		go send(event)
 	}
+}
+
+func (b *Backend) eventToBytes(event *v1alpha1.EventList) ([]byte, error) {
+
+	if bs, err := json.Marshal(event); err == nil {
+		return bs, nil
+	}
+
+	// Normally, the serialization failure is caused by the failure of RequestObject or ResponseObject serialization.
+	// To ensure the integrity of the auditing event to the greatest extent,
+	// it is necessary to delete RequestObject or ResponseObject and and then try to serialize again.
+	if event.Items[0].RequestObject != nil {
+		if _, err := json.Marshal(event.Items[0].RequestObject); err != nil {
+			event.Items[0].RequestObject = nil
+		}
+	}
+
+	if event.Items[0].ResponseObject != nil {
+		if _, err := json.Marshal(event.Items[0].ResponseObject); err != nil {
+			event.Items[0].ResponseObject = nil
+		}
+	}
+
+	return json.Marshal(event)
 }
--- a/pkg/kapis/devops/v1alpha2/devops.go
+++ b/pkg/kapis/devops/v1alpha2/devops.go
@@ -20,7 +20,6 @@ import (
 	"github.com/emicklei/go-restful"
 	log "k8s.io/klog"
 	"kubesphere.io/kubesphere/pkg/api"
-	"kubesphere.io/kubesphere/pkg/apiserver/request"
 	"kubesphere.io/kubesphere/pkg/models/devops"
 	"net/http"
 	"strings"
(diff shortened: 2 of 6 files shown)


benjaminhuo:
@zryfish  is this PR ok for you?

pkg/kapis/devops/v1alpha2/devops.go
```diff
@@ -92,6 +93,12 @@ func (h *ProjectPipelineHandler) StopPipeline(req *restful.Request, resp *restfu
 		return
 	}
 
+	if event := request.AuditEventFrom(req.Request.Context()); event != nil {
```
zryfish, line 96: Why auditing has to collect events inside devops module, while all the information needed can be extracted from request?

pkg/kapis/iam/v1alpha2/handler.go
```diff
@@ -980,6 +997,14 @@ func (h *iamHandler) CreateClusterMembers(request *restful.Request, response *re
 		}
 	}
 
+	if event := requestinfo.AuditEventFrom(request.Request.Context()); event != nil {
```
zryfish, line 1000: Same here, I think we'd better put auditing aside from these modules, to do better decoupling.

pkg/kapis/iam/v1alpha2/handler.go
```diff
@@ -899,6 +908,14 @@ func (h *iamHandler) CreateNamespaceMembers(request *restful.Request, response *
...
```
zryfish, line 911: Auditing happens all the operations are successful, but what if some of the members are created successfully, others are not? Don't we audit every MODIFY operation even there are errors?

pkg/apiserver/auditing/backend.go
```diff
@@ -107,3 +109,27 @@ func (b *Backend) worker() {
...
+func (b *Backend) eventToBytes(event *v1alpha1.EventList) ([]byte, error) {
+
+	if bs, err := json.Marshal(event); err == nil {
+		return bs, nil
+	}
+
+	// Normally, the serialization failure is caused by the failure of RequestObject or ResponseObject serialization.
+	// To ensure the integrity of the auditing event to the greatest extent,
+	// it is necessary to delete RequestObject or ResponseObject and and then try to serialize again.
+	if event.Items[0].RequestObject != nil {
```
zryfish, line 122: We need to come up with a better solution, in the worst case, you have to marshal four times to get the result, which is very time-consuming.

pkg/kapis/devops/v1alpha2/devops.go
wanjunlei: The most auditing information of most APIs can be collected through request and response in auditing component. For a few non-standard APIs, some audit information needs to be obtained through special processing of request and response, such as parsing the resource name from the response. If these special processing are placed in auditing component, then each request will match the request url at least once. Putting special processing of a few non-standard apis into the api handler will reduce the performance impact of the audit on the apiserver.

pkg/apiserver/auditing/backend.go
wanjunlei: Request body of all current apis are json formate, only response body of some few apis are not json formate, so i will delete the serializable detection when serialization failure，directly set ResponseObject to null, and then try serialization again. This is a compromise between data integrity and performance. I will consider how to optimize it here later.

pkg/kapis/devops/v1alpha2/devops.go
zryfish: Auditing should happen outside of the business module, decoupled with them, try to think what if we add more modules here(what we will do in v4.0, third party modules), do we need to write auditing logic for each of them inside that module? We will figure out other ways to reduce performance impact, but mixing auditing code with a specific module is not a good path we are going to take.  For non-standard APIs, we can make them become standard step by step.

wanjunlei: If all apis are standard, auditing will happen outside of the business module, decoupled with them. But there was som few non-standard APIs in 3.0, it is unreasonable to affect the performance of most standard APIs because of a small number of non-standard APIs, 
API should bear the responsibility of providing audit information because it is non-standard. 
This can also promote the standardization process of apis. 
We can optimize or even reconstruct auditing components in 3.1, or when all interfaces are standardized in 4.0, this problem does not exist. This is currently the fastest way to solve this problem.

zryfish: Are there any statistics or experiment data on how much performance impact we have to suffer if we do it the other way?

wanjunlei: We can test it in 3.1 to see if there are other solutions. The main goal should now be to release version 3.0.

zryfish: I think we should do it right in the first place, the current method is highly coupled with specific modules. @rayzhou2017 how do you think? Should we change it now or leave it to next version?

rayzhou2017: How long will it take if we decouple them, @wanjunlei ? It is NOT good to invade into other business code in terms of architecture. It is like logging. In cloud native application, it throws  logs into console, then logging component is responsible to pick up the logs. The component never cracks into the application to get logs. The same principle should go with auditing in my opinion. It must be decoupling. The only thing is how long it takes that we need to evaluate.

rayzhou2017: /cc @benjaminhuo please also help check this issue

rayzhou2017: Would a sidecar solve the problem?

benjaminhuo: If it depends on the standardization of KS API， then maybe we can just not add audit for non-standard API in KS v3.0. And create an issue to track or discuss with @wansir to standardize KS API.

For now, we can just add audit for standard KS API.

@wanjunlei @zryfish @rayzhou2017

rayzhou2017: OK, I got you. Now I have a suggestion. How serious it will impact in terms of performance for now? If it is not too bad, we can accept the current solution right now since it audits the components which is necessary for users. BUT, we need to create an issue tracking this PR for code refactoring later on. On the other hand, if it affects performance a lot, then do not accept the current solution, and create an issue to track the problem as well. So make a decision.

benjaminhuo:
@rayzhou2017 @zryfish , discussed with @wanjunlei . He will re-submit a PR in which the invasion audit will be removed. For auditing of non-standard API, @wanjunlei  will create a proposal to elaborate on the details to do this from audit module side with configured rules and then discuss with us

pkg/kapis/devops/v1alpha2/devops.go
benjaminhuo: @rayzhou2017 According to @wanjunlei , current solution in this PR have minimum performance impact , but it's indeed a kind of invasion to devops code. The alternative method is to move these non-standard audit to audit module and configure rules for these non-standard components apis and this will impact the performance a bit because audit module has to find out what kind of request it is auditing and then choose corresponding auditing methods.

benjaminhuo: Of course, the better way is to have standard KS APIs for all reqeusts. But most likely this is a long process and we need to discuss with @wansir  or @shaowenchen  together

rayzhou2017: using rules will decouple them, right? Is it the final decoupling solution?

wanjunlei: This is one of decoupling solution,i will perfect this solution and discuss it later.

wanjunlei: Now I will delete the code inside devops module and iam module. This pr only handler one situation，
get resource name from request body when request is a resource creating request and the resource is k8s resource or crd. 
The non-standard apis will be handled later.

benjaminhuo:
@zryfish @rayzhou2017 , @wanjunlei has remove devops related invasion audit code and changed this PR's description. Pls review it again

benjaminhuo:
@zryfish The travis-ci has been blocked for a long time, you know the reason?

PR #2253, opened 2020-06-24
fix wrong total audit event number

wanjunlei:
Signed-off-by: wanjunlei wanjunlei@yunify.com

What type of PR is this?
> /kind bug

Change total in histogram response to auditing  events count instead of buckets count when search auditing with param operation=histogram

### DIFF
--- a/pkg/models/tenant/tenant.go
+++ b/pkg/models/tenant/tenant.go
@@ -702,10 +702,9 @@ func (t *tenantOperator) Auditing(user user.Info, queryParam *auditingv1alpha1.Q
 		listEvts := authorizer.AttributesRecord{
 			User:            user,
 			Verb:            "list",
-			APIGroup:        "",
-			APIVersion:      "v1",
 			Resource:        "namespaces",
 			ResourceRequest: true,
+			ResourceScope:   request.ClusterScope,
 		}
 		decision, _, err := t.authorizer.Authorize(listEvts)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


benjaminhuo:
The code change is good, pls change this PR's title from `fix bug total result of auditing time topology error #2253` to `fix wrong total audit event number` @wanjunlei

pkg/models/tenant/tenant.go
```diff
@@ -702,10 +702,9 @@ func (t *tenantOperator) Auditing(user user.Info, queryParam *auditingv1alpha1.Q
 		listEvts := authorizer.AttributesRecord{
 			User:            user,
 			Verb:            "list",
-			APIGroup:        "",
-			APIVersion:      "v1",
 			Resource:        "namespaces",
 			ResourceRequest: true,
+			ResourceScope:   request.ClusterScope,
```
benjaminhuo, line 707: No need to add NamespaceScope as events does?

benjaminhuo, line 707: This looks like a different PR actually

wanjunlei: Delete modification of permission and submit as a pr separately

wanjunlei: Delete modification of permission and submit as a pr separately

benjaminhuo:
@zryfish pls review this PR

PR #2264, opened 2020-06-28
fix: password modify

### LINKED ISSUE #2242: Failed to modify admin password in KS 3.0.0
**Describe the Bug**
Modify admin password on new installed KS 3.0.0, UI popup 'Bad Request' error and password is not modified.

image

**Versions Used**
KubeSphere: v3.0.0

**Environment**

For example: Ubuntu 18.04.3 / 1 masters:  4cpu/8g

**How To Reproduce**
1. Download kk
2. Run ./kk create config --with-kubesphere
3. Modify config yaml to fit cluster config
4. After installation success, login KS UI with admin and modify admin password

**Expected behavior**
No error occurs and password modified as requested.

wansir:

**What type of PR is this?**

/kind bug

 fix password modify API.

Fixes #2242


pkg/kapis/oauth/handler.go
```diff
@@ -212,23 +204,59 @@ func (h *oauthHandler) OAuthCallBackHandler(req *restful.Request, resp *restful.
...
+	h.im.AuthSuccessRecord(user.GetName(), result.AccessToken)
+	resp.WriteEntity(result)
+}
+
+func (h *oauthHandler) Login(request *restful.Request, response *restful.Response) {
+	var loginRequest auth.LoginRequest
+
+	err := request.ReadEntity(&loginRequest)
+	if err != nil || loginRequest.Username == "" || loginRequest.Password == "" {
+		response.WriteHeaderAndEntity(http.StatusUnauthorized, fmt.Errorf("incorrect username or password"))
```
zryfish, line 222: Shouldn't the error message be "empty username or password" ?

pkg/models/iam/im/im.go
```diff
@@ -170,6 +206,101 @@ func (im *defaultIMOperator) CreateUser(user *iamv1alpha2.User) (*iamv1alpha2.Us
...
+		Reason:             "Authentication Failed",
+		Message:            err.Error(),
+	}
+	if user.Status.Conditions == nil {
+		user.Status.Conditions = make([]iamv1alpha2.UserCondition, 0)
+	}
+	user.Status.Conditions = append(user.Status.Conditions, condition)
+	// remove out date and out range record
+	record := 0
+	for i := len(user.Status.Conditions); i > 0; i-- {
```
zryfish, line 249: If you just want to remove condition out of slice, it's easy to understand if create another slice,  and copy only conditions needed over.

pkg/kapis/oauth/handler.go
wansir: fixed

pkg/models/iam/im/im.go
wansir: fixed

wansir:
fix: https://github.com/kubesphere/kubesphere/issues/2275

pkg/models/iam/im/im.go
zryfish: I don't think it's appropriate to save login attempts in the conditions. It's better to create another object, say Login, and logs every login attempt as a Login object. If we save login attempts as conditions, it's not easy to proceed, and user objects can be very big. For example, what if set maxFailureLoginAttempts to 1000.

pkg/models/iam/im/im.go
```diff
@@ -170,6 +206,108 @@ func (im *defaultIMOperator) CreateUser(user *iamv1alpha2.User) (*iamv1alpha2.Us
...
+		}
+	}
+	user.Status.Conditions = conditions
+	_, err = im.ksClient.IamV1alpha2().Users().Update(user)
+	if err != nil {
+		klog.Warning(err)
+	}
+}
+
+func (im *defaultIMOperator) AuthSuccessRecord(username string, token string) {
```
zryfish, line 270: same here, for example, create a Login object for every login attempt, count consecutive failed login attempts
```
kind: Login
metadata:
  name: admin-xxxx
  createTimestamp: 2020-07-01T14:00:00
  labels:
    user: admin
  ownerReference:
     kind: User
spec:
  sourceIP: xxx.xxx.xxx.xxx
  type: Succeess
  
```

pkg/models/iam/im/im.go
wansir: I think login record is also a part of user condition, we can easily get the user's status, it is necessary to set a reasonable maxFailureLoginAttempts, just like the resource limit needs a reasonable value. If maxFailureLoginAttempts is not needed, it can be set to zero. Even if maxFailureLoginAttempts is large than 1000, the user list response will not output `user.status.Conditions`. Tell the user how to set a reasonable value is better than adding a CRD.

zryfish: Why is here no user.status.Conditions in response?

wansir: `user.status.Conditions` is redundant in the user list. Set `user.status.Conditions` to nil before the API(`/kapis/ iam.kubesphere.io/v1alpha2/users`) returns.

zryfish: What you did here is just filter response returned to console, the status is still there in the kubernetes API response

PR #2270, opened 2020-06-29
fix: synchronize users to LDAP

### LINKED ISSUE #2198: Add  user to ldap
```bash
org.acegisecurity.userdetails.UsernameNotFoundException: User admin not found in directory.
	at org.acegisecurity.ldap.search.FilterBasedLdapUserSearch.searchForUser(FilterBasedLdapUserSearch.java:126)
	at hudson.security.LDAPSecurityRealm$LDAPUserDetailsService.loadUserByUsername(LDAPSecurityRealm.java:1314)
	at hudson.security.LDAPSecurityRealm$LDAPUserDetailsService.loadUserByUsername(LDAPSecurityRealm.java:1251)
	at jenkins.security.ImpersonatingUserDetailsService.loadUserByUsername(ImpersonatingUserDetailsService.java:30)
	at hudson.model.User.getUserDetailsForImpersonation(User.java:394)
```

DevOps use LDAP to manager users. It is necessary to sync users from Auth to OpenLDAP

version: 20200613 installer 
/area  iam
/assign @wansir

wansir:

**What type of PR is this?**

 /kind bug

Synchronize users to LDAP.

Fixes #2198

### DIFF
--- a/pkg/controller/user/user_controller.go
+++ b/pkg/controller/user/user_controller.go
@@ -43,6 +43,8 @@ import (
 	userlister "kubesphere.io/kubesphere/pkg/client/listers/iam/v1alpha2"
 	"kubesphere.io/kubesphere/pkg/constants"
 	"kubesphere.io/kubesphere/pkg/models/kubeconfig"
+	ldapclient "kubesphere.io/kubesphere/pkg/simple/client/ldap"
+	"kubesphere.io/kubesphere/pkg/utils/sliceutil"
 	"reflect"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"strconv"
@@ -67,6 +69,7 @@ type Controller struct {
 	cmSynced          cache.InformerSynced
 	fedUserCache      cache.Store
 	fedUserController cache.Controller
+	ldapClient        ldapclient.Interface
 	// workqueue is a rate limited work queue. This is used to queue work to be
 	// processed instead of performing it as soon as a change happens. This
 	// means we can ensure we only process a fixed amount of resources at a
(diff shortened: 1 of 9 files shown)


pkg/controller/user/user_controller.go
```diff
@@ -239,6 +243,48 @@ func (c *Controller) reconcile(key string) error {
...
+			if _, err := c.ksClient.IamV1alpha2().Users().Update(user); err != nil {
+				return err
+			}
+		}
+
+		// Our finalizer has finished, so the reconciler can do nothing.
+		return nil
+	}
+
+	if err = c.ldapSync(user); err != nil {
```
zryfish, line 283: What if ldapClient is nil ?

pkg/controller/user/user_controller.go
```diff
@@ -419,6 +464,25 @@ func (c *Controller) updateFederatedUser(fedUser *iamv1alpha2.FederatedUser) err
 	return nil
 }
 
+func (c *Controller) ldapSync(user *iamv1alpha2.User) error {
+	encrypted, _ := strconv.ParseBool(user.Annotations[iamv1alpha2.PasswordEncryptedAnnotation])
+	if encrypted {
```
zryfish, line 469: Don't sync user if the password is encrypted, why ?

wansir: Use encrypted password cannot logged in through LDAP. Jenkins authorization mode `Logged in users can do anything` just lookup authenticated user in LDAP, user password  is not important, because regular users are not allowed to log in to Jenkins directly.

pkg/controller/user/user_controller.go
zryfish: Is any comments on this?

wansir: LDAP needs to be enabled during installation, must not be nil. Use simple ldap client in debug mode.

zryfish: Then we'd better check client during controller initialization, make sure we don't get a nil ldap client

wansir: Is it necessary to use simple LDAP client if LDAP options not provide?

zryfish: No, if ldap not provided, just return error. Simple Ldap is only used for testing

wansir: done

PR #2287, opened 2020-07-02
lazy initializing es client

huanggze:

**What type of PR is this?**
 /kind api-change

Fixes #

```
```

```docs

```

### DIFF
--- a/pkg/simple/client/logging/elasticsearch/elasticsearch.go
+++ b/pkg/simple/client/logging/elasticsearch/elasticsearch.go
@@ -12,6 +12,7 @@ import (
 	"kubesphere.io/kubesphere/pkg/simple/client/logging/elasticsearch/versions/v7"
 	"kubesphere.io/kubesphere/pkg/utils/stringutils"
 	"strings"
+	"sync"
 )
 
 const (
@@ -22,7 +23,12 @@ const (
 
 // Elasticsearch implement logging interface
 type Elasticsearch struct {
-	c client
+	host    string
+	version string
+	index   string
+
+	c   client
+	mux sync.Mutex
 }
 
 // versioned es client interface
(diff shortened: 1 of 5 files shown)


pkg/simple/client/logging/elasticsearch/elasticsearch.go
```diff
@@ -34,76 +40,97 @@ type client interface {
...
+		c, err = v5.New(es.host, es.index)
+	case ElasticV6:
+		c, err = v6.New(es.host, es.index)
+	case ElasticV7:
+		c, err = v7.New(es.host, es.index)
+	default:
+		return fmt.Errorf("unsupported elasticsearch version %s", version)
+	}
+
+	es.mux.Lock()
```
wanjunlei, line 118: Why it only lock the assignment statement? This will cause multiple goroutines to initialize es client at the same time.

huanggze: true

benjaminhuo:
/assign @zryfish 
@wanjunlei @junotx pls sync the change to audit and event search, and we should refactor logging, event and audit code later to use one single es search code for all these 3 components.

pkg/simple/client/logging/elasticsearch/elasticsearch.go
```diff
@@ -34,76 +40,102 @@ type client interface {
...
 	}
 
-	return es, nil
+	return es, err
 }
 
-func (es *Elasticsearch) ES() *client {
-	return &es.c
+func (es *Elasticsearch) getClient() client {
+	es.mux.RLock()
```
zryfish, line 67: This is a typical singleton with lazy load,  there is a great post http://marcio.io/2015/07/singleton-pattern-in-go/

huanggze: I have seen these kind of posts around. The problem is that `sync.Once` performs exactly once and no more. What if the init task fails?

zryfish: OK, I see.

zryfish: It's better to initialize client in one place, such as
```suggestion
	func (es *Elasticsearch) getClient() (client, error) {
	  if es.c != nil {
	    return es.c, nil
	  }
	  
	  es.mux.Lock()
	  defer es.mu.Unlock()
	    err = es.initClient()
	    if err == nil && es.c != nil {
	      return es.c, nil
	    }
	    
	    return nil, err
	  
	}
```

benjaminhuo: to be more clear, maybe we could add a else branch:
```
	  if es.c != nil {
	    return es.c, nil
	  } else {
	    es.mux.Lock()
	    defer es.mu.Unlock()
	    err = es.initClient()
	    if err == nil && es.c != nil {
	      return es.c, nil
	    }
	    return nil, err
	  }
	}
```

huanggze: I put code comments to clarify this issue.

huanggze: I put code comments to clarify this issue.

benjaminhuo: @huanggze The new change looks better and pls ignore my previous code snippet, defer is only call at the end of a function, not the end of a scope

benjaminhuo:
@wanjunlei @junotx pls merge this PR for event and auditing

PR #2290, opened 2020-07-03
fix: implicit workspace info in query

wansir:

**What type of PR is this?**

/kind bug

Parse implicit workspace info in query params.

Workspace resource filter: `/api/clusters/m15-ap2/v1/watch/namespaces?labelSelector=kubesphere.io/workspace=demo-fed-ws`

Fixes #

```
```

```docs

```

### DIFF
--- a/pkg/apiserver/request/requestinfo.go
+++ b/pkg/apiserver/request/requestinfo.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog"
 	"kubesphere.io/kubesphere/pkg/api"
+	"kubesphere.io/kubesphere/pkg/constants"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


pkg/apiserver/request/requestinfo.go
```diff
@@ -193,6 +194,12 @@ func (r *RequestInfoFactory) NewRequestInfo(req *http.Request) (*RequestInfo, er
 		}
 	}
 
+	// URL forms: /api/v1/watch/namespaces?labelSelector=kubesphere.io/workspace=system-workspace
+	workspaceSelector := req.URL.Query().Get("labelSelector")
+	if groups := regexp.MustCompile(`^kubesphere.io/workspace=(\S+)$`).FindStringSubmatch(workspaceSelector); len(groups) == 2 {
```
zryfish, line 199: regex expression can be slow, and request info is in a critical code path, I'm not confident to put a regex expression here. And also, workspaceSelector can be empty, if then there is no need to do regex match

wansir: fixed,  use  `strings.HasPrefix` instead of `regex expression`.

PR #2333, opened 2020-07-07
fix: resource lifecycle

### LINKED ISSUE #2352: the number of workspace members dont decrease after user has been deleted
we remove the user from page "access/accounts"

wansir:

**What type of PR is this?**

/kind bug

Release resources after the user is deleted.

Fixes #2352

```
```

```docs

```

### DIFF
--- a/pkg/controller/user/user_controller.go
+++ b/pkg/controller/user/user_controller.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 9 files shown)


pkg/controller/user/user_controller.go
```diff
@@ -483,6 +489,59 @@ func (c *Controller) ldapSync(user *iamv1alpha2.User) error {
 	}
 }
 
+func (c *Controller) deleteRoleBindings(user *iamv1alpha2.User) error {
+	listOptions := metav1.ListOptions{LabelSelector: labels.SelectorFromSet(labels.Set{iamv1alpha2.UserReferenceLabel: user.Name}).String()}
+	if result, err := c.ksClient.IamV1alpha2().GlobalRoleBindings().List(listOptions); err != nil {
```
zryfish, line 494: No need to check the existence first, just issue the delete request, ignore the NotFoud error.

wansir: The logic like `kubectl delete <resources> -l  <key>=<value>`.

pkg/controller/user/user_controller.go
```diff
@@ -483,6 +523,88 @@ func (c *Controller) ldapSync(user *iamv1alpha2.User) error {
 	}
 }
 
+func (c *Controller) deleteRoleBindings(user *iamv1alpha2.User) error {
+
```
zryfish, line 527: This doesn't improve the situation, please use `DeleteCollection` in clientset, it's more efficient

wansir: fixed

zryfish:
Unit test failed, please have a look

PR #2368, opened 2020-07-09
using /v1alpha2/jenkins replace /jenkins.kubesphere.io api

shaowenchen:
**What type of PR is this?**
/kind api-change

There are some APIs forwarding by caddy in 2.1.1.  Now，we can use /kapis/devops.kubesphere.io/v1alpha2/jenkins to replace it.

### DIFF
--- a/pkg/kapis/devops/v1alpha2/register.go
+++ b/pkg/kapis/devops/v1alpha2/register.go
@@ -17,20 +17,25 @@ limitations under the License.
 package v1alpha2
 
 import (
+	"fmt"
 	"github.com/emicklei/go-restful"
 	"github.com/emicklei/go-restful-openapi"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/proxy"
+	"k8s.io/klog"
 	devopsv1alpha1 "kubesphere.io/kubesphere/pkg/apis/devops/v1alpha1"
 	"kubesphere.io/kubesphere/pkg/apiserver/runtime"
 	"kubesphere.io/kubesphere/pkg/client/clientset/versioned"
 	"kubesphere.io/kubesphere/pkg/client/informers/externalversions"
 	"kubesphere.io/kubesphere/pkg/constants"
+	"kubesphere.io/kubesphere/pkg/simple/client/devops/jenkins"
 	"kubesphere.io/kubesphere/pkg/simple/client/s3"
 	"kubesphere.io/kubesphere/pkg/simple/client/sonarqube"
+	"net/url"
+	"strings"
 
 	//"kubesphere.io/kubesphere/pkg/models/devops"
 	"kubesphere.io/kubesphere/pkg/simple/client/devops"
-
 	"net/http"
 )
 
@@ -41,7 +46,7 @@ const (
 
 var GroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1alpha2"}
 
-func AddToContainer(container *restful.Container, ksInformers externalversions.SharedInformerFactory, devopsClient devops.Interface, sonarqubeClient sonarqube.SonarInterface, ksClient versioned.Interface, s3Client s3.Interface) error {
+func AddToContainer(container *restful.Container, ksInformers externalversions.SharedInformerFactory, devopsClient devops.Interface, sonarqubeClient sonarqube.SonarInterface, ksClient versioned.Interface, s3Client s3.Interface, endpoint string) error {
 	ws := runtime.NewWebService(GroupVersion)
 
 	err := AddPipelineToWebService(ws, devopsClient)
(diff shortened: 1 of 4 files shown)


pkg/kapis/devops/v1alpha2/handler.go
```diff
@@ -34,6 +40,17 @@ type PipelineSonarHandler struct {
 	pipelineSonarGetter devops.PipelineSonarGetter
 }
 
+func jenkinsProxy(request *restful.Request, response *restful.Response) {
+	u := request.Request.URL
+	u.Host = "allinone.ks.dev.chenshaowen.com:30180"
```
zryfish, line 45: What's this? why the host has to be this ?

zryfish:
Why are we back to old days, proxying requests directly to Jenkins? I think we've already decoupled from Jenkins.

pkg/kapis/devops/v1alpha2/handler.go
shaowenchen: It's testing code. Removed.

shaowenchen:
> Why are we back to old days, proxying requests directly to Jenkins? I think we've already decoupled from Jenkins.

There are some APIs that get files from Jenkins，e.g download artifacts or test reports. These interfaces have no definite format but the same prefix. 

In 2.1, we proxy it using caddy in the API gateway. @zryfish

zryfish:
> > Why are we back to old days, proxying requests directly to Jenkins? I think we've already decoupled from Jenkins.
> 
> There are some APIs that get files from Jenkins，e.g download artifacts or test reports. These interfaces have no definite format but the same prefix.
> 
> In 2.1, we proxy it using caddy in the API gateway. @zryfish

Which APIs needed to proxy to Jenkins? What I really want to say is we shouldn't use Jenkins API directly, it's tightly coupled.

pkg/kapis/devops/v1alpha2/register.go
```diff
@@ -59,6 +64,11 @@ func AddToContainer(container *restful.Container, ksInformers externalversions.S
 		return err
 	}
 
+	err = AddJenkinsToContainer(ws, endpoint)
```
zryfish, line 67: If endpoint is empty, means Devops is not enabled, `AddJenkinsToContainer` will return an error

shaowenchen: fixed

shaowenchen:
> > > Why are we back to old days, proxying requests directly to Jenkins? I think we've already decoupled from Jenkins.
> > 
> > 
> > There are some APIs that get files from Jenkins，e.g download artifacts or test reports. These interfaces have no definite format but the same prefix.
> > In 2.1, we proxy it using caddy in the API gateway. @zryfish
> 
> Which APIs needed to proxy to Jenkins? What I really want to say is we shouldn't use Jenkins API directly, it's tightly coupled.

/job/new-devopsfvs8z/job/archive-zip/job/1/artifact/a/b/c/newfile.txt

If it's tightly coupled, we can add prefix /job/new-devopsfvs8z/job/archive-zip/job/1/artifact to check.

zryfish:
> > > > Why are we back to old days, proxying requests directly to Jenkins? I think we've already decoupled from Jenkins.
> > > 
> > > 
> > > There are some APIs that get files from Jenkins，e.g download artifacts or test reports. These interfaces have no definite format but the same prefix.
> > > In 2.1, we proxy it using caddy in the API gateway. @zryfish
> > 
> > 
> > Which APIs needed to proxy to Jenkins? What I really want to say is we shouldn't use Jenkins API directly, it's tightly coupled.
> 
> /job/new-devopsfvs8z/job/archive-zip/job/1/artifact/a/b/c/newfile.txt
> 
> If it's tightly coupled, we can add prefix /job/new-devopsfvs8z/job/archive-zip/job/1/artifact to check.

It's our job

pkg/kapis/devops/v1alpha2/register.go
```diff
@@ -685,3 +695,25 @@ func AddS2IToWebService(webservice *restful.WebService, ksClient versioned.Inter
...
+	}
+	parse.Path = strings.Trim(parse.Path, "/")
+	webservice.Route(webservice.GET("/jenkins/{path:*}").
+		To(func(request *restful.Request, response *restful.Response) {
+			u := request.Request.URL
+			u.Host = parse.Host
+			u.Scheme = parse.Scheme
+			jenkins.SetBasicBearTokenHeader(&request.Request.Header)
+			u.Path = strings.Replace(request.Request.URL.Path, fmt.Sprintf("/kapis/%s/%s/jenkins", GroupVersion.Group, GroupVersion.Version), "", 1)
+			httpProxy := proxy.NewUpgradeAwareHandler(u, http.DefaultTransport, false, false, nil)
```
zryfish, line 715: ErrorResponser can be useful for debugging, please don't pass nil

shaowenchen: fixed

PR #2408, opened 2020-07-13
modify the field networkIsolate in workspace

duanjiong:
**What type of PR is this?**

/kind bug

If use bool, this field will be ignored by fed controller, so change it to `* bool`

Fixes #

```
```

```docs

```


pkg/controller/network/nsnetworkpolicy/nsnetworkpolicy_controller.go
```diff
@@ -573,6 +573,13 @@ func (c *NSNetworkPolicyController) processNSNPWorkItem() bool {
 	return true
 }
 
+func workspaceNetworkIsolationEnabled(wksp *workspacev1alpha1.Workspace) bool {
```
zheng1, line 576: Please add some unit test for `workspaceNetworkIsolationEnabled` and `namespaceNetworkIsolateEnabled`

PR #2416, opened 2020-07-14
fix "Failed to list *v1beta1.VolumeSnapshotClass" on k8s 1.16

min-zh:
**What type of PR is this?**

 /kind bug

fix bug "Failed to list *v1beta1.VolumeSnapshotClass" on k8s 1.16

Fixes #

```
```

```docs

```

### DIFF
--- a/pkg/controller/storage/capability/capability_controller.go
+++ b/pkg/controller/storage/capability/capability_controller.go
@@ -101,6 +101,7 @@ func NewController(
 ) *StorageCapabilityController {
 
 	utilruntime.Must(crdscheme.AddToScheme(scheme.Scheme))
+
 	controller := &StorageCapabilityController{
 		storageClassCapabilityClient: capabilityClient,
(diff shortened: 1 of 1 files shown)


pkg/controller/storage/capability/capability_controller.go
```diff
@@ -356,7 +364,7 @@ func (c *StorageCapabilityController) deleteStorageCapability(name string) error
 }
 
 func (c *StorageCapabilityController) deleteSnapshotClass(name string) error {
-	if !c.snapshotSupported {
+	if !c.snapshotSupported || c.snapshotClassClient == nil {
```
zheng1, line 367: This line can be refine to an common func

min-zh: OK

min-zh: We already have unit test ‘TestCreateStorageClassNotSupportSnapshot ’ for this case.

PR #2423, opened 2020-07-14
significantly improve log search performance

huanggze:
Signed-off-by: huanggze <loganhuang@yunify.com>

**What type of PR is this?**

 /kind bug
 /kind cleanup

**What this PR does / why we need it**:
Instead of query against all indices, search indices that are within query time span only.

Example: 
```
# logging kapis: start_time=1594390613&end_time=1594638672

GET localhost:9200/ks-logstash-log*/_search

 --> 

GET localhost:9200/ks-logstash-log-2020.07.13,ks-logstash-log-2020.07.12,ks-logstash-log-2020.07.11,ks-logstash-log-2020.07.10/_search?ignore_unavailable=true
```

**Which issue(s) this PR fixes**:

Refer to KubeSphere forum: https://kubesphere.com.cn/forum/d/1074/15

**Special notes for reviewers**:
```
```

**Additional documentation, usage docs, etc.**:

```docs

```

PR #2441, opened 2020-07-16
lazy init events client and optimize events search

junotx:
**What type of PR is this?**

/kind bug
/kind cleanup

Lazily init events client  (like #2287)
Optimize events search perfomance (like #2423)

Fixes #

```
```

```docs

```


benjaminhuo:
@huanggze pls take a look at @junotx 's PR. Basically it looks similar but not exactly the same with yours.
It's ok for now because we're going to refine these similar logic for logging, auditing and events in the future.

huanggze:
I think junotx's PR looks good. We definitely need refactor logging, auditing and events in the next release.

/lgtm

PR #2517, opened 2020-07-21
add total num for runs

shaowenchen:

**What type of PR is this?**
/kind bug

add the total number for pipeline runs


pkg/simple/client/devops/pipeline.go
```diff
@@ -81,6 +81,7 @@ type Pipeline struct {
 // GetPipeBranchRun & SearchPipelineRuns
 type PipelineRunList struct {
 	Items []PipelineRun `json:"items"`
+	Total int           `json:"total_count"`
```
zryfish, line 84: We are switching to camelCase variable name instead of snake case, I think `total`  or `totalItems`(which is same with other API) is better here.

shaowenchen: fixed 
/cc @harrisonliu5

PR #2599, opened 2020-07-24
add new parameters allowedIngressNamespaces for user customization

### LINKED ISSUE #2550: should we allow ingress traffic from the kubesphere-monitoring-system default
By default we only allow traffic from the same workspace or project, but if user enable custom monitoring, then p8s will not be able to access the monitored data, should we allow it by default or let user add it by themselves?

I think not everyone will use custom monitoring, should let users add their own rules

duanjiong:

**What type of PR is this?**

/kind bug

add new parameters allowedIngressNamespaces  for user customization

Fixes #2550

### DIFF
--- a/pkg/controller/network/nsnetworkpolicy/nsnetworkpolicy_controller.go
+++ b/pkg/controller/network/nsnetworkpolicy/nsnetworkpolicy_controller.go
@@ -29,6 +29,7 @@ import (
 	"kubesphere.io/kubesphere/pkg/constants"
 	"kubesphere.io/kubesphere/pkg/controller/network"
 	"kubesphere.io/kubesphere/pkg/controller/network/provider"
+	options "kubesphere.io/kubesphere/pkg/simple/client/network"
 )
 
 const (
@@ -77,6 +78,7 @@ type NSNetworkPolicyController struct {
 	namespaceInformerSynced cache.InformerSynced
 
 	provider provider.NsNetworkPolicyProvider
+	options  options.NSNPOptions
 
 	nsQueue   workqueue.RateLimitingInterface
 	nsnpQueue workqueue.RateLimitingInterface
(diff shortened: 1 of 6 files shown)


pkg/simple/client/network/options.go
```diff
@@ -2,14 +2,22 @@ package network
 
 import "github.com/spf13/pflag"
 
+type NSNPOptions struct {
+	AllowedIngressNamespaces []string `json:"allowedIngressNamespaces,omitempty" yaml:"allowedIngressNamespaces,omitempty"`
```
zheng1, line 6: DefaultAllowedIngressNamespaces

pkg/controller/network/nsnetworkpolicy/nsnetworkpolicy_controller.go
```diff
@@ -301,7 +303,7 @@ func (c *NSNetworkPolicyController) generateNodeRule() (netv1.NetworkPolicyIngre
 	return rule, nil
 }
 
-func generateNSNP(workspace string, namespace string, matchWorkspace bool) *netv1.NetworkPolicy {
+func (c *NSNetworkPolicyController) generateNSNP(workspace string, namespace string, matchWorkspace bool) *netv1.NetworkPolicy {
```
zheng1, line 306: Please add some unit test for generateNSNP

duanjiong: This function has already been unit tested.

pkg/simple/client/network/options.go
duanjiong: This is not the default value, it can be set by the user, only if the user does not configure it then namespace ”kubesphere-moniting-system" is allowed by default.

zheng1: @zryfish Please take a look.

pkg/controller/network/nsnetworkpolicy/nsnetworkpolicy_controller.go
zheng1: OK, I see

pkg/simple/client/network/options.go
zryfish: I think we'd better set default values through configuration file, not in the code, cause we may need to change monitoring namespace name one day

duanjiong: > I think we'd better set default values through configuration file, not in the code, cause we may need to change monitoring namespace name one day

done

PR #2672, opened 2020-07-27
make ws admin can be aware of anything happened in its workspace.

### LINKED ISSUE #2648: missing audits in multi-tenant space
**Describe the Bug**
`ray` is the admin of the ws, `ws-self-provisioner1` is the self-provisioner of the ws. Use this user to create `demo-project1`. Now the default platform `admin` can see the audit of the project creation, but `ray` and `ws-self-provisioner1` can't. Please also check logging/events to see if having the same problem.
Screen Shot 2020-07-26 at 9 29 10 PM
Screen Shot 2020-07-26 at 9 29 40 PM
Screen Shot 2020-07-26 at 9 29 59 PM

**Versions Used**
KubeSphere: 3.0.0-dev
Kubernetes: 1.18.6

wanjunlei:

**What type of PR is this?**

/kind bug

Fixes #2648

### DIFF
--- a/pkg/apiserver/auditing/types.go
+++ b/pkg/apiserver/auditing/types.go
@@ -18,7 +18,8 @@ import (
 	"kubesphere.io/kubesphere/pkg/apiserver/request"
 	"kubesphere.io/kubesphere/pkg/client/listers/auditing/v1alpha1"
 	"kubesphere.io/kubesphere/pkg/informers"
-	resourcesv1alpha3 "kubesphere.io/kubesphere/pkg/models/resources/v1alpha3/resource"
+	"kubesphere.io/kubesphere/pkg/models/resources/v1alpha3"
+	"kubesphere.io/kubesphere/pkg/models/resources/v1alpha3/devops"
 	"kubesphere.io/kubesphere/pkg/utils/iputil"
 	"net"
 	"net/http"
@@ -41,18 +42,18 @@ type Auditing interface {
 }
 
 type auditing struct {
-	lister         v1alpha1.WebhookLister
-	resourceGetter *resourcesv1alpha3.ResourceGetter
-	cache          chan *auditv1alpha1.EventList
-	backend        *Backend
+	webhookLister v1alpha1.WebhookLister
+	devopsGetter  v1alpha3.Interface
+	cache         chan *auditv1alpha1.EventList
+	backend       *Backend
 }
 
 func NewAuditing(informers informers.InformerFactory, url string, stopCh <-chan struct{}) Auditing {
 
 	a := &auditing{
-		lister:         informers.KubeSphereSharedInformerFactory().Auditing().V1alpha1().Webhooks().Lister(),
-		resourceGetter: resourcesv1alpha3.NewResourceGetter(informers),
-		cache:          make(chan *auditv1alpha1.EventList, DefaultCacheCapacity),
+		webhookLister: informers.KubeSphereSharedInformerFactory().Auditing().V1alpha1().Webhooks().Lister(),
+		devopsGetter:  devops.New(informers.KubeSphereSharedInformerFactory()),
+		cache:         make(chan *auditv1alpha1.EventList, DefaultCacheCapacity),
 	}
 
 	a.backend = NewBackend(url, ChannelCapacity, a.cache, SendTimeout, stopCh)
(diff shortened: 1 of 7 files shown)


pkg/models/tenant/tenant.go
```diff
@@ -607,6 +607,40 @@ func (t *tenantOperator) listIntersectedNamespaces(workspaces, workspaceSubstrs,
 	return iNamespaces, nil
 }
 
+// listIntersectedWorkspaces returns a list of workspaces that MUST meet ALL the following filters:
+// 1. If `workspaces` is not empty, the workspace SHOULD be one of the specified workpsaces.
+// 2. Else f `workspaceSubstrs` is not empty, the workspace SHOULD be contains one of the specified substrings.
```
benjaminhuo, line 612: Else f => Else if

pkg/models/tenant/tenant.go
```diff
@@ -828,6 +871,28 @@ func (t *tenantOperator) Auditing(user user.Info, queryParam *auditingv1alpha1.Q
 			namespaceCreateTimeMap[ns.Name] = ns.CreationTimestamp.Time
 		}
 	}
+
```
benjaminhuo, line 874: Maybe we can add similar comments like the above ns section

pkg/models/tenant/tenant.go
```diff
@@ -607,6 +607,40 @@ func (t *tenantOperator) listIntersectedNamespaces(workspaces, workspaceSubstrs,
 	return iNamespaces, nil
 }
 
+// listIntersectedWorkspaces returns a list of workspaces that MUST meet ALL the following filters:
+// 1. If `workspaces` is not empty, the workspace SHOULD be one of the specified workpsaces.
+// 2. Else f `workspaceSubstrs` is not empty, the workspace SHOULD be contains one of the specified substrings.
+// 3. Else, return all workspace in the cluster.
+func (t *tenantOperator) listIntersectedWorkspaces(workspaces, workspaceSubstrs []string) ([]*tenantv1alpha1.Workspace, error) {
```
benjaminhuo, line 614: Do we have unit test code for this new function?

pkg/simple/client/auditing/elasticsearch/elasticsearch.go
```diff
@@ -322,6 +322,23 @@ func parseToQueryPart(f *auditing.Filter) interface{} {
...
+						"range": map[string]interface{}{
+							"RequestReceivedTimestamp": map[string]interface{}{
+								"gte": v,
+							},
+						},
+					}},
+				},
+			})
+		}
+
```
junotx, line 341: @wanjunlei the relationship between ws  and upper ns condition is or, which may cause problems. For example, an user is a ws admin who queries on an ns and may still see the entire ws.

pkg/models/tenant/tenant.go
wanjunlei: The logic of this function is very simple, list all workspace, compare with the elements in workspaces and workspaceSubstrs in turn, return results that meet the requirements

pkg/simple/client/auditing/elasticsearch/elasticsearch.go
```diff
@@ -360,6 +360,36 @@ func parseToQueryPart(f *auditing.Filter) interface{} {
 		return &bi
 	}
 
+	if len(f.ObjectRefNamespaces) > 0 {
+		if bi := shouldBoolbody("match_phrase_prefix", "ObjectRef.Namespace.keyword",
```
benjaminhuo, line 364: match_phrase_prefix should be change to match_phrase to be more accurate for both namespace and workspace query

pkg/apiserver/auditing/types.go
```diff
@@ -37,16 +41,18 @@ type Auditing interface {
...
 }
 
-func NewAuditing(lister v1alpha1.WebhookLister, url string, stopCh <-chan struct{}) Auditing {
+func NewAuditing(informers informers.InformerFactory, url string, stopCh <-chan struct{}) Auditing {
 
 	a := &auditing{
-		lister: lister,
-		cache:  make(chan *auditv1alpha1.EventList, DefaultCacheCapacity),
+		lister:         informers.KubeSphereSharedInformerFactory().Auditing().V1alpha1().Webhooks().Lister(),
+		resourceGetter: resourcesv1alpha3.NewResourceGetter(informers),
```
zryfish, line 54: It's not necessary to create a resourceGetter since just devops needed, we'd better create devopsGetter only

PR #2688, opened 2020-07-29
Allow global admins to view deleted namespace logs

huanggze:

**What type of PR is this?**
>
> /kind api-change

 /kind bug
 /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

- Allow global admins to view deleted namespace logs. (Don't append namespace creation time filter for global admin users) 
- Speed up query performance for global admins. (If global admin doesn't set any namespace filter, don't append namespace list filter to query body.)
- Remove query parameters `workspaces` and `workspace_query` due to never being used.

Fixes #

```
```

```docs

```


huanggze:
/assign @benjaminhuo 

/cc @junotx @wanjunlei 
As we discussed @wanjunlei , auditing and events have the same problem that is global admin users cannot view events, logs or audits from removed namespaces. It is because we always append namesace creation time filter to query body. For admin users, this requirement is unnecessary. Remvoing them may speed up query performance for admin users specially.

pkg/api/logging/v1alpha2/types.go
```diff
@@ -27,8 +27,6 @@ type APIResponse struct {
 
 type Query struct {
 	Operation       string
-	WorkspaceFilter string
```
benjaminhuo, line 30: workspacesearch filter and workspacesearch are both deleted?
log searching UI doesn't have workspace search condition?

wanjunlei: search log with workspace is a relatively common demand，it should be added in 3.1，this can be kept first

huanggze:
Workspace filter has never been used since 2,1.1. Namespace filter is enough to cover all use cases.

I think we can remove workspace filter until we see increasing demand.

image

pkg/api/logging/v1alpha2/types.go
huanggze: Workspace filter has never been used since 2,1.1. Namespace filter is enough to cover all use cases.

I think we can remove workspace filter until we see increasing demand. We have too many query parameters in logging APIs, this may confuse developers.

image

benjaminhuo: ok, agree

PR #2707, opened 2020-07-30
chore: docker build cache

fafucoder:
**What type of PR is this?**
/kind document

At present, the basic principle of docker using cache is that if the parent node has a cache, the current dockerfile statement has been built before and can use cache. So, to speed up compilation, some basic operations should be advanced.

Fixes #

```
```

```docs

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,8 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+DOCKER_TAG=latest
+
 OUTPUT_DIR=bin
 GOFLAGS=-mod=vendor
 define ALL_HELP_INFO
@@ -82,7 +84,10 @@ openapi:
 	go run ./tools/cmd/crd-doc-gen/main.go
 # Build the docker image
 docker-build: all
-	hack/docker_build.sh
+	hack/docker_build.sh ${DOCKER_TAG}
+
+docker-release: all
+	hack/docker_build.sh ${DOCKER_TAG} true
 
 # Run tests
 test: fmt vet
--- a/build/hypersphere/Dockerfile
+++ b/build/hypersphere/Dockerfile
@@ -2,6 +2,9 @@
 # Use of this source code is governed by an Apache license
 # that can be found in the LICENSE file.
 
+FROM alpine:3.9
+RUN apk add --update ca-certificates && update-ca-certificates
+
 FROM golang:1.12 as hypersphere-builder
 
 COPY / /go/src/kubesphere.io/kubesphere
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -12,6 +12,8 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+DOCKER_TAG=latest
```
zryfish, line 15: we are tagging image based on which branch we are on

build/hypersphere/Dockerfile
```diff
@@ -2,14 +2,15 @@
 # Use of this source code is governed by an Apache license
 # that can be found in the LICENSE file.
 
+FROM alpine:3.9
```
zryfish, line 5: hypersphere is deprecated, please just delete it

hack/docker_build.sh
```diff
@@ -14,11 +14,14 @@ tag_for_branch() {
 # push to kubespheredev with default latest tag
 REPO=${REPO:-kubespheredev}
 TAG=$(tag_for_branch $1)
+RELEASE=${2:-false}
```
zryfish, line 17: I don't see why `release` is necessary here, what you really need to change is `tag_for_branch` function, maybe return a default tag when no branch specified

zryfish:
Please sign your commit with `git commit -s `, also need to add more details about this PR, why we need it

Makefile
fafucoder: You are tag_for_branch() function need tag pararms, but you did't pass it

hack/docker_build.sh
fafucoder: make build only build image 
make release pull image to repo

use release params to distinguish them

build/ks-apiserver/Dockerfile
```diff
@@ -7,12 +7,13 @@ ENV UID=1002
...
-
 RUN apk add --no-cache ca-certificates && \
     addgroup --gid "$GID" "$USER" && \
-    adduser -D -g "" --ingroup "$USER" -u "$UID" "$USER" && \
-    chown -R kubesphere:"$GID" /usr/local/bin/ks-apiserver
+    adduser -D -g "" --ingroup "$USER" -u "$UID" "$USER"
+
+COPY  /bin/cmd/ks-apiserver /usr/local/bin/
+
+RUN chown -R kubesphere:"$GID" /usr/local/bin/ks-apiserver
```
LinuxSuRen, line 16: An extra command will increase the layer of the docker image. I'd prefer to keep them together.

build/ks-controller-manager/Dockerfile
```diff
@@ -3,10 +3,10 @@
 # that can be found in the LICENSE file.
 FROM alpine:3.11
 
-COPY  /bin/cmd/controller-manager /usr/local/bin/
-
 RUN apk add --no-cache ca-certificates
 
+COPY  /bin/cmd/controller-manager /usr/local/bin/
```
LinuxSuRen, line 8: Are there any reasons why we should change it? It'd good if you can explain it. Thanks.

LinuxSuRen (changes requested):
Hi @fanyanming2016 , Thanks for your contribution. I left some comments, please take a look at it.

PR #2720, opened 2020-07-31
add keyword filter for credential

shaowenchen:

**What type of PR is this?**

> /kind api-change

add  keyword filter to credential API, e.g conditions=keyword%3Dabc


pkg/kapis/devops/v1alpha3/handler.go
```diff
@@ -291,9 +291,19 @@ func (h *devopsHandler) GetCredential(request *restful.Request, response *restfu
 
 func (h *devopsHandler) ListCredential(request *restful.Request, response *restful.Response) {
 	devops := request.PathParameter("devops")
+
+	// parase filter parameter: keyword
+	conditions, err := params.ParseConditions(request)
+
+	if err != nil {
+		klog.Error(err)
```
zryfish, line 299: Should return the error message to the client
```suggestion
		api.HandleBadRequest(response, request, err)
		return
```

PR #2726, opened 2020-07-31
Improvement promopt msg

shaowenchen:
**What type of PR is this?**

/kind bug

fast failed when the error occurs.


pkg/kapis/devops/v1alpha3/handler.go
```diff
@@ -291,9 +291,19 @@ func (h *devopsHandler) GetCredential(request *restful.Request, response *restfu
 
 func (h *devopsHandler) ListCredential(request *restful.Request, response *restful.Response) {
 	devops := request.PathParameter("devops")
+
+	// parase filter parameter: keyword
+	conditions, err := params.ParseConditions(request)
+
+	if err != nil {
+		klog.Error(err)
+		return
```
zryfish, line 300: I think we should return the error to the client,

```suggestion
		api.HandleBadRequest(response, request, err)
		return
```

PR #2730, opened 2020-07-31
remove LastAppliedConfig annotation after password encrypted

wansir:

**What type of PR is this?**

 /kind bug

Remove `LastAppliedConfigAnnotation` annotation after password encrypted, make sure that plain text password is not stored anywhere.

Fixes #

```
```

```docs

```

### DIFF
--- a/pkg/controller/user/user_controller.go
+++ b/pkg/controller/user/user_controller.go
@@ -367,6 +367,8 @@ func (c *Controller) ensurePasswordIsEncrypted(user *iamv1alpha2.User) (*iamv1al
 		if user.Annotations == nil {
 			user.Annotations = make(map[string]string, 0)
 		}
+		// ensure plain text password won't be kept anywhere
+		delete(user.Annotations, corev1.LastAppliedConfigAnnotation)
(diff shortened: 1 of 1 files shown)


pkg/controller/user/user_controller.go
```diff
@@ -367,6 +367,8 @@ func (c *Controller) ensurePasswordIsEncrypted(user *iamv1alpha2.User) (*iamv1al
 		if user.Annotations == nil {
 			user.Annotations = make(map[string]string, 0)
 		}
+		// ensure plain text password won't be kept anywhere
+		delete(user.Annotations, corev1.LastAppliedConfigAnnotation)
 		user.Annotations[iamv1alpha2.PasswordEncryptedAnnotation] = "true"
```
rayzhou2017, line 372: Is the LastAppliedConfigAnnotation always existing? If no, the deleting will not crash?

wansir: > To delete a map entry, use the delete built-in function, whose arguments are the map and the key to be deleted. It's safe to do this even if the key is already absent from the map.

https://golang.org/doc/effective_go.html#maps

PR #2760, opened 2020-08-03
fix labelSelector not working

wansir:

**What type of PR is this?**

/kind bug

fix tenant resource API label selector not working.

Fixes https://github.com/kubesphere/console/issues/1053

```
```

```docs

```

### DIFF
--- a/pkg/models/tenant/tenant.go
+++ b/pkg/models/tenant/tenant.go
@@ -118,53 +118,49 @@ func (t *tenantOperator) ListWorkspaces(user user.Info, queryParam *query.Query)
 	}
 
 	decision, _, err := t.authorizer.Authorize(listWS)
-
 	if err != nil {
 		klog.Error(err)
 		return nil, err
 	}
 
+	// allowed to list all workspaces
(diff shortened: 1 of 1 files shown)


pkg/models/tenant/tenant.go
```diff
@@ -118,49 +118,41 @@ func (t *tenantOperator) ListWorkspaces(user user.Info, queryParam *query.Query)
...
 	}
 
 	workspaces := make([]runtime.Object, 0)
-
 	for _, roleBinding := range workspaceRoleBindings {
-
 		workspaceName := roleBinding.Labels[tenantv1alpha1.WorkspaceLabel]
-		workspace, err := t.resourceGetter.Get(tenantv1alpha2.ResourcePluralWorkspaceTemplate, "", workspaceName)
-
+		obj, err := t.resourceGetter.Get(tenantv1alpha2.ResourcePluralWorkspaceTemplate, "", workspaceName)
```
zryfish, line 144: ResourceGetter is big, should just use the specific resource you need, create that big object will consume much more resource

PR #2987, opened 2020-09-23
Add OpenPitrix API in swagger file

zheng1:
**What type of PR is this?**
/kind documentation

Fixes #

```
```

```docs

```

### DIFF
--- a/pkg/kapis/openpitrix/v1/register.go
+++ b/pkg/kapis/openpitrix/v1/register.go
@@ -36,9 +36,6 @@ const (
 var GroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1"}
 
 func AddToContainer(c *restful.Container, factory informers.InformerFactory, op op.Client) error {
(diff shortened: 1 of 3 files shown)


pkg/kapis/openpitrix/v1/register.go
```diff
@@ -36,17 +36,14 @@ const (
 var GroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1"}
 
 func AddToContainer(c *restful.Container, factory informers.InformerFactory, op op.Client) error {
-	if op == nil {
```
zryfish, line 39: why do we remove this

zheng1: In tools/cmd/doc-gen/main.go:118
```go
	urlruntime.Must(openpitrixv1.AddToContainer(container, informerFactory, nil))
```
After deletion, this register will be consistent with other register behavior.

PR #3002, opened 2020-10-06
Add node metric: node_pleg_quantile

yunkunrao:
**What type of PR is this?**
/kind optimization

This PR added a new  node metric node_pleg_quantile to monitor pleg module health status.

Take into account that "PLEG is not healthy" is a known bug and no real fix has been created to normalize nodes behavior, according to k8s community's feedback and discussion, we can know that this issue is a little complicated and involves many k8s components and aspects from network to container runtime.

But thanks for the prometheus monitor system, it's possible to gather pleg related metrics  to help us discover potential problems in advance and give us more clue when this issue reproduced. This PR add node_pleg_quantile which tracks the duration (in seconds) it takes for relisting pods in the Kubelet's PLEG and caculate corresponding histogram quantile. 

```
```

Generally pleg module will relist every second, so this metric value at least should be less than 1 for the most of time.

### DIFF
--- a/pkg/models/monitoring/named_metrics.go
+++ b/pkg/models/monitoring/named_metrics.go
@@ -112,6 +112,7 @@ var NodeMetrics = []string{
 	"node_load5",
 	"node_load15",
 	"node_pod_abnormal_ratio",
+	"node_pleg_quantile",
 }
 
 var WorkspaceMetrics = []string{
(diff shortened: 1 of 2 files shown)


benjaminhuo (approved):
/lgtm
Maybe we need to open an issue to frontend to add this to the node monitoring tab @Carlosfengv @leoendless

PR #3033, opened 2020-10-26
implement ippool

duanjiong:
1. support vlan ippool management
2. support calico ippool management

**What type of PR is this?**
/kind feature

This pr implements ip address management, which consists of two parts, one for interfacing with kubevirt for macvlan address allocation and the other for interfacing with calico for calico ippool address pool management.

```
```

```docs

```


config/crds/network.kubesphere.io_ippools.yaml
```diff
@@ -0,0 +1,121 @@
...
+          properties:
+            blockSize:
+              description: The block size to use for IP address assignments from this
+                pool. Defaults to 26 for IPv4 and 112 for IPv6.
+              type: integer
+            cidr:
+              description: The pool CIDR.
+              type: string
+            disabled:
+              description: When disabled is true, Calico IPAM will not assign addresses
```
zheng1, line 39: We need to change all `Calico IPAM` to ower selves.

duanjiong: Done. I looked,  that's the only thing I haven't changed.

duanjiong:
@zryfish  Now the VM controller relies on ippool.

pkg/apis/network/calicov3/block_affinity_types.go
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2019-2020 Tigera, Inc. All rights reserved.
```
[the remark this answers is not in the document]
duanjiong: As I told you before, calico doesn't provide client-go code, so if you import it directly, there are not only dependency problems, but also problems caused by log incompatibility.

PR #3048, opened 2020-10-30
remove istio-pilot config

zackzhangkai:
Signed-off-by: zackzhangkai <zackzhang@yunify.com>

**What type of PR is this?**

/kind enhancement

this API indeed is no need, it's only be used by kiali to get the istio version, but we didn't use this API  actually, so remove it to simplify our configuration

### DIFF
--- a/pkg/simple/client/servicemesh/options.go
+++ b/pkg/simple/client/servicemesh/options.go
@@ -20,9 +20,6 @@ import "github.com/spf13/pflag"
 
 type Options struct {
 
-	// istio pilot discovery service url
-	IstioPilotHost string `json:"istioPilotHost,omitempty" yaml:"istioPilotHost"`
-
 	// jaeger query service url
 	JaegerQueryHost string `json:"jaegerQueryHost,omitempty" yaml:"jaegerQueryHost"`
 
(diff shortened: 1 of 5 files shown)


zryfish:
If istio installed in a namespace other than istio-system, can Kiali auto-detect the endpoint of Istio?

zackzhangkai:
> If istio installed in a namespace other than istio-system, can Kiali auto-detect the endpoint of Istio?

@zryfish  Yes, I confirmed it, it is ok. At the same time ,  Prometheus configuration needed to be modified to adjust to the new namespace . When the prometheus can scrape the data, kiali will be ok to query from it.

pkg/simple/client/servicemesh/options.go
```diff
@@ -20,9 +20,6 @@ import "github.com/spf13/pflag"
 
 type Options struct {
 
-	// istio pilot discovery service url
```
zryfish, line 23: We really shouldn't remove this field this quick. I think it's better to deprecate it first, then remove it in the following releases. In this way, we could run KubeSphere with the old version configuration file, which will not cause inconsistency.

zackzhangkai: ok

zryfish: any update?

zackzhangkai:
To do in next version

PR #3098, opened 2020-11-10
fix: cascade delete resources

### LINKED ISSUE #2912: Unbundling cluster and multi-cluster enterprise space, not deleting related resources
**Describe the Bug**
When removing the cluster's authorization of enterprise space, prompted that this action will delete the resources of this enterprise space under the cluster
The test results:
Resources under a single cluster project can be deleted when the cluster and enterprise space are unbound
but, for a multi-cluster enterprise space, the resources under multi-cluster projects of that enterprise space under the cluster are not deleted after the authorization of the cluster is removed

移除集群

/kind bug
/area iam
/assign @wansir 
/milestone 3.1.0
/priority medium

wansir:

**What type of PR is this?**
/kind bug

Ensure that the namespace with workspace label is controlled by the existing workspace.

Fixes #2912

### DIFF
--- a/pkg/controller/namespace/namespace_controller.go
+++ b/pkg/controller/namespace/namespace_controller.go
@@ -33,6 +33,7 @@ import (
 	iamv1alpha2 "kubesphere.io/kubesphere/pkg/apis/iam/v1alpha2"
 	tenantv1alpha1 "kubesphere.io/kubesphere/pkg/apis/tenant/v1alpha1"
 	"kubesphere.io/kubesphere/pkg/constants"
+	"kubesphere.io/kubesphere/pkg/utils/k8sutil"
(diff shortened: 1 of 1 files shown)


pkg/controller/namespace/namespace_controller.go
```diff
@@ -205,6 +189,17 @@ func (r *ReconcileNamespace) bindWorkspace(namespace *corev1.Namespace) error {
 	return nil
 }
 
+// Reset the ownerReference of the namespace
```
zryfish, line 192: So this function actually removes Workspace from its ownerReference? I think we should choose another function name, reset usually means clear all.

zryfish: /cc @wansir

wansir: done

PR #3126, opened 2020-11-17
feature: add sync state for devops

shaowenchen:

**What type of PR is this?**
 /kind feature

add sync status to annotation for pipeline

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,4 +30,4 @@ coverage.txt
 kustomize/network/etcd
 apiserver.local.config
 tmp/
-
+kubesphere.yaml
--- a/pkg/models/devops/devops.go
+++ b/pkg/models/devops/devops.go
@@ -173,6 +173,8 @@ func (d devopsOperator) CreateDevOpsProject(workspace string, project *v1alpha3.
 	}
 	project.Name = ""
 	project.Labels[tenantv1alpha1.WorkspaceLabel] = workspace
+	project.Annotations[devopsv1alpha3.DevOpeProjectSyncStatusAnnoKey] = StatusPending
+	project.Annotations[devopsv1alpha3.DevOpeProjectSyncTimeAnnoKey] = GetSyncNowTime()
 	return d.ksclient.DevopsV1alpha3().DevOpsProjects().Create(project)
 }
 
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -30,4 +30,4 @@ coverage.txt
 kustomize/network/etcd
 apiserver.local.config
 tmp/
-
+kubesphere.yaml
```
LinuxSuRen, line 33: Where this file comes from? Do any programs generate it?

pkg/models/devops/devops.go
```diff
@@ -173,6 +173,8 @@ func (d devopsOperator) CreateDevOpsProject(workspace string, project *v1alpha3.
 	}
 	project.Name = ""
 	project.Labels[tenantv1alpha1.WorkspaceLabel] = workspace
+	project.Annotations[devopsv1alpha3.DevOpeProjectSyncStatusAnnoKey] = StatusPending
+	project.Annotations[devopsv1alpha3.DevOpeProjectSyncTimeAnnoKey] = GetSyncNowTime()
```
LinuxSuRen, line 177: Please consider putting line 176 and line 177 into a function instead of writing them four times.

LinuxSuRen (commented):
I guest this feature cannot work alone without the support of the front-end. In order to make it easy to track down. I'd suggest you create an issue in [ks console](https://github.com/kubesphere/console/issues).

I left some comments which are not blocked.

pkg/models/devops/devops.go
shaowenchen: It is not a good idea to optimize in advance. There's only one line of code.

.gitignore
shaowenchen: It is a configure file for the developer.  Can we provide a template file? @yuswift

yuswift: Which folder do you suggest we put it in?

shaowenchen: add `kubesphere.yaml.tpl` file to this repo, so the developer can just rename it and configure vars to run the project.

```yaml
cat kubesphere.yaml.tpl 
kubernetes:
  kubeconfig: "/root/.kube/config"
  master: https://192.168.0.5:6443
  $qps: 1e+06
  burst: 1000000
authentication:
  authenticateRateLimiterMaxTries: 10
  authenticateRateLimiterDuration: 10m0s
  loginHistoryRetentionPeriod: 168h
  maximumClockSkew: 10s
  multipleLogin: False
  kubectlImage: kubesphere/kubectl:v1.0.0
  jwtSecret: "xxx"
...
```

https://github.com/kubesphere/community/blob/master/developer-guide/development/quickstart.md#understand-kubesphere-cores-configuration

yuswift: Good suggestion.

PR #3127, opened 2020-11-18
add rolebinding api

RolandMa1986:

**What type of PR is this?**
/kind feature

Add Group binding APIs with Workspace Role and Project Role.

/assign @wansir

### DIFF
--- a/pkg/kapis/iam/v1alpha2/handler.go
+++ b/pkg/kapis/iam/v1alpha2/handler.go
@@ -167,7 +167,7 @@ func (h *iamHandler) RetrieveMemberRoleTemplates(request *restful.Request, respo
 			api.HandleInternalError(response, request, err)
 			return
 		}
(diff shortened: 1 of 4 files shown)


pkg/kapis/iam/v1alpha2/handler.go
```diff
@@ -65,6 +65,12 @@ type GroupMember struct {
 	GroupName string `json:"groupName"`
 }
 
+type GroupRoleBinding struct {
+	GroupName string `json:"groupName"`
+	RoleName  string `json:"roleName"`
+	Namespace string `json:"namespace,omitempty"`
+}
```
wansir, line 72: Why not use rolebinding directly, the subject can be group kind.

RolandMa1986: Refined. Use rolebinding and workspacerolebinding directly.

PR #3129, opened 2020-11-18
Avoid docker login everytime even it is done before

LinuxSuRen:
**What type of PR is this?**
/kind cleanup

Make it becomes more convenience when we build docker image.

None.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,8 @@ openapi:
 # Build the docker image
 docker-build: all
 	hack/docker_build.sh
+docker-build-no-test: ks-apiserver controller-manager
+	hack/docker_build.sh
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -79,6 +79,8 @@ openapi:
 # Build the docker image
 docker-build: all
 	hack/docker_build.sh
+docker-build-no-test: ks-apiserver controller-manager
+	hack/docker_build.sh
```
shaowenchen, line 83: What cases do we need to skip unit tests?

LinuxSuRen: It works well when just need to make a docker image on a personal computer quickly. Of course, it's not a good practice for using it.

PR #3138, opened 2020-11-20
refactor controllers

RolandMa1986:

**What type of PR is this?**
/kind cleanup

Simplify controllers with a BaseController compose pattern to impove UT coverage.

/assign @zryfish


pkg/controller/group/group_controller.go
```diff
@@ -38,6 +36,7 @@ import (
 	kubesphere "kubesphere.io/kubesphere/pkg/client/clientset/versioned"
 	iamv1alpha2informers "kubesphere.io/kubesphere/pkg/client/informers/externalversions/iam/v1alpha2"
 	iamv1alpha1listers "kubesphere.io/kubesphere/pkg/client/listers/iam/v1alpha2"
+	"kubesphere.io/kubesphere/pkg/utils/controller"
```
zryfish, line 39: I think we'd better move baseController to pkg `controller`, It will be more clear. How about `pkg/controller/controllerutils/`

RolandMa1986: I'd like to move it to  `pkg/controller/base/`,  would you prefer it?

Issue #3143, opened 2020-11-24
Kiali Proposal: directly make use of apis of the kiali pod instead of wrapping the source code


zackzhangkai:
**What's it about?**
In KubeSphere source code , we make use of apis of the kiali by wraping the kiali source code. The code now is getting old over time. Now we want to directly to deploy the kiali pod in `istio-system` namespace instead.

Reference issues:

https://github.com/kubesphere/kubesphere/issues/3106

zackzhangkai:
We will add a configuration in the kubesphere-config. After this configuration detected, use pod APIs.

```
$ kubectl get cm kubesphere-config -n kubesphere-system -oyaml
...
data:
  kubesphere.yaml: |
    servicemesh:
      kialiQueryHost: http://kiali.istio-system:20001
```

After use this APIs of pod, the ks-console is needed to be stay with the same version. Otherwise, the graph will be seem not available.

So, maybe the console can use detect this configuration to be compatible with the backend of ks-apiserver.

@leoendless

zackzhangkai:
After use kiali pod, how to be compatible with different members? If the two versions , old and new, exist in members, the console in host cluster will show problem.

zackzhangkai:
There is a problem in muti-cluster scene. If the host cluster upgrade to new version, then is will use a new API called `kapis/servicemesh.kubesphere.io/v1alpha2/namespaces/$(namespace)/applications/$(app)/graph`, the problem occured when the member cluster use the old version. 

![image](https://user-images.githubusercontent.com/20178386/101316065-beaa1700-3896-11eb-9073-8c38d7b837d3.png)

We haven't find proper solution for this problem.

zackzhangkai:
Resolved with multi-cluster scene. If you have a member with old version, such as below KubeSphere 3.0. Be sure your application CRD is new. If you cannot get result with command `kubectl get app`, it means you should upgrade by execute command `kubectl apply -f https://raw.githubusercontent.com/kubesphere/ks-installer/master/roles/common/files/ks-crds/app.k8s.io_applications.yaml`. Otherwise, your graph with kiali will show problem.

PR #3146, opened 2020-11-25
using Kiali API directly

zackzhangkai:
**What type of PR is this?**

/kind api-change
/kind cleanup

Use kiali pod instead of wrapping kiali code to reduce the development work of KubeSphere Project.

https://github.com/kubesphere/kubesphere/issues/3143

/assign @zryfish

### DIFF
--- a/pkg/kapis/servicemesh/metrics/v1alpha2/handler.go
+++ b/pkg/kapis/servicemesh/metrics/v1alpha2/handler.go
@@ -29,9 +29,31 @@ import (
 // default jaeger query api endpoint address
 var JaegerQueryUrl = "http://jaeger-query.istio-system.svc:16686"
 
+/*
+Use Kiali API directly if config existed in configmap.
+Such as:
+kubectl -n kubesphere-system get cm kubesphere-config -oyaml
+...
+kialiQueryHost: http://kiali.istio-system:20001
+...
+
+Otherwise, use the API provided by kiali code.
+
+Announce: The API provided by kiali code will deprecated in the future.
+*/
+
+var KialiQueryUrl string
+
 // Get app metrics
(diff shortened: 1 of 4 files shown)


pkg/kapis/servicemesh/metrics/v1alpha2/handler.go
```diff
@@ -43,17 +86,104 @@ func getWorkloadMetrics(request *restful.Request, response *restful.Response) {
...
 
-	handlers.WorkloadMetrics(request, response)
+	if len(KialiQueryUrl) == 0 {
+		handlers.WorkloadMetrics(request, response)
+	} else {
+		url := fmt.Sprintf("%s/kiali/api/namespaces/%s/workloads/%s/metrics?%s", KialiQueryUrl, namespace, workload, request.Request.URL.RawQuery)
+		resp, err := http.Get(url)
+		klog.V(4).Infof("Proxy workload metrics request to %s", url)
+
+		if err != nil {
```
zryfish, line 96: Don't Repeat Yourself, there are lots of code are exactly the same.

zackzhangkai: fixed

pkg/utils/net/net.go
```diff
@@ -45,3 +49,33 @@ func GetRequestIP(req *http.Request) string {
 
 	return address
 }
+
+func GetData(request *restful.Request, response *restful.Response, url string) {
```
zryfish, line 53: no need to move function here, we probably not going to use it elsewhere, I think place it in `pkg/kapis/servicemesh/metrics/v1alpha2/handler.go` would be better

zackzhangkai: done

PR #3148, opened 2020-11-26
add group filter for user api

RolandMa1986:

**What type of PR is this?**
/kind api-change

Frontend need to filter user by user's group
/assign @wansir


pkg/models/resources/v1alpha3/user/users.go
```diff
@@ -107,6 +108,10 @@ func (d *usersGetter) filter(object runtime.Object, filter query.Filter) bool {
 	switch filter.Field {
 	case iamv1alpha2.FieldEmail:
 		return user.Spec.Email == string(filter.Value)
+	case iamv1alpha2.InGroup:
```
wansir, line 111: What happened if `user.spec.group` has duplicate content.

RolandMa1986: There's no duplication here, The group name is unique in the cluster.

PR #3149, opened 2020-11-26
Mix up the pipeline list between ks and Jenkins

### LINKED ISSUE #3081: Inappropriate tips when I open the pipeline list page
As I talked about in the PR title. I got some inappropriate tips that might lead the users to think it's a bug.

The potential reason could be that the Jenkins is not ready. It might occur in a few different phases of Jenkins Pod:

* Init
image

* PodInitializing
image

In order to let you exactly know the meaning of the last screenshot, I already translated it into a human-readable picture.

image

### LINKED ISSUE #3108: Cannot create pipeline due to duplicate name which is invisible
**Describe the Bug**
When I try to create a pipeline, but ks tells me it's a duplicated name. But I cannot see it from the page.

**Versions Used**
KubeSphere: 3.0

**Environment**
None.

**How To Reproduce**

**Expected behavior**
No errors happen when someone creates a pipeline.

/kind bug

LinuxSuRen:
**What type of PR is this?**
/kind bug

Technically we should take the Pipeline list from ks or mixup the instead of only taking it from Jenkins. Because we always manipulate Pipeline from ks.

Fixes #3108
Fixes #3081

### DIFF
--- a/pkg/kapis/devops/v1alpha2/devops.go
+++ b/pkg/kapis/devops/v1alpha2/devops.go
@@ -62,7 +62,6 @@ func (h *ProjectPipelineHandler) getPipelinesByRequest(req *restful.Request) (ap
 
 	// parse query from the request
 	query := req.QueryParameter("q")
-	req.Request.PostForm.Set("limit", "1000")
 	for _, val := range strings.Split(query, ";") {
 		if strings.HasPrefix(val, "pipeline:") {
 			namespace = strings.TrimLeft(val, "pipeline:")
@@ -81,7 +80,7 @@ func (h *ProjectPipelineHandler) getPipelinesByRequest(req *restful.Request) (ap
 	defer req.Request.Form.Set("limit", "10000") // assume the pipelines no more than 10k
 
 	return h.devopsOperator.ListPipelineObj(namespace, func(list []*v1alpha3.Pipeline, i int, j int) bool {
-		return strings.Compare(list[i].Name, list[j].Name) < 0
+		return strings.Compare(strings.ToUpper(list[i].Name), strings.ToUpper(list[j].Name)) < 0
 	}, limit, start)
 }
 
--- a/pkg/kapis/devops/v1alpha2/devops.go
+++ b/pkg/kapis/devops/v1alpha2/devops.go
@@ -24,12 +24,14 @@ import (
 	"k8s.io/apiserver/pkg/authentication/user"
 	log "k8s.io/klog"
 	"kubesphere.io/kubesphere/pkg/api"
+	"kubesphere.io/kubesphere/pkg/apis/devops/v1alpha3"
 	iamv1alpha2 "kubesphere.io/kubesphere/pkg/apis/iam/v1alpha2"
 	"kubesphere.io/kubesphere/pkg/apiserver/request"
 	"kubesphere.io/kubesphere/pkg/constants"
 	"kubesphere.io/kubesphere/pkg/models/devops"
 	clientDevOps "kubesphere.io/kubesphere/pkg/simple/client/devops"
 	"net/http"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/simple/client/devops/pipeline.go
```diff
@@ -79,7 +79,7 @@ type Pipeline struct {
 	NumberOfFolders                int           `json:"numberOfFolders,omitempty" description:"number of folders"`
 	NumberOfPipelines              int           `json:"numberOfPipelines,omitempty" description:"number of pipelines"`
 	PipelineFolderNames            []interface{} `json:"pipelineFolderNames,omitempty" description:"pipeline folder names"`
-	WeatherScore                   int           `json:"weatherScore,omitempty" description:"the score to description the result of pipeline activity"`
```
shaowenchen, line 82: This may result in 500 if the field does not exist. Has it been fully tested?

pkg/kapis/devops/v1alpha2/devops.go
```diff
@@ -41,15 +44,81 @@ func (h *ProjectPipelineHandler) GetPipeline(req *restful.Request, resp *restful
...
+
+	// make sure we have an appropriate value
+	if start, err = strconv.Atoi(req.QueryParameter("start")); err != nil {
+		start = 0
+	}
+	if limit, err = strconv.Atoi(req.QueryParameter("limit")); err != nil {
+		limit = 10
+	}
+
+	defer req.Request.Form.Set("limit", "10000") // assume the pipelines no more than 10k
```
shaowenchen, line 74: get the value of the limit from QueryParmeter or set it as 10.  What's the intention here?

pkg/kapis/devops/v1alpha2/devops.go
```diff
@@ -41,15 +44,81 @@ func (h *ProjectPipelineHandler) GetPipeline(req *restful.Request, resp *restful
...
+
+	return h.devopsOperator.ListPipelineObj(namespace, func(list []*v1alpha3.Pipeline, i int, j int) bool {
+		return strings.Compare(list[i].Name, list[j].Name) < 0
```
shaowenchen, line 77: Please convert list[i].Name to uppercase and compare it

LinuxSuRen: Are you suggest it should be case insensitive? But according to the naming, the name seems can only be lower-case.

pkg/kapis/devops/v1alpha2/devops.go
LinuxSuRen: There are two lists here. And they might have a different order. So this list must be larger than another one. Anyway, the front-end will get the correct data.

pkg/simple/client/devops/pipeline.go
LinuxSuRen: The front-end cannot see `WeatherScore` if its value is `0`.  Because it is ignored due to `omitempty` field.
An int field always has the default value `0`. So it always has a value. I have no idea which cases might lead to a 500 result.

shaowenchen: When it unmarshal JSON data that the field of WeatherScore does not exist, it will result in an error.

pkg/kapis/devops/v1alpha2/devops.go
shaowenchen: Before the versions of 3.0, the case was sensitive.

pkg/kapis/devops/v1alpha2/devops.go
shaowenchen: It makes me a little confused. Why does it need to set req.Request.Form and req.Request.PostForm at the same time?

pkg/kapis/devops/v1alpha2/devops.go
LinuxSuRen: I see.

pkg/kapis/devops/v1alpha2/devops.go
LinuxSuRen: [The previous one](https://github.com/kubesphere/kubesphere/pull/3149/files#diff-38024724898ee9b3e4ca7f645d4c19869b100c394ccf587ef1e1ea554f3a7a2aR58) should be moved. It's useless.

pkg/simple/client/devops/pipeline.go
LinuxSuRen: I cannot produce the error case which you mentioned. But I try to unmarshal a JSON data which does not contain the field WeatherScore, nothing happened. You can run the demo code below:

https://play.golang.org/p/MJFcRRISY4i

pkg/kapis/devops/v1alpha2/devops.go
LinuxSuRen: Fixed it https://github.com/kubesphere/kubesphere/pull/3149/commits/62c297ec53049f2e08e873ea1ee5aa3412b54d5e#diff-38024724898ee9b3e4ca7f645d4c19869b100c394ccf587ef1e1ea554f3a7a2aR76

pkg/kapis/devops/v1alpha2/devops.go
LinuxSuRen: fixed it https://github.com/kubesphere/kubesphere/pull/3149/commits/62c297ec53049f2e08e873ea1ee5aa3412b54d5e#diff-38024724898ee9b3e4ca7f645d4c19869b100c394ccf587ef1e1ea554f3a7a2aL58

LinuxSuRen:
Please squash and merge this PR when it's ready. We don't need to keep all the commits, I keep them because force pushing will mash the review process up.

PR #3161, opened 2020-11-30
enable multicluster sync for groups

RolandMa1986:

**What type of PR is this?**
/kind feature

Groups are needed to be synced to member clusters when the MultiCluster feature is enabled. So member clusters can be authorized with group permission.


pkg/controller/group/group_controller.go
```diff
@@ -192,3 +236,57 @@ func (c *Controller) deleteRoleBindings(group *iam1alpha2.Group) error {
 
 	return nil
 }
+
+func (c *Controller) multiClusterSync(group *iam1alpha2.Group) error {
+
+	obj, err := c.federatedGroupLister.Get(group.Name)
+	if obj == nil {
```
wansir, line 243: Internal server error should be returned first, create the federation object if the error is `not found`.

pkg/controller/group/group_controller.go
```diff
@@ -192,3 +236,57 @@ func (c *Controller) deleteRoleBindings(group *iam1alpha2.Group) error {
...
+	}
+	if err != nil {
+		klog.Error(err)
+		return err
+	}
+
+	if !reflect.DeepEqual(obj.Spec.Template.Labels, group.Labels) ||
+		!reflect.DeepEqual(obj.Spec.Template.Annotations, group.Annotations) {
+
+		obj.Spec.Template.Annotations = group.Annotations
```
wansir, line 254: The annotations field is [managed by controllers in member clusters](https://github.com/kubernetes-sigs/kubefed/blob/master/docs/userguide.md#local-value-retention). If the field must be synchronized in multiple clusters, we should use [overrides](https://github.com/kubernetes-sigs/kubefed/blob/master/docs/userguide.md#overrides)

pkg/controller/group/group_controller.go
RolandMa1986: Refined.

pkg/controller/group/group_controller.go
RolandMa1986: The annotations field is not necessary, removed it.

pkg/controller/group/group_controller.go
```diff
@@ -106,10 +125,27 @@ func (c *Controller) reconcile(key string) error {
...
+		if c.multiClusterEnabled {
+			// Ensure not controlled by Kubefed
+			if group.Labels == nil || group.Labels[constants.KubefedManagedLabel] != "false" {
+				if g == nil {
+					g = group.DeepCopy()
+				}
+				if g.Labels == nil {
+					g.Labels = make(map[string]string, 0)
+				}
+				g.Labels[constants.KubefedManagedLabel] = "false"
```
zryfish, line 143: So what if the group doesn't have a `KubefedManagedLabel`, we are going to add a label to it.  I think we'd better do nothing If the group has no label `KubefedManagedLabel`. The L135 
```
    if group.Labels != nil && group.Labels[constants.KubefedManagedLabel] == "true" {
        group.Labels[constants.KubefedManagedLabel] = "false"
    }
```

pkg/controller/groupbinding/groupbinding_controller.go
```diff
@@ -101,14 +119,32 @@ func (c *Controller) reconcile(key string) error {
...
 		if !sliceutil.HasString(groupBinding.Finalizers, finalizer) {
-			groupBinding.ObjectMeta.Finalizers = append(groupBinding.ObjectMeta.Finalizers, finalizer)
-			if groupBinding, err = c.ksClient.IamV1alpha2().GroupBindings().Update(groupBinding); err != nil {
+			g = groupBinding.DeepCopy()
+			g.ObjectMeta.Finalizers = append(g.ObjectMeta.Finalizers, finalizer)
+		}
+
+		if c.multiClusterEnabled {
+			// Ensure not controlled by Kubefed
+			if groupBinding.Labels == nil || groupBinding.Labels[constants.KubefedManagedLabel] != "false" {
```
zryfish, line 130: same here

pkg/controller/group/group_controller.go
RolandMa1986: Group will be labeled with KubefedManagedLabel by Kubefed when multi-cluster enabled.  We must prevent Kubefed change Group on the Host cluster by the label.

zryfish: yes, but what if the group without KubefedManagedLabel, why do we still add a KubefedManagedLabel=false to it?

RolandMa1986: This Label will be added only when multi-cluster enabled and on the Master cluster, so it must exist. Immediately or to be updated by Kubefed when FedeatedGroup created.

PR #3163, opened 2020-11-30
Improve the process of build docker images.

LinuxSuRen:
**What type of PR is this?**
/kind cleanup

 Improve the process of build docker images.

None.

None.

None.

### DIFF
--- a/hack/gobuild.sh
+++ b/hack/gobuild.sh
@@ -41,6 +41,12 @@ BUILD_GOARCH=${GOARCH:-$(go env GOARCH)}
 GOBINARY=${GOBINARY:-go}
 LDFLAGS=$(kube::version::ldflags)
 
+# we need to build a particular binary instead of default one in some cases
+KS_BUILD_GOOS=${KS_BUILD_GOOS:-""}
+if [[ "${KS_BUILD_GOOS}" != "" ]]; then
+  BUILD_GOOS=${KS_BUILD_GOOS}
+fi
+
 # forgoing -i (incremental build) because it will be deprecated by tool chain.
 time GOOS=${BUILD_GOOS} CGO_ENABLED=0 GOARCH=${BUILD_GOARCH} ${GOBINARY} build \
         -ldflags="${LDFLAGS}" \
--- a/hack/gobuild.sh
+++ b/hack/gobuild.sh
@@ -41,12 +41,6 @@ BUILD_GOARCH=${GOARCH:-$(go env GOARCH)}
 GOBINARY=${GOBINARY:-go}
 LDFLAGS=$(kube::version::ldflags)
 
-# we need to build a particular binary instead of default one in some cases
-KS_BUILD_GOOS=${KS_BUILD_GOOS:-""}
-if [[ "${KS_BUILD_GOOS}" != "" ]]; then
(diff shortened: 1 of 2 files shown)


hack/gobuild.sh
```diff
@@ -41,6 +41,12 @@ BUILD_GOARCH=${GOARCH:-$(go env GOARCH)}
 GOBINARY=${GOBINARY:-go}
 LDFLAGS=$(kube::version::ldflags)
 
+# we need to build a particular binary instead of default one in some cases
+KS_BUILD_GOOS=${KS_BUILD_GOOS:-""}
```
zryfish, line 45: Why do we create a new variable KS_BUILD_GOOS here, you can just use GOOS to do cross-build.

LinuxSuRen: `GOOS` comes from Golang itself. And it's a global environment variable. Here's my case, I don't want to run command `export GOOS=darwin` every time. Also, it might affect all go build commands if I put it into `~/bashrc`.
So I can build a Linux docker image without global effection if we invest a new var.

zryfish: no need to export, just run `GOOS=darwin hack/gobuild.sh`

LinuxSuRen: I know that. But if I don't export it then I need to add `GOOS=darwin` every tiime.

zryfish: Yeah, that's on purpose, so users can know which os they are building for. And also, we are trying to add multiple platform building like K8s, to have env KS_BUILD_OS is confusing

LinuxSuRen: In my view, they're not conflicted between `GOSO` and `KS_BUILD_GOOS`. For those who want to know which os they want to build, they can use `GOOS`. For those who want to be more convenient, they can use `KS_BUILD_GOOS`.

About adding env `KS_BUILD_OS` might make people feel confused. Would you please give me a few use cases?

zryfish: For instance, if `KS_BUILD_OS` set to `darwin`, run `GOOS=linux hack/gobuild.sh` will result in building darwin executables. And what if I want to build a test package for Linux temporary when `KS_BUILD_OS=darwin`, I have two choices, 1) `export KS_BUILD_OS=linux` 2) `KS_BUILD_OS=linux hack/gobuild.sh`,  what the difference between the old way?

LinuxSuRen: OK. I guess it's difficult to find the ultimate solution for now. I can discard these changes.

How about these changes for the output docker image with the digest.

zryfish: That's ok, we can keep that part

LinuxSuRen: It's done.

LinuxSuRen: It still confuses me a lot because I never need a ks binary with darwin OS. And put `GOOS=linux` ahead of the build command every time is pretty annoying. In case anyone has the same trouble with me. I offer [an alternative solution](https://github.com/LinuxSuRen/ks-scripts). Please feel free to use it even help me to improve it.

Step1: install it via `curl https://gitee.com/linuxsuren/ks-scripts/raw/master/setup.sh -L | sh`
Step2: build your image: `ks-build`

PR #3167, opened 2020-12-01
config data desensitization

wansir:

**What type of PR is this?**

/kind feature

API `/config.kubesphere.io/v1alpha2/configs` can be accessed anonymously,  the returned data needs to be desensitized.


RolandMa1986 (commented):
/lgtm

PR #3175, opened 2020-12-03
add kubeconfig check

zackzhangkai:
Signed-off-by: zackzhangkai <zackzhang@yunify.com>

**What type of PR is this?**

/kind failing-test

/area microservice

add kubeconfig check 

/assign @zryfish

### DIFF
--- a/cmd/ks-apiserver/app/server.go
+++ b/cmd/ks-apiserver/app/server.go
@@ -120,14 +120,16 @@ func initializeServicemeshConfig(s *options.ServerRunOptions) {
 	kconfig.Set(config)
 
 	// Set kiali config
-	kubeconfig, err := clientcmd.BuildConfigFromFlags("", s.KubernetesOptions.KubeConfig)
-	if err != nil {
-		fmt.Println(err)
+	if len(s.KubernetesOptions.KubeConfig) != 0 {
+		kubeconfig, err := clientcmd.BuildConfigFromFlags("", s.KubernetesOptions.KubeConfig)
+		if err != nil {
+			fmt.Println(err)
+		}
+		k8sClient, err := kubernetes.NewClientFromConfig(kubeconfig)
+		if err != nil {
+			fmt.Println(err)
+		}
+		prometheusClient, _ := prometheus.NewClient()
+		business.SetWithBackends(k8sClient, prometheusClient)
 	}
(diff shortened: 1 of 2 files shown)


cmd/ks-apiserver/app/server.go
```diff
@@ -120,14 +120,16 @@ func initializeServicemeshConfig(s *options.ServerRunOptions) {
...
-	if err != nil {
-		fmt.Println(err)
-	}
-	k8sClient, err := kubernetes.NewClientFromConfig(kubeconfig)
-	if err != nil {
-		fmt.Println(err)
+	if len(s.KubernetesOptions.KubeConfig) != 0 {
+		kubeconfig, err := clientcmd.BuildConfigFromFlags("", s.KubernetesOptions.KubeConfig)
+		if err != nil {
+			fmt.Println(err)
```
zryfish, line 126: If there is an error initialize kubeconfig, apiserver wouln't run normally, so exit.

zackzhangkai: Yes, but it will first meet an error of kiali with unreadable error.

zryfish: A little bit confused here, what does that mean? Can you give an example.

zackzhangkai: If the kubeconfig leaves blank, then the error is as follows:

```bash
github.com/kiali/kiali/kubernetes.NewClientFromConfig(0x0, 0x0, 0x0, 0x0)
	/home/roland/go/pkg/mod/github.com/kubesphere/kiali@v0.15.1-0.20201110082537-0c2b977257d4/kubernetes/client.go:174 +0xa3
kubesphere.io/kubesphere/cmd/ks-apiserver/app.initializeServicemeshConfig(0xc0007e26f0)
```

zryfish: So we don't want to display this raw error message back to the user, instead we can return formated message, for example
```
   fmt.Sprintf("create kubernetes client failed, please double check your kubeconfig file, %v", err)
```

zackzhangkai: We cannot throw errors whenever the kubeconfig leaves blank. Kubeconfig is only needed when we debug locallly.

zryfish: OK, I see

PR #3178, opened 2020-12-03
filter groupbinding by user field

RolandMa1986:

**What type of PR is this?**
/kind api-change
/kind feature

In Frontend, we need to query the user's Groupbindings by user name.  
/assign @wansir


pkg/models/resources/v1alpha3/interface.go
```diff
@@ -114,6 +115,9 @@ func DefaultObjectMetaFilter(item metav1.ObjectMeta, filter query.Filter) bool {
 		return false
 	// /namespaces?page=1&limit=10&name=default
 	case query.FieldName:
+		return strings.Contains(item.Name, string(filter.Value))
+		// /namespaces?page=1&limit=10&generatename=default
+	case query.FieldGenerateName:
```
zryfish, line 120: I believe we can squash these two cases into one for simplicity

RolandMa1986: Yes, I see. GenerateName is always a substring of the Name field, so it's not necessary. Dropped the change.

PR #3186, opened 2020-12-08
support for querying pods metrics from the cluster level

junotx:
**What type of PR is this?**

/kind api-change

support for querying pods monitoring metrics from the cluster level

/area monitoring
/cc @benjaminhuo


zryfish:
I believe this api can only be called by cluster administrators. @wansir Can the current authorization framework cover this API?

pkg/simple/client/monitoring/prometheus/promql.go
```diff
@@ -349,6 +349,12 @@ func makePodMetricExpr(tmpl string, o monitoring.QueryOptions) string {
 		}
 	}
 
+	// For monitoring pods in the whole cluster
+	// Get /pods
+	if o.NamespaceName == "" && o.NodeName == "" {
+		podSelector = fmt.Sprintf(`pod=~"%s"`, o.ResourceFilter)
```
benjaminhuo, line 355: Please add an entry in promql_test.go for unit test

benjaminhuo (changes requested):

benjaminhuo:
> I believe this api can only be called by cluster administrators. @wansir Can the current authorization framework cover this API?

should be similar to call API for node's pods

benjaminhuo (approved):

PR #3203, opened 2020-12-14
[BugFix] Fix cluster unjoin federation plane when using proxy connection

### LINKED ISSUE #3202: cluster controller should not join or unjoin the cluster when proxy agent disconnected from proxy server
**Describe the Bug**
Cluster controller produces errors when the proxy agent disconnected from the proxy server. In fact, it should not join or unjoin the cluster under that situation. It should judge the agent status in advance, if not ready, update the cluster status to unready and return.
**Versions Used**
KubeSphere:v3.0.0
**How To Reproduce**
Scale the proxy agent replicas to zero and the cluster controller will produce some error.
**Expected behavior**
Judge the agent status in advance, if not ready, update the cluster status to unready and return.

yuswift:

**What type of PR is this?**
/kind bug

Cluster controller produces errors when the proxy agent disconnected from the proxy server. In fact, it should not join or unjoin the cluster under that situation. It should judge the agent status in advance, if not ready, update the cluster status to unready and return.
Fixes #3202

### DIFF
--- a/pkg/controller/cluster/cluster_controller.go
+++ b/pkg/controller/cluster/cluster_controller.go
@@ -550,9 +550,26 @@ func (c *clusterController) syncCluster(key string) error {
 		}
 	}
 
-	// kubeconfig not ready, nothing to do
-	if len(cluster.Spec.Connection.KubeConfig) == 0 {
-		return nil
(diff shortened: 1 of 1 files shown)


yuswift:
> /lgtm

thanks!

PR #3208, opened 2020-12-16
feature: add kubeovn  api

fafucoder:
Signed-off-by: linruichao <linruichao@ruijie.com.cn>

**What type of PR is this?**
/kind feature

add kubeovn api as the console can request it , the front web page like this.
image

```docs
https://github.com/alauda/kube-ovn
```


duanjiong:
I don't think that's appropriate. This pr strongly couples kubeovn, a cni plugin, to kubesphere, and we should abstract its capabilities to an intermediate layer, so that subsequent adaptations to other plugins will be more elegant

duanjiong:
For example, if you integrate the subnet management function of kube-ovn here, the ippool crd is included in the kubesphere, and you only need to implement the corresponding ippool provider accordingly. For details, please refer to https://github.com/kubesphere/community/blob/master/sig-network/concepts-and-designs/ippool.md

fafucoder:
> I don't think that's appropriate. This pr strongly couples kubeovn, a cni plugin, to kubesphere, and we should abstract its capabilities to an intermediate layer, so that subsequent adaptations to other plugins will be more elegant

Abstract an intermediate layer is not easy, different cni have a different api definition. The kuveovn has provided subnet and ip api, I don't think it's necessary to abstract a middle layer.

zheng1:
We have updated the version of client-go, please fix the conflict in this PR

pkg/apiserver/config/config.go
```diff
@@ -237,6 +242,10 @@ func (conf *Config) stripEmptyOptions() {
 		conf.NetworkOptions = nil
 	}
 
+	if conf.KubeovnOptions != nil && !conf.KubeovnOptions.Enabled {
+		conf.NetworkOptions = nil
```
zheng1, line 246:
```suggestion
		conf.KubeovnOptions = nil
```

pkg/apiserver/apiserver.go
```diff
@@ -553,6 +553,24 @@ func (s *APIServer) waitForResourceSync(stopCh <-chan struct{}) error {
...
+			klog.Warningf("resource %s not exists in the cluster", gvr)
+		} else {
+			_, err = kubeovnInformerFactory.ForResource(gvr)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	kubeovnInformerFactory.Start(stopCh)
+	kubeovnInformerFactory.WaitForCacheSync(stopCh)
```
zheng1, line 573: Please check kubeovn is enable or not

pkg/models/resources/v1alpha3/resource/resource.go
```diff
@@ -18,7 +18,11 @@ package resource
 
 import (
 	"errors"
-	snapshotv1beta1 "github.com/kubernetes-csi/external-snapshotter/client/v3/apis/volumesnapshot/v1beta1"
```
wansir, line 21: Unnecessary changes.

PR #3212, opened 2020-12-18
add a es client for auditing, events, and logging

wanjunlei:

/kind cleanup

The auditing, events, and logging all use elasticsearch as a search engine, they have the similar code that create elasticsearch client, build query DSL, and search data. This pr add a elasticsearch client to do these things, the auditing, events, and logging are no need to deal with things related to the underlying search engine. It is very usefully for changing some things about elasticsearch, such as x-pack support.


benjaminhuo (changes requested):

benjaminhuo (approved):
/lgtm
The code changes look good, please continue to add unit test cases when you have time.

benjaminhuo:
@junotx please take a look at the event part

junotx (approved):

zryfish:
@wanjunlei Please resolve conflicts and push again

pkg/apiserver/apiserver.go
```diff
@@ -130,7 +130,7 @@ type APIServer struct {
 	OpenpitrixClient openpitrix.Client
 
 	//
-	LoggingClient logging.Interface
+	LoggingClient logging.Client
```
zryfish, line 133: Why changing to Client? Interface is useful for unit testing.

wanjunlei: The `client` is more accurate for describing the function of this interface, which is conducive to improving code readability.

benjaminhuo: logging.Client is still an interface, just the interface name is changed from Interface to Client

pkg/simple/client/logging/options.go
```diff
@@ -42,21 +42,20 @@ func (s *Options) ApplyTo(options *Options) {
...
 func (s *Options) Validate() []error {
-	errs := []error{}
-
+	errs := make([]error, 0)
 	return errs
 }
 
 func (s *Options) AddFlags(fs *pflag.FlagSet, c *Options) {
-	fs.StringVar(&s.Host, "elasticsearch-host", c.Host, ""+
+	fs.StringVar(&s.Host, "logging-elasticsearch-host", c.Host, ""+
```
zryfish, line 50: We should be more careful with configuration name change, it will break backward compatibility.

wanjunlei: https://github.com/kubesphere/kubesphere/blob/master/pkg/simple/client/events/elasticsearch/options.go#L51
image

It has a same configuration name in the `events` module, so I changed these configuration name.

PR #3257, opened 2021-01-07
Fix the issues which caused by upper-case

LinuxSuRen:
**What type of PR is this?**
/kind bug
/area iam

Got issues with invalid username. It cannot be upper-case.
Error msg:
```
Invalid user name. Username can only contain lowercase letters and numbers.
```

### DIFF
--- a/pkg/apiserver/authentication/identityprovider/aliyunidaas/idaas.go
+++ b/pkg/apiserver/authentication/identityprovider/aliyunidaas/idaas.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"github.com/mitchellh/mapstructure"
 	"io/ioutil"
+	"strings"
 
 	"golang.org/x/oauth2"
 	"kubesphere.io/kubesphere/pkg/apiserver/authentication/identityprovider"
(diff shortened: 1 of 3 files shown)


wansir:
Just commit to the release-3.0 branch, we have added the account confirmation process in 3.1.

image

LinuxSuRen:
> We have added the account confirmation process in 3.1.
> 

Yes, I saw it. But the back-end just returns the username with upper-case.

wansir:
> > We have added the account confirmation process in 3.1.
> 
> Yes, I saw it. But the back-end just returns the username with upper-case.

I think it should be converted by the front-end user. If the username contains special characters, it needs to be reconfirmed. We cannot ensure that the username format of all identity providers is consistent,  should return the original username (conversion is unnecessary).

LinuxSuRen:
> > > We have added the account confirmation process in 3.1.
> > 
> > 
> > Yes, I saw it. But the back-end just returns the username with upper-case.
> 
> I think it should be converted by the front-end user. If the username contains special characters, it needs to be reconfirmed. We cannot ensure that the username format of all identity providers is consistent, should return the original username (conversion is unnecessary).

Fair enough. I'm going to close it.

PR #3267, opened 2021-01-11
fix application reconcile

zackzhangkai:
Signed-off-by: zackzhang <zackzhang@yunify.com>

**What type of PR is this?**

1. When add component, it will not do reconcile if the component has no OwnerRef section. Use the `EnqueueRequestsFromMapFunc` instead of `EnqueueRequestForOwner`.

2. Application of kubernetes-sigs, is defined only by `app.kubernetes.io/name` `app.kubernetes.io/version` labels .  So the resources , such as ingress which has no `app` label, can be added to application.

3. Refactor utils of controller/virturalservices/utils to controller/utils. So the common codes can be used by other package, such as `application controller` and `destinationrule controller`.

/kind bug
/area microservice
/assign @zryfish


pkg/controller/virtualservice/util/util.go
```diff
@@ -36,7 +36,6 @@ const (
 var ApplicationLabels = [...]string{
 	ApplicationNameLabel,
 	ApplicationVersionLabel,
-	AppLabel,
```
zryfish, line 39: `app` label is required by istio, why do we remove it?

pkg/controller/application/application_controller.go
```diff
@@ -133,9 +138,11 @@ func (r *ReconcileApplication) Reconcile(request reconcile.Request) (reconcile.R
 	return reconcile.Result{}, nil
 }
 
-func isApp(o metav1.Object) bool {
-	if o.GetLabels() == nil || !util.IsApplicationComponent(o.GetLabels()) {
-		return false
+func isApp(obs ...metav1.Object) bool {
+	for _, o := range obs {
+		if o.GetLabels() != nil && util.IsApplicationComponent(o.GetLabels()) {
```
zryfish, line 143: So, if any one of obs has required labels, return true, not all of them?

zackzhangkai: Yes.

pkg/controller/virtualservice/util/util.go
zackzhangkai: Use another AppLabels , which only have the necessary labels for kubernetes-sigs/application. So the resources ,such as ingress which has no `app` labels can be treated as application.

pkg/controller/destinationrule/destinationrule_controller.go
```diff
@@ -19,6 +19,7 @@ package destinationrule
 import (
 	"context"
 	"fmt"
+	"kubesphere.io/kubesphere/pkg/models/servicemesh"
```
zryfish, line 22: why do we move util package into models? That would makes the controller depend on the models package.

zackzhangkai: There are some other controllers depend on the package `util`, such as application controller、destinationrule controller. The `util` package located under the folder of virtualservice controller . Its path is `pkg/controller/virtualservice/util/util.go`.  I think it's unresonable.

zackzhangkai: change `controller/virtualservice/util` to `controller/util/servicemesh`.

Issue #3278, opened 2021-01-18
DevOps/The pipeline list is wrong when more than 10 pipelines are created


wenxin-01:
**Describe the Bug**
When the 11th pipeline is created successfully, the last pipeline in the pipeline list disappears and an exception pipeline is generated
For UI issues please also add a screenshot that shows the issue.

**Versions Used**
KubeSphere: 

**Environment**
http://allinone:30880

**Preset conditions**
There is a DevOps project  'test', which consists of 10 pipelines

**How To Reproduce**
1. Go to 'DevOps project 'test''
2. Create a new pipeline 'test1'

**Expected behavior**
1、pipeline test1 create successful。
2、There are 11 pipelines in the pipeline list and they are in the correct state

**Actual behavior**
When the 11th pipeline is created successfully, the last pipeline in the pipeline list disappears and an exception pipeline is generated

/kind bug
/area devops
/assign @shaowenchen 
/milestone 3.1.0
/priority high

LinuxSuRen:
This issue was fixed by #3279

PR #3279, opened 2021-01-18
Fix the pipeline list with incorrect paging calculating

LinuxSuRen:
**What type of PR is this?**
/kind bug
/area devops

As mentioned from the title, the pipeline list paging calculating is incorrect.

But I found anonther problem, there're two pages even there're only 10 pipelines. It should only has one page instead of two.
@harrisonliu5


LinuxSuRen:
@shaowenchen

LinuxSuRen:
fix #3278

Issue #3284, opened 2021-01-18
The interface reported an error when editing the pipeline created via JenkinsFile.


wenxin-01:
**Describe the Bug**
The interface reported an error when editing the pipeline created via JenkinsFile using the graphical interface.

**Versions Used**
KubeSphere:

**Environment**
http://allinone:30880/

**How To Reproduce**
1. Create pipeline ’test‘ using the default Settings.
2. Go to the details page of pipeline ’test‘.
3. click edit JenkinsFile.
4. Enter the correct JenkinsFile information and save. for example：
pipeline {
    agent any
    stages {
        stage('Example') {
            steps {
                echo 'Hello tester'
            }
        }
    }
}

5、click 'edit JenkinsFile' or 'edit pipeline', and then click save.

**Expected behavior**
Edit pipeline successful

**Actual behavior**
The page prompts an error message “Operation cannot be fulfilled on pipelines.devops.kubesphere.io "pipe2": the object has been modified; please apply your changes to the latest version and try again”

/kind bug
/area devops
/assign @shaowenchen
/milestone 3.1.0
/priority high

wenxin-01:
fixed

PR #3285, opened 2021-01-18
Fix the operation cannot be fulfilled due to fully update pipeline

LinuxSuRen:
**What type of PR is this?**
/kind bug
/area devops

Fix the below errors which caused by updating pipeline.
```
Operation cannot be fulfilled on pipelines.devops.kubesphere.io "pipe2": the object has been modified; please apply your changes to the latest version and try again
```


LinuxSuRen:
fix #3284

go.mod
```diff
@@ -9,6 +9,7 @@ go 1.13
 require (
 	code.cloudfoundry.org/bytefmt v0.0.0-20190710193110-1eb035ffe2b6
 	github.com/PuerkitoBio/goquery v1.5.0
+	github.com/appscode/jsonpatch v1.0.1
```
zryfish, line 12: I would suggest using `github.com/evanphx/json-patch` which also used by kubernetes repo. For example https://github.com/kubernetes/kubernetes/blob/cea1d4e20b4a7886d8ff65f34c6d4f95efcb4742/cmd/kubeadm/app/util/patches/patches.go#L31.

zryfish: Also could avoid potential conflicting with kubernetes repo in the future

LinuxSuRen: I cannot find any methods that could create a JSON path after go through `github.com/evanphx/json-patch`. It looks like only providing the way to apply a patch.

zryfish: OK, I see what's going on here. You were trying to solve the issue `Operation cannot be fulfilled on pipelines.devops.kubesphere.io "pipe2": the object has been modified; please apply your changes to the latest version and try again` when updating an object. This happens when updating with resourceVersion in the body, if still want to update even the object has been modified by other clients, I think you'd better update without resourceVersion. Patching makes things more complicated.

LinuxSuRen: The `ResourceVersion` cannot be empty. Because it'll cause the error: `metadata.resourceVersion: Invalid value: 0x0: must be specified for an update`. Or do you have any good examples to achieve that?

zryfish: As @leoendless said, get resourceVersion before updating would help decrease these errors. I don't think patching is the best solution here.

LinuxSuRen: It can reduce the chance of this kind of errors happen. But if you believe that we should take this solution instead of the patch. I'd prefer to create a new PR. In my view, get the latest resourceVersion then update it as some kind of patch.

see also #3349

Issue #3283, opened 2021-01-18
The creator of pipeline cannot audit this pipeline.


wenxin-01:
**Describe the Bug**
There is account A, whose platform role is platform-admin, and workspace role is admin, which is the creator and runner of the pipeline, and It is not the reviewer of the pipeline,  it cannot audit the pipeline.

**Versions Used**
KubeSphere:

**Environment**
http://allinone:30880/

**Preset conditions**
There is account A, whose platform role is platform-admin, and workspace role is admin

**How To Reproduce**
1. Log in with account 'A'.
2. create pipe 'test', add auditor B to pipeline 'test'.
3. running pipe 'test'.
4. Use account 'A' to audit the pipeline 'test'.

**Expected behavior**
account 'A' can audit pipeline 'test'.

**Actual behavior**
account 'A' can not audit pipeline 'test'.

/kind bug
/area devops
/assign @shaowenchen
/milestone 3.1.0
/priority high

LinuxSuRen:
related front-end issue https://github.com/kubesphere/console/issues/1431

LinuxSuRen:
It was fixed in `nightly-20210322`.

/close

Issue #3282, opened 2021-01-18
The pipeline cannot be audited by the admin user of workspace.


wenxin-01:
**Describe the Bug**
There is account 'A' , whose platform role is platform-regular and workspace role is viewer and It is  an auditor of the pipeline, it cannot audit the pipeline.

**Versions Used**
KubeSphere:

**Environment**
http://allinone:30880/

**Preset conditions**
1、There is account 'A' , whose platform role is platform-regular and workspace role is admin and It is an auditor of the pipeline.
2、There is account 'B', it created pipeline 'test'. This pipeline needs to be audited during operation.

**How To Reproduce**
1. Log in with account B, and running pipeline 'test'.
2. Log in with account A and enter the activity details page of pipeline ’test‘.
3. Make pipeline 'test' pass the audit.

**Expected behavior**
The operation is successful and the pipeline continues.

**Actual behavior**
The operation is failed,  user 'A' can not audited  pipeline 'test'.

/kind bug
/area devops
/assign @shaowenchen
/milestone 3.1.0
/priority high

LinuxSuRen:
related front-end issue https://github.com/kubesphere/console/issues/1431

wenxin-01:
issue fixed

PR #3286, opened 2021-01-20
Fix the incorrect approvable check of Pipeline input

LinuxSuRen:
**What type of PR is this?**
/kind bug
/area devops

We should support the pre-build admin role and the user-defined admin role as well.

related issues #3283 and #3282

The front-end also needs to fix it. See also kubesphere/console#1431

### DIFF
--- a/pkg/kapis/devops/v1alpha2/devops.go
+++ b/pkg/kapis/devops/v1alpha2/devops.go
@@ -23,9 +23,10 @@ import (
 	"github.com/emicklei/go-restful"
 	"k8s.io/apiserver/pkg/authentication/user"
 	log "k8s.io/klog"
+	"k8s.io/klog/v2"
 	"kubesphere.io/kubesphere/pkg/api"
 	"kubesphere.io/kubesphere/pkg/apis/devops/v1alpha3"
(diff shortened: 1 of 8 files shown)


pkg/kapis/devops/v1alpha2/devops.go
```diff
@@ -277,11 +278,48 @@ func (h *ProjectPipelineHandler) GetPipelineRunNodes(req *restful.Request, resp
 	resp.WriteAsJson(res)
 }
 
-func (h *ProjectPipelineHandler) approvableCheck(nodes []clientDevOps.NodesDetail, req *restful.Request) {
-	currentUserName, roleName := h.getCurrentUser(req)
+func (h *ProjectPipelineHandler) approvableCheck(nodes []clientDevOps.NodesDetail, pipe pipelineParam) {
```
shaowenchen, line 281: It is needed to describe the approval process by the note.

LinuxSuRen: Of course.

zryfish:
@shaowenchen any more comments on this PR?

Issue #3308, opened 2021-01-26
can't create workloads using self-signed certificate harbor


shaowenchen:
## General remarks

**Describe the Bug**

A harbor deployed with a self-signed certificate cannot create an application

![image](https://user-images.githubusercontent.com/43693241/105845664-1bfc5200-6016-11eb-85dd-b3072d59a7c2.png)

**Versions Used**

KubeSphere: 3.0.0
Kubernetes: (If KubeSphere installer used, you can skip this)

/area devops
/project devops
@kubesphere/sig-devops

shaowenchen:
let me fix it.
/assign @shaowenchen

PR #3317, opened 2021-01-28
add parameter to allow ignore cert

shaowenchen:

**What type of PR is this?**

 /kind bug

Fixes https://github.com/kubesphere/kubesphere/issues/3308

add parameter `ignorecert=true` to query image blob.

e.g. /kapis/resources.kubesphere.io/v1alpha2/registry/blob?namespace=harbor&image=harbor.dev.oa.com%2Fnginx%2Funit%3Alatest&secret=harbor&ignorecert=true

### DIFF
--- a/pkg/kapis/resources/v1alpha2/register.go
+++ b/pkg/kapis/resources/v1alpha2/register.go
@@ -17,8 +17,10 @@ limitations under the License.
 package v1alpha2
 
 import (
+	"net/http"
+
 	"github.com/emicklei/go-restful"
-	"github.com/emicklei/go-restful-openapi"
+	restfulspec "github.com/emicklei/go-restful-openapi"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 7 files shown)


pkg/kapis/resources/v1alpha2/register.go
```diff
@@ -145,6 +145,9 @@ func AddToContainer(c *restful.Container, k8sClient kubernetes.Interface, factor
 		Param(webservice.QueryParameter("secret", "secret name").
 			Required(false).
 			DataFormat("secret=%s")).
+		Param(webservice.QueryParameter("ignorecert", "whether verify cert if using https repo").
```
zryfish, line 148: How about using `insecure` as parameter name which used by [curl](https://curl.se/docs/manpage.html#-k ), or `skipTlsVerify` which used by `kubernetes` 
```
--insecure-skip-tls-verify
--
  | If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure
```

zryfish: The rest of the PR is good.

shaowenchen:
fixed

pkg/kapis/resources/v1alpha2/register.go
shaowenchen: fixed

PR #3319, opened 2021-01-28
remove http1.1 support code

yuswift:

**What type of PR is this?**
/kind cleanup

We've updated the dependencies and don't need to copy the `supportsHTTP11()` function  from `https://github.com/kubernetes/apimachinery/blob/master/pkg/util/proxy/dial.go` manaully.

### DIFF
--- a/pkg/apiserver/dispatch/dispatch.go
+++ b/pkg/apiserver/dispatch/dispatch.go
@@ -18,24 +18,25 @@ package dispatch
 
 import (
 	"fmt"
+	"net/http"
+	"net/url"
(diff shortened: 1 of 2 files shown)


zryfish:
Can we also remove same code in https://github.com/kubesphere/kubesphere/blob/master/pkg/apiserver/filters/kubeapiserver.go#L76

yuswift:
> Can we also remove same code in https://github.com/kubesphere/kubesphere/blob/master/pkg/apiserver/filters/kubeapiserver.go#L76

Done.

PR #3345, opened 2021-02-03
add heartbeat package for the websocket connection of the console

### LINKED ISSUE #3293: kubectl terminal session disconnect rapidly
**Describe the Bug**

The container terminal connection uses the WebSocket protocol. WebSockets will be affected by `proxy read timeout` if a reverse proxy is used. We need to increase the timeout or implement a ping message to keep the connection alive. The ping solution has the added benefit of discovering if the connection was closed unexpectedly.

image

**Expected behavior**

Keep the WebSocket session alive.

/assign @wansir @leoendless 
/milestone v3.1.0

yuswift:

**What type of PR is this?**
/kind bug

By adding heartbeat package for websocket connection, we can keep the connection alive so that it will not be killed by the proxy server(e.g. ingress-nginx).
Fixes #3293

### DIFF
--- a/pkg/models/terminal/terminal.go
+++ b/pkg/models/terminal/terminal.go
@@ -26,6 +26,7 @@ import (
 	"github.com/gorilla/websocket"
 	"io"
 	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
@@ -36,7 +37,8 @@ import (
 
 const (
 	// Time allowed to write a message to the peer.
-	writeWait = 10 * time.Second
+	writeWait         = 10 * time.Second
+	heartBeatInterval = 15 * time.Second
 )
 
 // PtyHandler is what remotecommand expects from a pty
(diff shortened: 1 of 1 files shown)


pkg/models/terminal/terminal.go
```diff
@@ -200,8 +219,12 @@ func (t *terminaler) HandleSession(shell, namespace, podName, containerName stri
 	var err error
 	validShells := []string{"sh", "bash"}
 
+	stopChan := make(chan struct{})
+
 	session := &TerminalSession{conn: conn, sizeChan: make(chan remotecommand.TerminalSize)}
 
+	go wait.Until(session.sendHeartBeatPackage, time.Duration(3)*time.Second, stopChan)
```
wansir, line 226: The default interval should be constant, I think the interval should be increased appropriately.

yuswift: Should we increase the interval such as 1s, 2s, 4s, 8s?

pkg/models/terminal/terminal.go
```diff
@@ -200,8 +222,12 @@ func (t *terminaler) HandleSession(shell, namespace, podName, containerName stri
 	var err error
 	validShells := []string{"sh", "bash"}
 
+	stopChan := make(chan struct{})
```
zryfish, line 225: Use stopCha created by apiserver https://github.com/kubesphere/kubesphere/blob/master/cmd/ks-apiserver/app/server.go#L57

pkg/models/terminal/terminal.go
```diff
@@ -36,7 +37,8 @@ import (
 
 const (
 	// Time allowed to write a message to the peer.
-	writeWait = 10 * time.Second
+	writeWait         = 10 * time.Second
+	heartBeatInterval = time.Duration(15) * time.Second
```
zryfish, line 41:
```suggestion
	heartBeatInterval = 15 * time.Second
```

go.mod
```diff
@@ -96,6 +96,7 @@ require (
 	k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6
 	k8s.io/kubectl v0.18.6
 	k8s.io/kubernetes v1.14.0
+	k8s.io/metrics v0.18.6
```
zryfish, line 99: Why there is a new module imported? I didn't see any code change depended on this module.

yuswift: Me neither. But after I run make `ks-apiserver`, the go.mod & go.sum files are changed. Should we remove the dependency?

pkg/models/terminal/terminal.go
yuswift: I think the lifecycle of the heartbeat channel should not be controlled by apiserver. If the user closes the terminal page, the server shouldn't send heartbeats anymore, so we close the channel to tell the goroutine to stop.

PR #3377, opened 2021-02-23
feature: add user email to built-in ldap

hchenc:
add user email to built-in ldap

What type of PR is this?
/kind feature

By adding user email to built-in ldap which can support some components(e.g gitlab,harbor) to adopt the built-in ldap to authenticating user when no external ldap component.

### DIFF
--- a/pkg/simple/client/ldap/ldap.go
+++ b/pkg/simple/client/ldap/ldap.go
@@ -254,6 +254,10 @@ func (l *ldapInterfaceImpl) Create(user *iamv1alpha2.User) error {
 				Type: ldapAttributeUserPassword,
 				Vals: []string{user.Spec.EncryptedPassword},
 			},
+			{
+				Type: ldapAttributeMail,
+				Vals: []string{user.Spec.Email},
+			},
 		},
 	}
 


wansir:
@hchenc  The email attribute is able to modify and may be empty. We should reduce the data coupling. We plan to replace the internal LDAP with the OIDC protocol. GitLab can also be integrated https://docs.gitlab.com/ee/administration/auth/oidc.html

hchenc:
> @hchenc The email attribute is able to modify and may be empty. We should reduce the data coupling. We plan to replace the internal LDAP with the OIDC protocol. GitLab can also be integrated https://docs.gitlab.com/ee/administration/auth/oidc.html
> 
> Related to #3378

For my team(5+) internal ldap is very useful so i can register user account once and login anywhere without modifying action again(we don't have independent ldap). Maybe there are many little dev teams like mine need the built-in ldap to support some component

wansir:
OIDC protocol is more suitable than LDAP. As I said, data needs to be synchronized after data changes. This will couple the user controller and OpenLDAP. OpenLDAP is a stateful service that requires additional maintenance. In our past experience, this will reduce the availability of the ks-controller-manager. Synchronizing data is an additional overhead. You can see that the current way of writing passwords directly to LDAP is not suitable either, this is not a good practice.

hchenc:
> OIDC protocol is more suitable than LDAP. As I said, data needs to be synchronized after data changes. This will couple the user controller and OpenLDAP. OpenLDAP is a stateful service that requires additional maintenance. In our past experience, this will reduce the availability of the ks-controller-manager. Synchronizing data is an additional overhead. You can see that the current way of writing passwords directly to LDAP is not suitable either, this is not a good practice.

totally agree with you

PR #3380, opened 2021-02-23
refactor cluster controller

### LINKED ISSUE #3234: The tower server shouldn't update the cluster.status
**What's it about?**
Currently, when cluster.spec.connection.type is proxy, the tower server updates the cluster.status filed, and the cluster controller determines if the cluster is ready according to the status.

Actually, the tower server shouldn't update the cluster.status filed.
The cluster controller can try to connect to the proxy service, determines if the cluster is ready by the response HTTP code.
This proposal can reduce the complexity between tower server and clsuter-controller.
/area multicluster
/kind feature-request

yuswift:

**What type of PR is this?**
/kind design
Reduce the complexity between tower server and clsuter-controller. Remove the `port allocation` `proxy creation` `token generation` steps, add `cluster ready detection`  step.
Fixes #3234

### DIFF
--- a/pkg/controller/cluster/cluster_controller.go
+++ b/pkg/controller/cluster/cluster_controller.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"math/rand"
 	"net/http"
 	"reflect"
 	"sync"
@@ -33,7 +32,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/util/intstr"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -44,6 +42,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 	fedv1b1 "sigs.k8s.io/kubefed/pkg/apis/core/v1beta1"
@@ -56,12 +55,12 @@ import (
 )
 
 // Cluster controller only runs under multicluster mode. Cluster controller is following below steps,
-//   1. Populates proxy spec if cluster connection type is proxy
-//     1.1 Wait for cluster agent is ready if connection type is proxy
+//   1. Wait for cluster agent is ready if connection type is proxy
 //   2. Join cluster into federation control plane if kubeconfig is ready.
 //   3. Pull cluster version and configz, set result to cluster status
 // Also put all clusters back into queue every 5 * time.Minute to sync cluster status, this is needed
 // in case there aren't any cluster changes made.
+// Also check if all of the clusters are ready by the spec.connection.kubeconfig every resync period
 
 const (
 	// maxRetries is the number of times a service will be retried before it is dropped out of the queue.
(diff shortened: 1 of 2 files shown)


pkg/controller/cluster/cluster_controller.go
```diff
@@ -348,6 +345,80 @@ func (c *clusterController) reconcileHostCluster() error {
...
+			klog.Error(err)
+			continue
+		}
+
+		config, err := clientConfig.ClientConfig()
+		if err != nil {
+			klog.Error(err)
+			continue
+		}
+		config.Timeout = 10 * time.Second
```
zryfish, line 370: What if there are lots of clusters, saying 50, and each cluster takes 9s to finish probing, that would be 450s, > resyncPeriod.

pkg/simple/client/multicluster/options.go
```diff
@@ -79,5 +79,5 @@ func (o *Options) AddFlags(fs *pflag.FlagSet, s *Options) {
 		"This field is used when generating deployment yaml for agent.")
 
 	fs.DurationVar(&o.ClusterControllerResyncSecond, "cluster-controller-resync-second", s.ClusterControllerResyncSecond,
-		"Cluster controller resync second to sync cluster resource.")
+		"Cluster controller resync second to sync cluster resource. e.g. 30s 60s 120s...")
```
zryfish, line 82: better to start `2m, 5m, 10m`, small resync period increases load

pkg/controller/cluster/cluster_controller.go
```diff
@@ -348,6 +345,80 @@ func (c *clusterController) reconcileHostCluster() error {
 	return err
 }
 
+func (c *clusterController) judgeIfClusterIsReady() error {
```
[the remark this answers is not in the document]
yuswift: Agreed.

pkg/controller/cluster/cluster_controller.go
yuswift: 10 seconds seem to be too long. But the case you are talking about is very rare. It's unlikely that every cluster connection takes 9s. In most cases, one connection takes several ms. How about change the timeout to 3s?

pkg/simple/client/multicluster/options.go
yuswift: Agree with that. I will update the comment.

pkg/controller/cluster/cluster_controller.go
zryfish: If there are network issues on the node where the `ks-controller-manager` pod residing, it's possible.

zryfish: What I did before put cluster back to working queue every resyncPeriod, and check its readiness on main sync loop.

yuswift: I didn't see ` put cluster back to working queue every resyncPeriod`, but I saw `check its readiness on main sync loop.`. I think we don't need to `put cluster back to working queue every resyncPeriod` manually, the cluster informer does that automatically. The reason I check the cluster readiness separately is to check all of the cluster readiness, not only proxy connection. What you did before only checks the proxy cluster if has `agent availably` status, then updates the cluster status to ready or not. By using kubeconfig, I think it's more reliable.

- [ ]

yuswift: If put in the main sync loop, once the cluster is created/updated/deleted, the check will be performed, which may be too frequent. On the other hand, the check may be too long and will affect the sync loop. what is your suggestion?

zryfish: That's true, but we need to update `cluster.status.configz` every resyncPeriod too. So I suggest make `config.timeout` shorter, and probe in the main loop.

yuswift: Currently, we update the `cluster.status.configz` every resyncPeriod at the end main loop. The update doesn't change.

zryfish: OK, better to make `config.timeout` shorter

yuswift: `config.Timeout` has been set as 3s by default. We can merge this pr now.

PR #3394, opened 2021-02-24
Update calico lib

### LINKED ISSUE #3387: After deleting a pod that uses Ippool, the Ippool can still not be successfully deleted
**Describe the Bug**
After deleting a pod that uses Ippool, the Ippool can still not be successfully deleted

**steps**
1. Create an Ippool
2. Assign an enterprise space to ippool
3. Go to the project below the enterprise space and create the workload using the ippool above
4. Delete the workload created above
5. Delete ippool created in step 1

**Expected results**
Ippool can be removed successfully

**Actual results**
The interface prompts to delete successfully, but actually does not delete

image

ks version：nightly-20210222

/kind bug
/assign @duanjiong
/priority High
/milestone 3.1
/area networking

duanjiong:

**What type of PR is this?**
 /kind bug

fix struct definition error in v3

Fixes #3387

```
```

```docs

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -55,7 +55,7 @@ require (
 	github.com/patrickmn/go-cache v2.1.0+incompatible // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/projectcalico/kube-controllers v3.8.8+incompatible
-	github.com/projectcalico/libcalico-go v1.7.2-0.20191104213956-8f81e1e344ce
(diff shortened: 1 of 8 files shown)


zryfish:
Did you run `hack/update-vendor.sh`, and please use `hack/pin-dependency.sh` instead of editing go.mod manually.

duanjiong:
> Did you run `hack/update-vendor.sh`, and please use `hack/pin-dependency.sh` instead of editing go.mod by hand.

It's done according to what you said

PR #3403, opened 2021-02-25
support selective application reconcilation

### LINKED ISSUE #3404: Selectively reconcile application objects
**What's it about?**
KubeSphere application managment is built on top of [sig-application](https://github.com/kubernetes-sigs/application), using object `application` from this repo. There are other projects depend on this repo too. If KubeSphere was installed alongside these projects, conflicts will happen. Because there will two dependent controllers trying to reconcile the same object.

In order to avoid conflicts, KubeSphere need to reconcile application selectively. 

/area app-management
/area microservice
/kind feature-request

zryfish:
Signed-off-by: Jeff <jeffzhang@yunify.com>

**What type of PR is this?**
/kind feature

Add `ApplicationSelector` to the application controller to reconcile applications only matched with the given selector, skip reconciliation if unmatched. Still reconcile all applications if this configuration left empty, which means doesn't introduce any breaking change.

Fixes #3404

```
```

```docs
gfs.StringVar(&s.ApplicationSelector, "application-selector", s.ApplicationSelector, ""+
		"Only reconcile application(sigs.k8s.io/application) objects match given selector, this could avoid conflicts with "+
		"other projects built on top of sig-application. Default behavior is to reconcile all of application objects.")
```


vendor/kubesphere.io/client-go
```diff
@@ -1 +0,0 @@
-../../staging/src/kubesphere.io/client-go
```
RolandMa1986, line 1: The In-repo modules shouldn't be added to the vendor directory. it could be an issue of *pin-dependency.sh*. I will try to fix the script. Please move it out from the commit.

pkg/controller/application/application_controller.go
```diff
@@ -18,45 +18,239 @@ package application
...
+	}); err != nil {
+		return fmt.Errorf("failed to update status of Application %s/%s: %v", nn.Namespace, nn.Name, err)
+	}
+	return nil
+}
+
+func (r *ApplicationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	c, err := ctrl.NewControllerManagedBy(mgr).
+		Named("application-controller").
+		For(&appv1beta1.Application{}).Build(r)
```
zackzhangkai, line 252: When the application object changed, will do the reconcile to modify the status of the application. But if the deployments (service or strategy and so on)  which belongs to the application changed, should trigger to modify the application. How to guarantee this situation?

zryfish: https://github.com/kubesphere/kubesphere/pull/3403/files#diff-8704122abcfce9d13573e5e86a42be45da623b4e5ff6a5f1ad7176bbf9a5d3a5L73-L97 
Didn't remove code from the last PR

zryfish:
@RolandMa1986 @zackzhangkai Any more comments?

vendor/kubesphere.io/client-go
zryfish: Fixed

pkg/controller/application/application_controller.go
zackzhangkai: Cannot locate the code by your link.

zryfish: Just about a few lines below here

```
for _, s := range sources {
		// Watch for changes to Application
		err = c.Watch(
			&source.Kind{Type: s},
			&handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(
				func(h handler.MapObject) []reconcile.Request {
					return []reconcile.Request{{NamespacedName: types.NamespacedName{
						Name:      servicemesh.GetApplictionName(h.Meta.GetLabels()),
						Namespace: h.Meta.GetNamespace()}}}
				})},
			predicate.Funcs{
				UpdateFunc: func(e event.UpdateEvent) bool {
					return isApp(e.MetaOld, e.MetaNew)
				},
				CreateFunc: func(e event.CreateEvent) bool {
					return isApp(e.Meta)
				},
				DeleteFunc: func(e event.DeleteEvent) bool {
					return isApp(e.Meta)
				},
			})
		if err != nil {
			return err
		}
	}
```

PR #3433, opened 2021-03-04
limit login record entries

### LINKED ISSUE #3191: Limit login record maximum entries
**What's it about?**
Currently, the login record retention time can be limited by `loginHistoryRetentionPeriod`. But it will still generate a lot of data (frequently request the login API), which will increase the pressure of etcd. To avoid resource abuse, it is necessary to limit the maximum record entries.

/kind feature-request
/milestone 3.1.0

wansir:

**What type of PR is this?**

/kind optimization

1. Add `loginHistoryMaximumEntries` authentication configuration.
2. Delete orphan objects if related user not found.
3. Limit login record entries, default to 48 per user.

Fixes #3191


pkg/apiserver/authentication/options/authenticate_options.go
```diff
@@ -58,6 +61,7 @@ func NewAuthenticateOptions() *AuthenticationOptions {
 		AuthenticateRateLimiterDuration: time.Minute * 30,
 		MaximumClockSkew:                10 * time.Second,
 		LoginHistoryRetentionPeriod:     time.Hour * 24 * 7,
+		LoginHistoryMaximumEntries:      48,
```
zryfish, line 64: Isn't 48 too small? Maybe we could increase it to, say 100?

pkg/apiserver/authentication/options/authenticate_options.go
```diff
@@ -68,7 +72,10 @@ func NewAuthenticateOptions() *AuthenticationOptions {
 func (options *AuthenticationOptions) Validate() []error {
 	var errs []error
 	if len(options.JwtSecret) == 0 {
-		errs = append(errs, fmt.Errorf("jwt secret is empty"))
+		errs = append(errs, errors.New("JWT secret should not be empty"))
```
zryfish, line 75:
```suggestion
		errs = append(errs, errors.New("JWT secret MUST not be empty"))
```
When talking about something that is forbidden, use `MUST not`, use `MUST` when being postitive.

pkg/controller/loginrecord/loginrecord_controller.go
```diff
@@ -136,28 +154,44 @@ func (c *loginRecordController) reconcile(key string) error {
...
+	for _, r := range oldEntries {
+		err = c.ksClient.IamV1alpha2().LoginRecords().Delete(context.TODO(), r.Name, metav1.DeleteOptions{})
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *loginRecordController) relatedUser(loginRecord *iamv1alpha2.LoginRecord) (*iamv1alpha2.User, error) {
```
zryfish, line 190: `relatedUser` not an obvious function name, how about `userForLoginRecord`

pkg/controller/loginrecord/loginrecord_controller.go
```diff
@@ -136,28 +154,44 @@ func (c *loginRecordController) reconcile(key string) error {
...
 		user.Status.LastLoginTime = &loginRecord.CreationTimestamp
-		user, err = c.ksClient.IamV1alpha2().Users().UpdateStatus(context.Background(), user, metav1.UpdateOptions{})
+		_, err := c.ksClient.IamV1alpha2().Users().UpdateStatus(context.Background(), user, metav1.UpdateOptions{})
 		return err
 	}
 	return nil
 }
+
+// limitEntriesFor will delete old entries out of limit
+func (c *loginRecordController) limitEntriesFor(user *iamv1alpha2.User) error {
```
zryfish, line 169: I think the word `shrink` is more suitable here, `shrinkEntriesFor`

pkg/models/auth/authenticator.go
```diff
@@ -67,22 +68,22 @@ type userGetter struct {
...
 		authOptions: options,
 	}
 	return passwordAuthenticator
 }
 
-func NewOAuth2Authenticator(ksClient kubesphere.Interface,
-	userLister iamv1alpha2listers.UserLister,
-	options *authoptions.AuthenticationOptions) OAuth2Authenticator {
+func NewOAuthAuthenticator(ksClient kubesphere.Interface,
+	ksInformer informers.SharedInformerFactory,
```
zryfish, line 82: Why change to SharedInformerFactory since only UserList used here

pkg/apiserver/authentication/identityprovider/ldap/ldap.go
```diff
@@ -126,7 +126,10 @@ func (l ldapProvider) Authenticate(username string, password string) (identitypr
 		return nil, err
 	}
 
-	filter := fmt.Sprintf("(&(%s=%s)%s)", l.LoginAttribute, username, l.UserSearchFilter)
+	filter := fmt.Sprintf("(%s=%s)", l.LoginAttribute, ldap.EscapeFilter(username))
```
RolandMa1986, line 129: Not sure whether this is releated with the feature. Please create another PR if it related some other issue..

wansir: Use a clearer way to splice, code optimization.

pkg/models/auth/authenticator.go
wansir: Just like we don’t use `iamv1alpha2.IamV1alpha2Interface` directly, I think `kubesphere.SharedInformerFactory` and `kubesphere.Interface` are an integral part, and using `iamv1alpha2.UserLister` directly will make the code very complicated, right?

pkg/apiserver/authentication/options/authenticate_options.go
wansir: The default value has been modified to 100.

pkg/apiserver/authentication/options/authenticate_options.go
wansir: fixed

pkg/controller/loginrecord/loginrecord_controller.go
wansir: fixed

pkg/controller/loginrecord/loginrecord_controller.go
wansir: fixed

pkg/models/auth/authenticator.go
zryfish: not like that. sharedinfomerfactory is very big, you just use a small part of it, it means somewhere else has to create such a big instance for you. Also it's not friendly for unit testing

wansir: fixed

Issue #3428, opened 2021-03-04
kubesphere 3.0 servicemesh not support kubernetes service has multi-port


haormj:
**Describe the Bug**

kubesphere servicemesh not support kubernetes service with multi-port, if your service has multi-port, ks create VirtualService will route traffic to the first port

**Versions Used**
KubeSphere: 3.0
Kubernetes: (If KubeSphere installer used, you can skip this)

**Environment**
How many nodes and their hardware configuration:

**How To Reproduce**
1. enable servicemesh
2. create service with multi-port
3. look at VirtualService crd

**Expected behavior**
kubesphere servicemesh support k8s service multi-port

haormj:
@zackzhangkai

haormj:
@zackzhangkai nice, I will test it

PR #3437, opened 2021-03-08
support virtualservice multi port

zackzhangkai:

**What type of PR is this?**

If a service has multi port, the traffic will always be routed to the first one. So we should support multi port service.

https://github.com/kubesphere/kubesphere/issues/3428

### DIFF
--- a/pkg/controller/utils/servicemesh/application.go
+++ b/pkg/controller/utils/servicemesh/application.go
@@ -118,6 +118,7 @@ func FillDestinationPort(vs *v1alpha3.VirtualService, service *corev1.Service) {
 				vs.Spec.Http[i].Route[j].Destination.Port = &apiv1alpha3.PortSelector{
 					Number: uint32(service.Spec.Ports[0].Port),
 				}
+				vs.Spec.Http[i].Match[j].Port = uint32(service.Spec.Ports[0].Port)
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


pkg/controller/utils/servicemesh/application.go
```diff
@@ -135,7 +136,18 @@ func FillDestinationPort(vs *v1alpha3.VirtualService, service *corev1.Service) {
...
+				vs.Spec.Tcp[i].Match[j].Port = uint32(service.Spec.Ports[0].Port)
 			}
 		}
 	}
 }
+
+func SupportHttpProtocol(protocol string) bool {
+	httpPro := []string{"http", "http2", "grpc"}
+	for _, i := range httpPro {
+		if protocol == i || strings.HasPrefix(protocol, i+"-") {
```
zryfish, line 148: case sensitive match?

zackzhangkai: https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/  shows supporting lower case

zryfish: https://github.com/istio/istio/blob/master/pkg/config/protocol/instance.go#L61 Please take a look at this, and I think we should add some test cases

zackzhangkai: Fixed

Issue #3422, opened 2021-03-02
Alarm is not retriggered after changing duration of alarm or alarm rule.


wenxin-01:
**Describe the bug**
Modify an alert policy that has already generated an alert without retriggering the alert.

**Environment**
http://xxxx.79:30104

Preset conditions
1、There is account A, which has cluster management privileges.
2、Use account A to log in to KS, and go to Platform management--》Cluster "B" management--》Alerting Policies.
3、There is an alarm policy 'test' that has already generated an alarm.
**To Reproduce**
1. Modify the alarm duration or alarm rules for the alarm policy 'test'.
2. Deploy applications to enable alarm policy to generate alarm.
3. Check to see if the alarm policy test has retriggered the alarm.

**Expected behavior**
The alarm policy 'test' has retriggered alert message.

The previously generated alert state should be cleared immediately after the rule change.

**Actual behavior**
The policy  ‘test’ did not trigger the alarm again.

/priority high
/area alerting
/cc @junotx
/kind bug
/milestone 3.1.0

junotx:
Did the rule trigger an alert before the modification? If so, does the application deployed after the modification satisfy the rule condition?  @wenxin-01

wenxin-01:
Alarm have been triggered before the modification, and the modified rule satisfies the trigger condition. @junotx

junotx:
If tha same alert was triggered before and after the rule changed, it would generally continue the previous alert until the duration condition is met, and then send the alert downstream.

wenxin-01:
【Expected behavior】The previously generated alert state should be cleared immediately after the rule change.

PR #3440, opened 2021-03-09
fix alerts not retriggered when updating rule

junotx:
**What type of PR is this?**

/kind bug
/area alerting

ensure alerts retriggered when updating alerting rule

Fixes https://github.com/kubesphere/kubesphere/issues/3422

/cc @benjaminhuo

### DIFF
--- a/pkg/models/alerting/rules/ruler.go
+++ b/pkg/models/alerting/rules/ruler.go
@@ -102,35 +102,99 @@ func (r *ruleResource) deleteAlertingRules(rules ...*RuleWithGroup) (bool, error
 // If there are rules to be updated, return true to indicate the resource should be updated.
 func (r *ruleResource) updateAlertingRules(rules ...*RuleWithGroup) (bool, error) {
 	var (
-		commit  bool
-		spec    = r.Spec.DeepCopy()
-		ruleMap = make(map[string]*RuleWithGroup)
+		commit     bool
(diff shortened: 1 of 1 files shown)


pkg/models/alerting/rules/ruler.go
```diff
@@ -102,38 +102,108 @@ func (r *ruleResource) deleteAlertingRules(rules ...*RuleWithGroup) (bool, error
...
 			}
 		}
 	}
 
+	for groupMax++; len(ruleMap) > 0; groupMax++ {
+		g := promresourcesv1.RuleGroup{Name: fmt.Sprintf("%s%d", customRuleGroupDefaultPrefix, groupMax)}
+		var num = customRuleGroupSize
+		if num > 0 {
+			for name, rule := range ruleMap {
+				if num <= 0 {
```
zryfish, line 189: why there is `num <= 0` since the condition is `num > 0`

junotx: `num` is a decreasing var to record the number of rules that can be added to the group

zryfish: I see. Isn't it better to make it a function? There are similar codes above.

junotx: It is made now

benjaminhuo (approved):
/lgtm

PR #3442, opened 2021-03-09
add extra annotations and labels to helm release

### LINKED ISSUE #1388: app store部署的app自带的pvc没有创建者
* [ ] 如下图，应该是创建此app的当前登录用户
Screen Shot 2019-11-06 at 6 03 37 PM

xyz-li:
Signed-off-by: LiHui <andrewli@yunify.com>

**What type of PR is this?**
Add extra annotations and labels to helm release when install new release or upgrade a release. 

/kind bug
/area app-management

Fixes #1388

### DIFF
--- a/build/ks-apiserver/Dockerfile
+++ b/build/ks-apiserver/Dockerfile
@@ -1,12 +1,17 @@
 # Copyright 2020 The KubeSphere Authors. All rights reserved.
 # Use of this source code is governed by an Apache license
 # that can be found in the LICENSE file.
-FROM alpine/helm:3.4.2 as helm-base
-
 FROM alpine:3.11
 
+ARG HELM_VERSION=v3.5.2
+
 RUN apk add --no-cache ca-certificates
-COPY --from=helm-base /usr/bin/helm /usr/bin/helm
+# install helm
+RUN wget https://get.helm.sh/helm-${HELM_VERSION}-linux-amd64.tar.gz && \
+    tar xvf helm-${HELM_VERSION}-linux-amd64.tar.gz && \
+    rm helm-${HELM_VERSION}-linux-amd64.tar.gz && \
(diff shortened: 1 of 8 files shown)


build/ks-apiserver/Dockerfile
```diff
@@ -1,12 +1,11 @@
 # Copyright 2020 The KubeSphere Authors. All rights reserved.
 # Use of this source code is governed by an Apache license
 # that can be found in the LICENSE file.
-FROM alpine/helm:3.4.2 as helm-base
+FROM openpitrix/helm:v0.1.0 as base
```
LinuxSuRen, line 4: Why change it? By the way, we cannot know the version of helm from image `openpitrix/helm:v0.1.0`. It's unclear.

xyz-li: 1. alpine/helm only has one `arm` version, which is alpine/helm:3.5.0-rc.1. So we beed build arm version ourself. [https://github.com/kubesphere/docker-helm/blob/main/Dockerfile#L4-L5](docker-helm) 
2. ks-controller-manager needs kustomize,  openpitrix/helm:v0.1.0 contains both helm and kustomize.

zryfish: yeah, agree with @LinuxSuRen we should set helm/kustomize version explicitly in Dockerfile, shouldn't depend on `openpitrix/helm:v0.1.0`, if not, what if the user wants to build an image with lastest helm/kustomize version?

zheng1: Neither helm nor kustomize has a suitable docker image (the community's current version does not support arm64), so we opened a separate project https://github.com/kubesphere/docker-helm to build the image we needed. helm and kustomize versions are defined in this file https://github.com/kubesphere/docker-helm/blob/main/Dockerfile#L4-L5

PR #3447, opened 2021-03-10
fix bug can not search auditing logs with nil namespace

wanjunlei:

**What type of PR is this?**
>
/kind bug

fix bug can not search auditing logs with nil namespace

```
/assign @zryfish 
/assign @benjaminhuo 
```


benjaminhuo (approved):
/lgtm

PR #3588, opened 2021-03-25
Fix fields missing when convert Jenkinsfile to JSON

LinuxSuRen:
**What type of PR is this?**
/kind bug
/area devops

Some fields was missing due to the struct does not have all the fields which comes from Jenkins. Such as, the fields `environment` missed.

I changed the struct to a generic data. There're two reasons here:
* A fixed go struct might need to change again once Jenkins has new features
* No other references with the struct, so it's ok that we removed it

Fixes kubesphere/console#1575

### DIFF
--- a/pkg/kapis/devops/v1alpha2/register.go
+++ b/pkg/kapis/devops/v1alpha2/register.go
@@ -614,6 +614,13 @@ func AddPipelineToWebService(webservice *restful.WebService, devopsClient devops
 			Writes(devops.ResJenkinsfile{}))
 
 		// match /pipeline-model-converter/toJson
+		/*
+		 * Considering the following reasons, we use a generic data struct here.
+		 * - A fixed go struct might need to change again once Jenkins has new features.
+		 * - No refer requirement for the specific data struct
+		 * Please read the official document if you want to know more details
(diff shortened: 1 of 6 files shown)


pkg/kapis/devops/v1alpha2/register.go
```diff
@@ -621,8 +621,8 @@ func AddPipelineToWebService(webservice *restful.WebService, devopsClient devops
 			Produces("application/json", "charset=utf-8").
 			Doc("Convert jenkinsfile to json format. Usually the frontend uses json to show or edit pipeline").
 			Reads(devops.ReqJenkinsfile{}).
-			Returns(http.StatusOK, RespOK, devops.ResJson{}).
-			Writes(devops.ResJson{}))
+			Returns(http.StatusOK, RespOK, map[string]interface{}{}).
+			Writes(map[string]interface{}{}))
```
zryfish, line 625: Why changing to `map[string]interface`, this is not going to generate readable API documentation.

LinuxSuRen: As mentioned https://github.com/kubesphere/kubesphere/pull/3588#issue-600595695 here, we get the JSON data directly from Jenkins. It always causes problems if the Jenkins side makes changes. Users can read the document from Jenkins if they really needs that one.

zryfish: OK, please add these comments to codebase

LinuxSuRen: Done.

PR #3654, opened 2021-03-31
Fix owner reference conflict

### LINKED ISSUE #3652: infinite reconciling loop between kubefed namespace controller and ks namespace controller
**Describe the Bug**
Ks ns controller will add `ownerReference` field to a ns if the ns has label such as `kubesphere.io/workspace: xccc`. However, if the ns is created by kubefed(e.g. federatednamespace), the kubefed ns controller will remove the `ownerReference` field. In this situation, it can cause competition between the two controllers. Finally, we can not create `federatednamespace` anymore.

**Versions Used**
KubeSphere: latest

**How To Reproduce**
1. create a multi-cluster-project that contains `host` cluster.
2. create another multi-cluster-project contains `host` and `member` cluster.
3. can not create ns in the `member` cluster.

**Expected behavior**
The ns should be created in the `member` cluster.

wansir:

**What type of PR is this?**

/kind bug

When namespace managed by kubefed controller manager, setting workspace owner reference will conflict.

Create namespace on a host cluster,  need to add the `kubefed.io/managed: false` label. 

/assign @leoendless 

Fixes #3652

### DIFF
--- a/pkg/controller/namespace/namespace_controller.go
+++ b/pkg/controller/namespace/namespace_controller.go
@@ -133,13 +133,19 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	// Bind to workspace if the namespace created by kubesphere
 	_, hasWorkspaceLabel := namespace.Labels[tenantv1alpha1.WorkspaceLabel]
-	if hasWorkspaceLabel {
-		if err := r.bindWorkspace(rootCtx, logger, namespace); err != nil {
-			return ctrl.Result{}, err
-		}
-	} else {
-		if err := r.unbindWorkspace(rootCtx, logger, namespace); err != nil {
-			return ctrl.Result{}, err
+	// if the namespace doesn't have a label like kubefed.io/managed: "true" (single cluster environment)
+	// or it has a label like kubefed.io/managed: "false"(multi-cluster environment), we set the owner reference filed.
+	// Otherwise, kubefed controller will remove owner reference.
+	kubefedManaged := namespace.Labels[constants.KubefedManagedLabel] == "true"
+	if !kubefedManaged {
+		if hasWorkspaceLabel {
+			if err := r.bindWorkspace(rootCtx, logger, namespace); err != nil {
+				return ctrl.Result{}, err
+			}
+		} else {
+			if err := r.unbindWorkspace(rootCtx, logger, namespace); err != nil {
+				return ctrl.Result{}, err
+			}
 		}
 	}
 	// Initialize roles for devops/project namespaces if created by kubesphere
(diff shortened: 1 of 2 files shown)


pkg/controller/namespace/namespace_controller.go
```diff
@@ -133,7 +133,8 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	// Bind to workspace if the namespace created by kubesphere
 	_, hasWorkspaceLabel := namespace.Labels[tenantv1alpha1.WorkspaceLabel]
-	if hasWorkspaceLabel {
+	kubefedManaged := namespace.Labels[constants.KubefedManagedLabel] == "true"
```
yuswift, line 136: I think we should add some comments under this line. Just like:
// if the namespace doesn't have a label like `kubefed.io/managed: "true"`(single cluster environment) or it has a label like `kubefed.io/managed: "false"`(multi-cluster environment), we set the owner reference filed. Otherwise, we don't set the owner reference filed.

pkg/controller/workspace/workspace_controller.go
```diff
@@ -137,8 +137,9 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 }
 
 func (r *Reconciler) bindWorkspace(ctx context.Context, logger logr.Logger, namespace *corev1.Namespace, workspace *tenantv1alpha1.Workspace) error {
+	kubefedManaged := namespace.Labels[constants.KubefedManagedLabel] == "true"
 	// owner reference not match workspace label
-	if !metav1.IsControlledBy(namespace, workspace) {
+	if !metav1.IsControlledBy(namespace, workspace) && !kubefedManaged {
```
yuswift, line 142: Why do we check the kubefedManaged again? We have checked that before the function call.

pkg/controller/namespace/namespace_controller.go
wansir: done

pkg/controller/namespace/namespace_controller.go
```diff
@@ -133,7 +133,11 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	// Bind to workspace if the namespace created by kubesphere
 	_, hasWorkspaceLabel := namespace.Labels[tenantv1alpha1.WorkspaceLabel]
-	if hasWorkspaceLabel {
+	// if the namespace doesn't have a label like kubefed.io/managed: "true" (single cluster environment)
+	// or it has a label like kubefed.io/managed: "false"(multi-cluster environment), we set the owner reference filed.
+	// Otherwise, we don't set the owner reference filed.
+	kubefedManaged := namespace.Labels[constants.KubefedManagedLabel] == "true"
+	if hasWorkspaceLabel && !kubefedManaged {
```
RolandMa1986, line 140:
```suggestion
	if !kubefedManaged {
		if hasWorkspaceLabel {
			if err := r.bindWorkspace(rootCtx, logger, namespace); err != nil {
				return ctrl.Result{}, err
			}
		} else {
			if err := r.unbindWorkspace(rootCtx, logger, namespace); err != nil {
				return ctrl.Result{}, err
			}
		}
	}
```
The results should be the same, but the logic seems more clear.

wansir: fixed

PR #3662, opened 2021-03-31
fix:group was updated continuously in member cluster

RolandMa1986:
**What type of PR is this?**
/kind bug

This is a similar issue to PR https://github.com/kubesphere/kubesphere/pull/3654
Groups in the member clusters shouldn't be updated by the group controller manager. Because they are managed by the kubefed.

ref #3654

### DIFF
--- a/pkg/controller/group/group_controller.go
+++ b/pkg/controller/group/group_controller.go
@@ -146,7 +146,7 @@ func (c *Controller) reconcile(key string) error {
 		}
 
 		if group.Labels != nil {
-			// Set OwnerReferences when the group has a parent. And it's owned by kubefed
+			// Set OwnerReferences when the group has a parent. And it's not owned by kubefed
 			if group.Labels[constants.KubefedManagedLabel] != "true" {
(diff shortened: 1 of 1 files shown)


pkg/controller/group/group_controller.go
```diff
@@ -146,29 +146,30 @@ func (c *Controller) reconcile(key string) error {
...
-				}
-				groupParent, err := c.groupLister.Get(parent)
-				if err != nil {
-					if errors.IsNotFound(err) {
-						utilruntime.HandleError(fmt.Errorf("Parent group '%s' no longer exists", key))
-						delete(group.Labels, iam1alpha2.GroupParent)
-					} else {
-						klog.Error(err)
-						return err
+			// Set OwnerReferences when the group has a parent. And it's owned by kubefed
```
wansir, line 149:
```suggestion
			// Set OwnerReferences when the group has a parent. And it's not owned by kubefed.
```

RolandMa1986: fixed

PR #3691, opened 2021-04-07
add kubesphereversion field for cluser.status

yuswift:

**What type of PR is this?**
/kind feature

Add kubesphereVersion field for cluser.status, which can be used to distinguish between different ks versions.

### DIFF
--- a/pkg/apis/cluster/v1alpha1/cluster_types.go
+++ b/pkg/apis/cluster/v1alpha1/cluster_types.go
@@ -136,6 +136,9 @@ type ClusterStatus struct {
 	// GitVersion of the kubernetes cluster, this field is populated by cluster controller
 	KubernetesVersion string `json:"kubernetesVersion,omitempty"`
 
+	// GitVersion of the /kapis/version api response, this field is populated by cluster controller
+	KubeSphereVersion string `json:"kubeSphereVersion,omitempty"`
+
 	// Count of the kubernetes cluster nodes
 	// This field may not reflect the instant status of the cluster.
 	NodeCount int `json:"nodeCount,omitempty"`
--- a/pkg/controller/cluster/cluster_controller.go
+++ b/pkg/controller/cluster/cluster_controller.go
@@ -51,6 +51,7 @@ import (
 	clusterclient "kubesphere.io/kubesphere/pkg/client/clientset/versioned/typed/cluster/v1alpha1"
 	clusterinformer "kubesphere.io/kubesphere/pkg/client/informers/externalversions/cluster/v1alpha1"
 	clusterlister "kubesphere.io/kubesphere/pkg/client/listers/cluster/v1alpha1"
+	"kubesphere.io/kubesphere/pkg/version"
 )
 
 // Cluster controller only runs under multicluster mode. Cluster controller is following below steps,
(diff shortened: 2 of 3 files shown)


pkg/apis/cluster/v1alpha1/cluster_types.go
```diff
@@ -136,6 +136,9 @@ type ClusterStatus struct {
 	// GitVersion of the kubernetes cluster, this field is populated by cluster controller
 	KubernetesVersion string `json:"kubernetesVersion,omitempty"`
 
+	// GitVersion of the /kapis/version api response, this field is populated by cluster controller
+	KubeSphereVersion string `json:"kubesphereVersion,omitempty"`
```
LinuxSuRen, line 140: letter S should be capitalized.

`kubesphereVersion` to `kubeSphereVersion`

pkg/controller/cluster/cluster_controller.go
```diff
@@ -660,6 +668,42 @@ func (c *clusterController) tryToFetchKubeSphereComponents(host string, transpor
...
+
+	info := version.Info{}
+	decoder := json.NewDecoder(response.Body)
+	err = decoder.Decode(&info)
+	if err != nil {
+		return "", err
+	}
+
+	//for kubesphere v3.0.0, the gitVersion is always v0.0.0, so we return v3.0.0
+	if info.GitVersion == "v0.0.0" {
```
LinuxSuRen, line 696: What about the previous version? Such 2.1?

yuswift: Currently, we don't support `Kubesphere v2.1` as a member cluster. @zryfish

pkg/apis/cluster/v1alpha1/cluster_types.go
yuswift: Thanks for your suggestion. I've pushed code.

pkg/controller/cluster/cluster_controller.go
LinuxSuRen: I'd suggest that return `unknown` instead of a possible value.

pkg/apis/cluster/v1alpha1/cluster_types.go
```diff
@@ -136,6 +136,9 @@ type ClusterStatus struct {
 	// GitVersion of the kubernetes cluster, this field is populated by cluster controller
 	KubernetesVersion string `json:"kubernetesVersion,omitempty"`
 
+	// GitVersion of the /kapis/version api response, this field is populated by cluster controller
+	KubesphereVersion string `json:"kubesphereVersion,omitempty"`
```
LinuxSuRen, line 140: Actually, I suggest make it to `KubeSphereVersion string `json:"kubeSphereVersion,omitempty"``

`kube` and `sphere` are two words.

pkg/controller/cluster/cluster_controller.go
yuswift: `KubeSphere v2.1` will return 404 cause the version doesn't implement the `/kapis/version` API. Then we will not update the status.kubeSphereVersion filed. Do we still add an `unknown` value?

pkg/apis/cluster/v1alpha1/cluster_types.go
yuswift: Sorry for the misunderstanding. Done.

LinuxSuRen: No worries.

pkg/controller/cluster/cluster_controller.go
LinuxSuRen: Got it. How about adding these comments to the source file? It would be much more clear for other contributors.

yuswift: Writing comments is quite necessary. In my opinion, `KubeSphere v2.1` can not be joined as a member cluster, so it will never be reconciled. Maybe adding those comments will confuse other contributors.

yuswift: I added some comments. Hope it will be helpful.

LinuxSuRen (approved):
/lgtm

PR #3694, opened 2021-04-07
check release resources status

xyz-li:
**What type of PR is this?**
/kind feature
/area app-management

Check release resource status. When all the resource are ready, update helm release status to active.


xyz-li:
#3701

PR #3696, opened 2021-04-07
fix bug member cluster can not view the notification channels

wanjunlei:

**What type of PR is this?**

 /kind bug

In the multi-cluster environment, the host cluster will synchronize the secret of notification configs and receivers by `FederatedSecret`. 
The `FederatedSecret` cannot synchronize the label to the member through the `template`, the member cluster will  display the notification configuration incorrectly because of the secret in the member cluster is missing some required labels. 
So it need to use `Overrides` to synchronize the labels. The notification controller will watch cluster resources. When a member cluster is created or deleted, the notification controller will update the `FederatedSecret` Overrides

```
/assign @zryfish 
/assign @benjaminhuo 
```

### DIFF
--- a/pkg/controller/notification/notification_controller.go
+++ b/pkg/controller/notification/notification_controller.go
@@ -124,7 +124,7 @@ func (c *Controller) setEventHandlers() error {
 		}
 	}
 
-	// Watch the cluster add and delete operator.
+	// Watch the cluster add and delete operations.
 	if informer, err := c.ksCache.GetInformer(context.Background(), &v1alpha1.Cluster{}); err != nil {
 		klog.Errorf("get cluster informer error, %v", err)
 		return err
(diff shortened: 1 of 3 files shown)


pkg/controller/notification/notification_controller.go
```diff
@@ -433,9 +473,72 @@ func (c *Controller) syncFederatedSecret(obj *corev1.Secret) error {
...
+					},
+				},
+			},
+		})
+	}
+
+	return nil
+}
+
+// Update the annotations of secrets which the notification controller managed, rigger a  reconcile.
```
benjaminhuo, line 522: =>
// Update the annotations of secrets managed by the notification controller to trigger a reconcile.

pkg/controller/notification/notification_controller.go
```diff
@@ -224,15 +239,28 @@ func (c *Controller) reconcile(obj interface{}) error {
 	}
 
 	name := accessor.GetName()
-	kind := runtimeObj.GetObjectKind().GroupVersionKind().String()
+
+	// If the cluster changed, add or delete, it should update the annotations of secrets
```
benjaminhuo, line 243: =>
	// The notification controller should update the annotations of secrets managed by itself
         // whenever a cluster is added or deleted. This way, the controller will have a chance to override the secret.

benjaminhuo (changes requested):

benjaminhuo (approved):
/lgtm

PR #3698, opened 2021-04-08
fix bug where the edge node container groups tab could not see pod metrics

zhu733756:
Signed-off-by: zhu733756 <talonzhu@yunify.com>

**What type of PR is this?**
> /kind bug

With the console api changed, the backend cannnot get  namespace/podname combinations from the previous parameter `resources_filter` because of the new parameter changed to `namespaced_resources_filter`. This pr aims to fix this bug. 

fix bug where the edge node container groups could not see pod metrics


pkg/simple/client/monitoring/metricsserver/metricsserver_test.go
```diff
@@ -213,32 +213,36 @@ func TestGetNamedMetrics(t *testing.T) {
...
 		},
 		{
-			metrics:       []string{"pod_cpu_usage", "pod_memory_usage_wo_cache"},
-			filter:        "default/pod2$",
-			expected:      "metrics-vector-5.json",
-			podName:       "",
-			namespaceName: "",
+			metrics:          []string{"pod_cpu_usage", "pod_memory_usage_wo_cache"},
+			filter:           "",
+			namespacedFilter: "default/pod2$",
```
benjaminhuo, line 234: Should we add more entries here in the test? because `namespacedFilter` could be like this `default/pod2|default/pod3|default/pod4`

benjaminhuo (changes requested):
See my comments above

pkg/simple/client/monitoring/metricsserver/metricsserver_test.go
zhu733756: Fixed

benjaminhuo (approved):
Why adding a WIP to the title？

zhu733756:
> Why adding a WIP to the title？
Add this label when the entries was under collecting. Now I remove it.

benjaminhuo (approved):
/lgtm

PR #3702, opened 2021-04-08
upgrade crd to apiextensions.k8s.io/v1 && generate crd files

yuswift:

**What type of PR is this?**
/kind cleanup

upgrade crd to apiextensions.k8s.io/v1 and put those files into one folder.


zryfish:
Unit test failed, please have a look. Also, our resources API still depend on `v1beta1` https://github.com/kubesphere/kubesphere/blob/master/pkg/models/resources/v1alpha3/customresourcedefinition/customresourcedefinitions.go#L45, maybe we need to create an issue to track

yuswift:
> Unit test failed, please have a look. Also, our resources API still depend on `v1beta1` https://github.com/kubesphere/kubesphere/blob/master/pkg/models/resources/v1alpha3/customresourcedefinition/customresourcedefinitions.go#L45, maybe we need to create an issue to track
Sure. I will create an issue for the upgrade. BTW, I think we also need to update the codes that are stilling using crd/v1beta1 to crd/v1 cause v1.15 is not officially supported.

zackzhangkai:
@yuswift  Should use controller-tools 0.4.1 to avoid the `Any` bug. Then do the `make manifest`, some type error will occur. Just ignore the errors. 

```
	sigs.k8s.io/controller-tools => sigs.k8s.io/controller-tools v0.4.1
```

zackzhangkai:
> @yuswift Should use controller-tools 0.4.1 to avoid the `Any` bug. Then do the `make manifest`, some type errors will occur. Just ignore the errors. The error is a known issue. [istio/api#1482](https://github.com/istio/api/issues/1482). Which can be fixed by the 0.5.0 by comment with `crd:validation:Schemaless`
> 
> ```
> go.mod 
>   replace
> 	sigs.k8s.io/controller-tools => sigs.k8s.io/controller-tools v0.4.1
> ```
> 
> After the crd generated, should modify the value to comment. [kubernetes-sigs/controller-tools#245](https://github.com/kubernetes-sigs/controller-tools/issues/245) :
> 
> `value: {}` -> `#value: {}`

 Change `value: {}` to

```
  value: {
    type: number
  }
```

seems greater.

yuswift:
/cc @zryfish 
/cc @zackzhangkai 
This pr also upgraded `controller-tools` to v0.4.1.

Issue #3695, opened 2021-04-07
when changing the alarm rule of the alarm policy, cannot receive the notification after the alarm is triggered


liuyp2018:
**Environment**
ks v3.1

**How To Reproduce**
1. Create a alert policy with a alert rule for CPU utilization
2. Wait until the alarm policy triggers the alarm
3. Change the alert rule of the alert policy to available memory
4. Wait until the alert policy reaches the trigger condition, and the alert policy's state changes to trigger
5. Check the alert notification 

**Expected results:**
can receive the alert notification 

**Actual results:**
No alert notification received

/kind bug
/assign @junotx 
/priority High
/milestone 3.1.0
/area alerting

benjaminhuo:
@liuyp2018 This should have been fixed

liuyp2018:
verified, and fixed
/close

PR #3706, opened 2021-04-08
fix notification triggering again after alerting rule changes

junotx:
**What type of PR is this?**

/kind bug

add `rule_id` to rule labels to so that the new alert message will be different from the previous's to be sent as notification in time.

ref https://github.com/kubesphere/kubesphere/issues/3695

/area alerting
/cc @benjaminhuo


benjaminhuo (approved):
/lgtm

PR #3743, opened 2021-04-15
Fix: return all items when limit=-1

### LINKED ISSUE #3739: When querying all apps in a project, the interface returns 500
**Describe bug**

when set limit=-1, the interface returns 500

**Environment**
http://139.198.121.240:30881

**How To Reproduce**

kapis/openpitrix.io/v1/workspaces/ws/clusters/host/namespaces/hpro/applications?conditions=status=creating|active|failed|deleting|upgrading|created|upgraded&paging=limit=-1,page=1&orderBy=status_time

**Expected behavior**
The interface returns all apps

**Actual behavior**
image

image

/priority medium
/assign @xyz-li 
/kind bug
/milestone 3.1.0

xyz-li:
Signed-off-by: LiHui <andrewli@yunify.com>

**What type of PR is this?**
/kind bug

Fix: return all items when parameter limit equals 1

Fixes #3739

### DIFF
--- a/pkg/models/openpitrix/applications.go
+++ b/pkg/models/openpitrix/applications.go
@@ -22,6 +22,8 @@ import (
 	"strings"
 	"time"
 
+	"kubesphere.io/kubesphere/pkg/apiserver/query"
+
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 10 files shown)


zryfish:
The whole limit, page idea is to avoid return all of items at one time, we should set default limit instead.

xyz-li:
> The whole limit, page idea is to avoid return all of items at one time, we should set default limit instead.

Resources api are all return all the items when limit equals -1

pkg/models/openpitrix/applications.go
```diff
@@ -308,19 +308,20 @@ func (c *applicationOperator) ListApps(conditions *params.Conditions, orderBy st
 		sort.Sort(HelmApplicationList(apps))
 	}
 
-	items := make([]interface{}, 0, limit)
+	totalCount := len(apps)
```
zryfish, line 311: why not using https://github.com/kubesphere/kubesphere/blob/master/pkg/apiserver/query/types.go#L39 to parse request parameters since there are already code to deal with pagination?

xyz-li: The parameter of openpitrix is different from resource api, so I just use the function GetValidPagination.

zheng1: This part of the code will be refactored in the next version. This version is temporarily compatible with the old api

Issue #3766, opened 2021-04-19
Case insensitive problem with LDAP users


liuyp2018:
**steps**
1. Log in to KS using the TesUser1 user in LDAP and enter the user's mailbox and user name configuration page
2. Set the user email as user1@test.com, set the user name as testuser1, and log in KS
3. Log in to KS with testuser1, and the user's email and username configuration page pops up again

Cause of problem: LDAP users are validated in both uppercase and lowercase

https://user-images.githubusercontent.com/36271543/115207325-12264e80-a12e-11eb-8e3b-9730654a3d55.mp4

/priority Medium
/kind bug
/milestone 3.1.0
/assign @wansir

liuyp2018:
fixed
/close

PR #3768, opened 2021-04-19
Fix issue with case-sensitive username

wansir:

**What type of PR is this?**

/kind bug

In case of a case-insensitive authentication (which is the default behavior : https://stackoverflow.com/questions/29897684/is-ldap-dn-case-insensitive), the final User username can be wrong (for instance, USERNAME instead of UserName).

To avoid this issue, using the `uid` attribute of the Ldap\Entry instead.

Fixes #https://github.com/kubesphere/kubesphere/issues/3766

### DIFF
--- a/pkg/apiserver/authentication/identityprovider/ldap/ldap.go
+++ b/pkg/apiserver/authentication/identityprovider/ldap/ldap.go
@@ -119,11 +119,11 @@ func (l ldapProvider) Authenticate(username string, password string) (identitypr
 		klog.Error(err)
 		return nil, err
 	}
+
 	conn.SetTimeout(time.Duration(l.ReadTimeout) * time.Millisecond)
 	defer conn.Close()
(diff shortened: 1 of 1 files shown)


pkg/apiserver/authentication/identityprovider/ldap/ldap.go
```diff
@@ -147,21 +147,27 @@ func (l ldapProvider) Authenticate(username string, password string) (identitypr
 		return nil, err
 	}
 
-	if len(result.Entries) != 1 {
-		return nil, errors.NewUnauthorized("incorrect password")
+	var entry *ldap.Entry
+	switch n := len(result.Entries); n {
```
zryfish, line 151: Using the switch statement here is wired to me. What's wrong with a normal `if else` statement?

wansir: Use `if else` statement instead.

PR #3796, opened 2021-04-23
Fix SelectorFromSet ignore errors and select everything

### LINKED ISSUE #3784: [priority high]After creating a long user name in KubeSphere 3.0, deleting the user will cause the system to crash.
**Describe the Bug**
KubeSphere 3.0 installed with ks-installer,after creating a long user name in KubeSphere 3.0, deleting the user will cause the system to crash.

**How To Reproduce**
1.For example, the user name as shown below.
image
2.Delete the user in the image above.
3.When accessing the system again, the browser reported the following status, which made it unavailable.
image

**Versions Used**
KubeSphere: 3.0（It is expected that 3.1 will also have this problem.）
Kubernetes: 1.18.8

**Expected behavior**
1.Add verification when creating users, and prohibit creating users with long user names

/kind bug
/priority high

wansir:

**What type of PR is this?**

/kind bug

Fix SelectorFromSet ignore errors and select everything, should check label value early. Ignore invalid label value error if the user name is more than 63 characters.

Fixes #3784

### DIFF
--- a/pkg/controller/group/group_controller.go
+++ b/pkg/controller/group/group_controller.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"reflect"
 
+	"k8s.io/apimachinery/pkg/util/validation"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


zryfish:
Will not merge while we approaching release date
/hold

PR #3823, opened 2021-04-28
multi arch build error

lxm:
Signed-off-by: lxm <lxm.xupt@gmail.com>

fix multi arch build error

### DIFF
--- a/hack/docker_build.sh
+++ b/hack/docker_build.sh
@@ -40,7 +40,6 @@ if [[ $? != 0 ]]; then
   echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
 fi
 
-TAG=$TAG-multiarch
 docker build -f build/ks-apiserver/Dockerfile -t $REPO/ks-apiserver:$TAG .
 docker push $REPO/ks-apiserver:$TAG
 # print the full docker image path for your convience
--- a/.github/workflows/build-multiarch.yaml
+++ b/.github/workflows/build-multiarch.yaml
@@ -5,7 +5,6 @@ on:
     branches:
       - 'master'
       - 'release*'
-      - '*'
     tags:
       - 'v*'
   pull_request:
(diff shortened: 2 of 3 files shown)


hack/docker_build.sh
```diff
@@ -40,7 +40,7 @@ if [[ $? != 0 ]]; then
   echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
 fi
 
-TAG=$TAG-multiarch
+TAG=$TAG
```
LinuxSuRen, line 43: I think this line should be removed.

`TAG=$TAG` has no meaning.

LinuxSuRen (changes requested):
Thanks for the fixing. Please check out my comments.

By the way, I don't understand the following code line in the multi-arch build GitHub action:

https://github.com/kubesphere/kubesphere/blob/cad6296f4e27765ca58b931e2c3a4afb4af6a963/.github/workflows/build-multiarch.yaml#L8

```
on:
  push:
    branches:
      - 'master'
      - 'release*'
      - '*'
```

The wildcard `*` means all branches. I would suggest removing the wildcard. Or please provide a reasonable description.

lxm:
> Thanks for the fixing. Please check out my comments.
> 
> By the way, I don't understand the following code line in the multi-arch build GitHub action:
> 
> https://github.com/kubesphere/kubesphere/blob/cad6296f4e27765ca58b931e2c3a4afb4af6a963/.github/workflows/build-multiarch.yaml#L8
> 
> ```
> on:
>   push:
>     branches:
>       - 'master'
>       - 'release*'
>       - '*'
> ```
> 
> The wildcard `*` means all branches. I would suggest removing the wildcard. Or please provide a reasonable description.

is was used to test in my own repo, now remove it

hack/docker_build.sh
lxm: right, removed this line . it was unexpected to add to this file

LinuxSuRen (approved):
/lgtm

Issue #3845, opened 2021-05-06
灰度策略自定义 Header设置的值无法保存，一直为空


linkt:
kubesphere版本：3.1
设置了对应的header保存后，value为空，没有任何报错
![image](https://user-images.githubusercontent.com/7455401/117315092-f48d0f00-aeb9-11eb-8843-cb9f6841a2e2.png)
![image](https://user-images.githubusercontent.com/7455401/117314867-c14a8000-aeb9-11eb-82a1-3f1b0aec8ab4.png)
各位大佬是我设置的值格式不对吗

zackzhangkai:
Please use English. 

You can go to [forum](https://kubesphere.com.cn/forum/) to ask questions with Chinese.

Exec the command as belows to see if the configurations saved.

```bash
kubectl -n xxx get strategy tomcat3 -oyaml 
```


By the way, request by traffic percentage is ok?

Maybe this is a bug, I will check this later.

linkt:
@zackzhangkai 
[root@mvxl8417 mdauto]# kubectl get strategy tomcat3 -n my-test -o yaml
apiVersion: servicemesh.kubesphere.io/v1alpha2
kind: Strategy
metadata:
  annotations:
    kubesphere.io/creator: admin
    servicemesh.kubesphere.io/newWorkloadName: tomcat3-v2
    servicemesh.kubesphere.io/oldWorkloadName: tomcat3-v1
    servicemesh.kubesphere.io/workloadReplicas: "1"
    servicemesh.kubesphere.io/workloadType: deployments
  creationTimestamp: "2021-05-06T14:21:52Z"
  generation: 2
  labels:
    app: tomcat3
    app.kubernetes.io/name: tomcat3
    app.kubernetes.io/version: v1
  name: tomcat3
  namespace: my-test
  ownerReferences:
  - apiVersion: app.k8s.io/v1beta1
    blockOwnerDeletion: true
    controller: false
    kind: Application
    name: tomcat3
    uid: 8f10aceb-31d9-4fd5-a268-0ee678a4e2dc
  resourceVersion: "312701462"
  selfLink: /apis/servicemesh.kubesphere.io/v1alpha2/namespaces/my-test/strategies/tomcat3
  uid: f93e73f3-6501-469d-a840-593cbc9dbe48
spec:
  principal: v1
  selector:
    matchLabels:
      app: tomcat3
      app.kubernetes.io/name: tomcat3
      app.kubernetes.io/version: v1
[...]

PR #3795, opened 2021-04-23
add RawExtension type annotation

RolandMa1986:
Signed-off-by: Roland.Ma <rolandma@yunify.com>

**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
`Role` Filed in `RoleBas`e type was missing when upgrading CustomResourceDefinition from apiextensions.k8s.io/v1beta1 to apiextensions.k8s.io/v1.
According to the [Kubernetes doc ](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#rawextension), we need to set the below properties:

```yaml
type: object
properties:
  foo:
    x-kubernetes-embedded-resource: true
    x-kubernetes-preserve-unknown-fields: true
```

**Which issue(s) this PR fixes**:

Fixes #3766 3794

/cc @zryfish @wansir

PR #3848, opened 2021-05-07
fix servicemesh crd

zackzhangkai:

**What type of PR is this?**

Fix https://github.com/kubesphere/kubesphere/issues/3845

### DIFF
--- a/config/crds/servicemesh.kubesphere.io_strategies.yaml
+++ b/config/crds/servicemesh.kubesphere.io_strategies.yaml
@@ -259,8 +259,10 @@ spec:
                                     additionalProperties:
                                       description: Describes how to match a given string in HTTP headers. Match is case-sensitive.
                                       type: object
+                                      x-kubernetes-preserve-unknown-fields: true
                                     description: "The header keys must be lowercase and use hyphen as the separator, e.g. _x-request-id_. \n Header values are case-sensitive and formatted as follows: \n - `exact: \"value\"` for exact string match \n - `prefix: \"value\"` for prefix-based match \n - `regex: \"value\"` for ECMAscript style regex-based match \n If the value is empty and only the name of header is specfied, presence of the header is checked. **Note:** The keys `uri`, `scheme`, `method`, and `authority` will be ignored."
                                     type: object
+                                    x-kubernetes-preserve-unknown-fields: true
                                   ignore_uri_case:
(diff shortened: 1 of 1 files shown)


config/crds/servicemesh.kubesphere.io_strategies.yaml
```diff
@@ -259,8 +259,10 @@ spec:
                                     additionalProperties:
                                       description: Describes how to match a given string in HTTP headers. Match is case-sensitive.
                                       type: object
+                                      x-kubernetes-preserve-unknown-fields: true
```
zryfish, line 262: We should avoid change CRD YAML directly because it will be overwritten by controller-gen.

zackzhangkai: As usually, we should avoid change directly, but istio use [cue-gen](https://github.com/istio/tools/blob/master/cmd/cue-gen/README.md#generate-crds) tool to generate CRD. So there is something different about the markers from controller-gen. So we should modify it by hand, because controller-gen cannot generate it correctlly.

zryfish: https://github.com/kubesphere/kubesphere/pull/3795 Does this PR help?

zackzhangkai: no use. The markers only take effect in the root object. But this place should do in the lower object.https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#controlling-pruning

yuswift: I think we need to use cue-gen to regenerate crd files again instead of patching it manually, cause we don't know if we have modified the files completely. Need more suggestions from @zackzhangkai.

zackzhangkai: cue-gen is a tool just from istio. we just have two istio crds, no need to use this tool.  Indeed, we no need to focus on the tools, they are just help us to generate manifest, which based on [open API v3]([https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#schemaObject](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#schemaObject)). In the fact, we just to take a full test on our products, that's OK, I think.

yuswift: I will create another issue to track the test on service mesh.

zackzhangkai:
@zryfish Pls verify this PR.

PR #3859, opened 2021-05-10
Fixed namespace resource filter.

yunkunrao:
Signed-off-by: Rao Yunkun <yunkunrao@yunify.com>

**What type of PR is this?**

/kind bug

This PR fixed the error that namespace resource filter didn't take effect when namespace parameter was empty.


yunkunrao:
Fixed issue #3860

Issue #3847, opened 2021-05-07
Package core component with helm chart


RolandMa1986:
**What's it about?**
Create a helm chart for the core component. And ks-install would use the helm chart to install the core component.

**What's the reason why we need it?**
Currently, the Core component is deployed in ks-install with kubectl and YAML files. This means we had to maintain 2 copies of code, such as the CRDs. And there's no lifecycle management when using the kubectl, so we can hardly uninstall kubesphere.

To create a helm chart:
- We can automatically package charts during build the time. And just needs to synchronize from KubeSphere to ks-installer. 
- Full lifecycle management with helm such as install, upgrade, uninstall.
- Leverage built-in functions or webhooks, such as genSignedCert.

Please leave your comments below if there's anyone who agrees with me. Or just give me a thumb up.
@pixiake @zryfish 
**Area Suggestion**

/area installation
/kind feature-request

zryfish:
That's what we want to do for a long time.

yuswift:
It's really necessary for us to install ks-core with helm. I'm willing to work on this issue with Roland together.

PR #3896, opened 2021-05-19
create helm chart for ks-core

RolandMa1986:

**What type of PR is this?**
/kind feature

This is part of feature #3847. 

Fixes ##3847

1. Version label was removed from the Pod template. But the label will be kept when upgrading from an old version.

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: ks-apiserver
    tier: backend
    version: {{ .Chart.AppVersion }}
  name: ks-apiserver
  namespace: kubesphere-system
spec:
  selector:
    matchLabels:
      app: ks-apiserver
      tier: backend
      # version: {{ .Chart.]ppVersion }}
  template:
    metadata:
      labels:
        app: ks-apiserver
        tier: backend
        # version: {{ .Chart.AppVersion }}
```

Labels in template metadata are immutable. We can't apply the manifest again if the Version updated. If we use --force options, a rolling upgrade would be broken.

2. Should initial data be included in the Chart?

~~While I think it's not part of the Chart, Charts should focus on the infrastructures. Initial data like admin users and built-in roles can(should) be created when all the components are ready. So currently I left those steps in the ks-installer.
Please let me known, if you have other opinions.~~

Initial Admin User was moved to the Helm Chart because we need to control the creation sequence. User needs to be created before `users.iam.kubesphere.io` webhook or `ks-controller-managed` has been started successfully. Since the `ks-controller-managed`  won't be started before the creation of kubesphere-config. So we have to create it before the webhook. Moving it to the chart is the only way we can do it.

3. Which CRDs should be packaged to the Chart?

Currently, all the CRDs in the `config\crds` folders would be copied to the chart except the federated crds.

`make helm-package` to package ks-core helm chart
`make helm-deploy`  to install or upgrade ks-core to k8s cluster. 

4. Hardcode namespaces
Currently, the "kubesphere-controls-system" namespace is still hardcoded. This namespace and resources may be split into a sub-chart instead. But I prefer to create another PR if it's necessary.

### DIFF
--- a/config/ks-core/templates/ks-apiserver.yml
+++ b/config/ks-core/templates/ks-apiserver.yml
@@ -1,0 +1,129 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: ks-apiserver
+    tier: backend
+    version: {{ .Chart.AppVersion }}
+  name: ks-apiserver
+  namespace: kubesphere-system
+spec:
+  strategy:
+    rollingUpdate:
+      maxSurge: 0
+    type: RollingUpdate
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      app: ks-apiserver
+      tier: backend
+      # version: {{ .Chart.AppVersion }}
+  template:
+    metadata:
+      labels:
+        app: ks-apiserver
+        tier: backend
+        # version: {{ .Chart.AppVersion }}
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
+      - command:
+        - ks-apiserver
+        - --logtostderr=true
+        image: {{ .Values.image.ks_apiserver_repo }}:{{ .Values.image.ks_apiserver_tag | default .Chart.AppVersion }}
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        name: ks-apiserver
+        ports:
+        - containerPort: 9090
+          protocol: TCP
+        resources:
+          {{- toYaml .Values.apiserverResources | nindent 12 }}
+        volumeMounts:
+        - mountPath: /var/run/docker.sock
+          name: docker-sock
+        - mountPath: /etc/kubesphere/ingress-controller
+          name: ks-router-config
+        - mountPath: /etc/kubesphere/
+          name: kubesphere-config
+        - mountPath: /etc/localtime
+          name: host-time
+        livenessProbe:
+          failureThreshold: 8
+          httpGet:
+            path: /kapis/version
+            port: 9090
+            scheme: HTTP
+          initialDelaySeconds: 15
+          timeoutSeconds: 15
+      serviceAccountName: {{ include "ks-core.serviceAccountName" . }}
+      {{- with .Values.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      affinity:
+        nodeAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            preference:
+              matchExpressions:
+              - key: node-role.kubernetes.io/master
+                operator: In
+                values:
+                - ""
+{{- if gt .Values.replicaCount 1.0 }}
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - topologyKey: kubernetes.io/hostname
+            labelSelector:
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - ks-apiserver
+            namespaces:
+            - kubesphere-system
+{{- end }}
+      volumes:
+      - hostPath:
+          path: /var/run/docker.sock
+          type: ""
+        name: docker-sock
+      - configMap:
+          defaultMode: 420
+          name: ks-router-config
+        name: ks-router-config
+      - configMap:
+          defaultMode: 420
+          name: kubesphere-config
+        name: kubesphere-config
+      - hostPath:
+          path: /etc/localtime
+          type: ""
+        name: host-time
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    kubernetes.io/created-by: kubesphere.io/ks-apiserver
+  labels:
+    app: ks-apiserver
+    tier: backend
+    version: {{ .Chart.AppVersion }}
+  name: ks-apiserver
+  namespace: kubesphere-system
+spec:
+  ports:
+  - port: 80
+    protocol: TCP
+    targetPort: 9090
+  selector:
+    app: ks-apiserver
+    tier: backend
+    # version: {{ .Chart.AppVersion }}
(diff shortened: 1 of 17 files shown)


config/ks-core/templates/kubesphere-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+data:
+  kubesphere.yaml: |
+    authentication:
+      authenticateRateLimiterMaxTries: {{ .Values.config.authentication.authenticateRateLimiterMaxTries | default 10 }}
+      authenticateRateLimiterDuration: {{ .Values.config.authentication.authenticationRateLimiterDuration | default "10m0s" }}
+      loginHistoryRetentionPeriod: {{ .Values.config.authentication.loginHistoryRetentionPeriod | default "168h"  }}
+      maximumClockSkew: {{ .Values.config.authentication.maximumClockSkew | default "10s" }}
+      multipleLogin: {{ .Values.console.enableMultiLogin | default true }}
+      kubectlImage: {{ .Values.image.ks_kubectl_repo }}:{{ .Values.image.ks_kubectl_tag | default "latest" }}
+      jwtSecret: "{{ .Values.jwtSecret }}"
```
wansir, line 16: If `jwtSecret` is empty use a random string instead.

yuswift:
> /cc @pixiake @yuswift

Maybe we need to move charts [here](https://github.com/kubesphere/helm-charts/tree/master/src)
/cc @zryfish 
/cc @wansir 
/cc @pixiake 
/cc @RolandMa1986

config/ks-core/templates/kubesphere-config.yaml
RolandMa1986: @wansir  How about using `default (sha1sum uuidv4 ) ` which will generate a hash of the random uuid?

wansir: > @wansir How about using `default (sha1sum uuidv4 ) ` which will generate a hash of the random uuid?

[A key of the same size as the hash output (for instance, 256 bits for"HS256") or larger MUST be used with this algorithm.](https://tools.ietf.org/html/rfc7518#section-3.2)

```
jwtSecret:  {{ randAlphaNum 32 | quote }} `
```

RolandMa1986: Resoved

config/ks-core/templates/ks-console.yml
```diff
@@ -0,0 +1,118 @@
...
+          name: host-time
+        livenessProbe:
+          tcpSocket:
+            port: 8000
+          initialDelaySeconds: 15
+          timeoutSeconds: 15
+          periodSeconds: 10
+          successThreshold: 1
+          failureThreshold: 8
+      serviceAccount: kubesphere
```
yuswift, line 48: How about using variable here?

config/ks-core/templates/ks-console.yml
```diff
@@ -0,0 +1,118 @@
...
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - topologyKey: kubernetes.io/hostname
+            labelSelector:
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - ks-console
+            namespaces:
+            - kubesphere-system
```
yuswift, line 75: I'd prefer using {{ Release.Namespace }} to replace it.

config/ks-core/templates/ks-console.yml
```diff
@@ -0,0 +1,118 @@
...
+
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: ks-console
+    tier: frontend
+    version: {{ .Chart.AppVersion }}
+  name: ks-console
+  namespace: kubesphere-system
```
yuswift, line 104: The namespace shouldn't be fixed. I'd prefer to delete it and the template will be rendered when executing `helm -n kubesphere-system ...`

config/ks-core/templates/ks-controller-manager.yaml
```diff
@@ -0,0 +1,129 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: ks-controller-manager
+    tier: backend
+    version: {{ .Chart.AppVersion }}
+  name: ks-controller-manager
+  namespace: kubesphere-system
```
yuswift, line 9: As above.

config/ks-core/templates/ks-controller-manager.yaml
```diff
@@ -0,0 +1,129 @@
...
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - topologyKey: kubernetes.io/hostname
+            labelSelector:
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - ks-controller-manager
+            namespaces:
+            - kubesphere-system
```
yuswift, line 105: I'd prefer using `{{ Release.Namespace }} ` to replace it.

config/ks-core/templates/ks-controller-manager.yaml
```diff
@@ -0,0 +1,129 @@
...
+
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: ks-controller-manager
+    tier: backend
+    version: {{ .Chart.AppVersion }}
+  name: ks-controller-manager
+  namespace: kubesphere-system
```
yuswift, line 118: The namespace shouldn't be fixed. I'd prefer to delete it and the template will be rendered when executing `helm -n kubesphere-system ...`

config/ks-core/templates/ks-router-cm.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ks-router-config
+  namespace: kubesphere-system
```
yuswift, line 5: As above.

config/ks-core/templates/kubesphere-config.yaml
```diff
@@ -0,0 +1,34 @@
+{{- if .Values.config.create -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kubesphere-config
+  namespace: kubesphere-system
```
yuswift, line 6: As above.

config/ks-core/templates/ks-apiserver.yml
```diff
@@ -0,0 +1,129 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: ks-apiserver
+    tier: backend
+    version: {{ .Chart.AppVersion }}
+  name: ks-apiserver
+  namespace: kubesphere-system
```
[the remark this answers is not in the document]
RolandMa1986: @yuswift, the issue here is not that the namespace was hardcode in the helm chart, but it was also hardcode in the Go. So you can't install it in another namespace at all currently. A fixed namespace can avoid misunderstanding or incorrect config.
If there is a request that the ks-core would run in another namespace, I would change it.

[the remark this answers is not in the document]
RolandMa1986: @yuswift Updated with helm release namespace.

PR #3902, opened 2021-05-20
fix servicemesh traffic governance

zackzhangkai:

**What type of PR is this?**

Can not save configure for traffic governance, because of the change api of crd.

Just copy the generated crd from [istio api](https://github.com/istio/api/blob/master/kubernetes/customresourcedefinitions.gen.yaml). No need generate them by ourselves, for the different tools used.

### DIFF
--- a/config/crds/servicemesh.kubesphere.io_servicepolicies.yaml
+++ b/config/crds/servicemesh.kubesphere.io_servicepolicies.yaml
@@ -69,933 +69,1131 @@ spec:
                     description: Metadata of the virtual services created from this template
                     type: object
                   spec:
-                    description: Spec indicates the behavior of a destination rule.
+                    description: 'Configuration affecting load balancing, outlier detection,
+                      etc. See more details at: https://istio.io/docs/reference/config/networking/destination-rule.html'
                     properties:
-                      export_to:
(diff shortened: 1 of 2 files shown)


yuswift:
@zackzhangkai could we open a pr for ks-installer in the meanwhile?

zackzhangkai:
> @zackzhangkai could we open a pr for ks-installer in the meanwhile?

@yuswift Yes, ks-installer has changed already.

yuswift:
> > @zackzhangkai could we open a pr for ks-installer in the meanwhile?
> 
> @yuswift Yes, ks-installer has changed already.

Can we find a way to avoid generating those two files? If not, we will still generate them when executing `make manifests`

zackzhangkai:
@yuswift In fact, we still use controller-gen to generate crd for servciesmesh partly, for the places where has no virtualservice and destinationrule objects. Though it has some errors, we use some of it. And replace the virtualservice and destinationrule from the istio officially manually. 

Do you have some better ideas how to make a balance between these?

yuswift:
> @yuswift In fact, we still use controller-gen to generate crd for servciesmesh partly, for the places where have no virtualservice and destinationrule objects. Though it has some errors, we use some of it. And replace the virtualservice and destinationrule from the istio officially manually.
> 
> Do you have some better ideas how to make a balance between these?

I think it's too complicated... Will the crd files change in the future? If so, can we write a script for them?

zackzhangkai:
@yuswift  The crd is defined by us, so if we changed the type, the manifest should be updated, but it will not changed as usually. When we execute `make manifest`,  we specified the path `paths=./pkg/apis/...`. Do you have some mind about how to exclude the servicemesh types?

yuswift:
> @yuswift The crd is defined by us, so if we changed the type, the manifest should be updated, but it will not changed as usually. When we execute `make manifest`, we specified the path `paths=./pkg/apis/...`. Do you have some mind about how to exclude the servicemesh types?

#3919 has excluded service mesh crd.

Issue #3912, opened 2021-05-25
How to run kubebuilder to regenerate yaml file


regardfs:
**What's your question**
I had changed the PipelineSpec by adding GitlabCiPipeline as
![image](https://user-images.githubusercontent.com/8295227/119502544-9d81a800-bd9c-11eb-8892-74a314950454.png)
but how should i run kubebuilder or other cmd to take this change make effort，and generate or update yaml file，and apply to k8s

**Environment: OS & Hardware Information**

> Important: You must describe your environment clearly, e.g. VMware or Bare Metal, CentOS 7.5, 8 C / 16 G (If you install on Linux), Or Kubernetes v1.16 (If you install on K8s).
Kubernetes v1.19
Linux Centos7.8

**Error logs or message (Attach logs or screenshot)**

**Installer Version**

> v3.1.0

regardfs:
when I run make manifests，it failed with errors as follows：
```
zhangkekedeMacBook-Pro:kubesphere zhangkeke$ make manifests
go run ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go object:headerFile=./hack/boilerplate.go.txt paths=./pkg/apis/... rbac:roleName=controller-perms "crd:trivialVersions=true" output:crd:artifacts:config=config/crds
/Users/zhangkeke/go/pkg/mod/istio.io/api@v0.0.0-20201113182140-d4b7e3fc2b44/networking/v1alpha3/virtual_service.pb.go:2437:2: encountered struct field "MatchType" without JSON tag in type "StringMatch"
/Users/zhangkeke/go/pkg/mod/istio.io/api@v0.0.0-20201113182140-d4b7e3fc2b44/networking/v1alpha3/virtual_service.pb.go:2976:2: encountered struct field "HttpDelayType" without JSON tag in type "HTTPFaultInjection_Delay"
/Users/zhangkeke/go/pkg/mod/istio.io/api@v0.0.0-20201113182140-d4b7e3fc2b44/networking/v1alpha3/virtual_service.pb.go:3293:23: unsupported type "float64"
/Users/zhangkeke/go/pkg/mod/istio.io/api@v0.0.0-20201113182140-d4b7e3fc2b44/networking/v1alpha3/virtual_service.pb.go:3136:2: encountered struct field "ErrorType" without JSON tag in type "HTTPFaultInjection_Abort"
[...]

Issue #3830, opened 2021-04-29
Split KubeSphere api to it's own repository


RolandMa1986:
**What's it about?**
Split KubeSphere api to its own repository. 

**What's the reason why we need it?**
For the third part of applications that need to interact with KubeSphere APIs, The API definition often needs to be referred to. So we need to create a separate repo for it. 

**Area Suggestion**
/area api
/area apiserver
/kind feature

RolandMa1986:
/assign

@wanjunlei, The APIs would be split into a new repo. So you don't have to import the kubesphere/kubesphere anymore.

PR #3919, opened 2021-05-27
change controller-gen crds source path

RolandMa1986:
**What type of PR is this?**
/kind bug

1. CRD can't be generated by run the `make manifest` command when verifying #3912 in the master branch.
2. Generate specific CRDs by folder.
3. Clean out unnecessary CRDs. 

**Root Cause:**
Since we moved APIs definitions from the `pkg/apis` package to the `kubesphere.io/api` module, `make manifest` no longer works with old path "paths=./pkg/apis/...".
Correct the path with "paths=kubesphere.io/api/...".

Fixes ##3912
Related #3830


yuswift:
duplicate #3848 
we need to exclude service mesh dir to avoid generating crd files.

RolandMa1986:
Ok, seems there are still some issues need to be fixed. To resolve the case above, we need to exclude the service mesh CRDs.
CRDs generate strategy needs to be changed. Before that, I'd like to discuss which CRD should be put into the "config/crds" folder?

Currently, most of the CRDs that be used by KubeSphere were put into the "config/crds"  folder, for example:
1. Third-party components crds like `istio-crds.yaml`, `snapshot.storage.k8s.io_volumesnapshot.yaml`.  They were copied into the folder when related to feature check-in.
2. Third-party go API references and generated by `make manifest`, like `crd.projectcalico.org_blockaffinities.yaml`. Also include Auditing, monitoring, and Notification.
3. Federated resources, the type defined in the API package. 
4. Other API packages defined by KubeSphere.

Oboiversly case 1 and case 2 are not necessary. Those crds are managed by their own Helm chart or YAML when installing those components. For case 3, those federated crds were actually not be used, because we typically use the "kubefedctl" command to generate them instead of "make manifest".   I think that CRDs in cases 1 to 3 be should be removed from the folder. Firstly, we don't use them at all, compare with the ks-installer, ks-core only maintains CRDs defined by KS. Secondly, those CRD would be out of date when those components update to another version.

The "config/crds" folder should only contain CRDs that be defined and used by KubeSphere controllers. **So I propose to generated CRDs for each specific folder instead of all folders under "./apis/...".** 

Special cases like Service Mesh should be excluded, developers should maintain the CRD by themself.

/cc @zackzhangkai @yuswift @zryfish

yuswift:
Agree with you. But for snapshot, maybe we need to keep the crd files cause if we remove them, the controller manager will output some error logs.

yuswift:
Please make sure the files removed exist in their own helm charts.

RolandMa1986:
> Agree with you. But for snapshot, maybe we need to keep the crd files cause if we remove them, the controller manager will output some error logs.

I don't think it's necessary. We can warn users that the snapshot was not installed on the cluster, or the CSI doesn't support the snapshot.

yuswift:
> > Agree with you. But for snapshot, maybe we need to keep the crd files cause if we remove them, the controller manager will output some error logs.
> 
> I don't think it's necessary. We can warn users that the snapshot was not installed on the cluster, or the CSI doesn't support the snapshot.

Is snapshot crd necessary for ks-controller-manager? If not, I think we can remove it from ks-core. BTW, how do we install snapshot crd after we have installed ks-core with helm? @RolandMa1986

stoneshi-yunify:
> > > Agree with you. But for snapshot, maybe we need to keep the crd files cause if we remove them, the controller manager will output some error logs.
> > 
> > 
> > I don't think it's necessary. We can warn users that the snapshot was not installed on the cluster, or the CSI doesn't support the snapshot.
> 
> @stoneshi-yunify Is snapshot crd necessary for ks-controller-manager? If not, I think we can remove it from ks-core. BTW, how do we install snapshot crd after we have installed ks-core with helm? @RolandMa1986

No, snapshot crd is just a crd like any other crds. It's absolute unnecessary for ks-core. User can install the crd the snapshot-controller anytime. Just need to ensure ks won't crash without them.

RolandMa1986:
> Please make sure the files removed exist in their own helm charts.
All the CRDs that had been deleted can be found in the following folders.
https://github.com/kubesphere/ks-installer/tree/master/roles/common/files/snapshot-controller
https://github.com/kubesphere/ks-installer/tree/master/roles/ks-monitor/files/monitoring-dashboard
https://github.com/kubesphere/ks-installer/tree/master/roles/ks-monitor/files/notification-manager
https://github.com/kubesphere/ks-installer/tree/master/roles/ks-auditing/files/kube-auditing
https://github.com/kubesphere/ks-installer/tree/master/roles/ks-istio/files/istio
https://github.com/kubesphere/ks-installer/tree/master/roles/ks-multicluster/files/kubefed/federatedcrds

wansir:
I think you should split the unit test change to another PR, and squash commits.  CRD generation rules can be supplemented by others.

RolandMa1986:
> I think you should split the unit test change to another PR, and squash commits. CRD generation rules can be supplemented by others.

Those 3 UT test would be failed, when removed the Dashboard and Notification's CRDs.

PR #3931, opened 2021-06-02
Feature: support installing kubesphere-core with helm

yuswift:
**What type of PR is this?**
/kind feature

Support installing kubesphere-core with helm.
Since #3896 has created helm chart for ks-core, we just need to create `kubesphere-controls-system` namespace and some role templates for iam, then we can install ks-core with `helm`, which can reduce the complexity of installation.
To install ks-core with helm, just run 
```
make helm-deploy
```
To uninstall it, run
```
make helm-uninstall
```

see #3847 

I'm not sure if we need to update README to let kubesphere users know how to install ks-core with helm. if so, I will ask sig-docs to do that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,11 +98,19 @@ docker-build-no-test: ks-apiserver ks-controller-manager
 	hack/docker_build.sh
 
 helm-package:
-	ls config/crds/ | grep -v types.kubefed.io | xargs -i cp -r config/crds/{} config/ks-core/crds/
+	ls config/crds/ | xargs -i cp -r config/crds/{} config/ks-core/crds/
 	helm package config/ks-core --app-version=v3.1.0 --version=0.1.0 -d ./bin
 
 helm-deploy:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -102,7 +102,15 @@ helm-package:
 	helm package config/ks-core --app-version=v3.1.0 --version=0.1.0 -d ./bin
 
 helm-deploy:
+	ls config/crds/ | grep -v types.kubefed.io | xargs -i cp -r config/crds/{} config/ks-core/crds/
```
RolandMa1986, line 105: `grep -v types.kubefed.io` can be removed now, since we have deleted all the CRDs from config/crds/ folder.

yuswift: Ok. I will also remove it from `make package`.

yuswift:
/retitle Feature: support installing kubesphere-core with helm

yuswift:
@zryfish 
updated the pr description, hope it is useful :)

PR #3946, opened 2021-06-09
Moving to a fork for jwt-go

RolandMa1986:

**What type of PR is this?**
/kind cleanup

Use form3tech-oss/jwt-go to replace the dgrijalva/jwt-go package.

github.com/dgrijalva/jwt-go v3.2.0 is known to have a CVE-2020-26160 with severity rating as high. kubernetes is going to replace it with form3tech-oss/jwt-go, ref kubernetes/kubernetes#100401

The vulnerable function verifyAudience is not used in KubeSphere too, but it better to move form3tech-oss/jwt-go to prevent the false alarm.

Fixes #


zryfish:
Do we need to port this to release-3.1?

RolandMa1986:
I think it's not necessary. We won't be impacted by this issue since we don't call **VerifyAudience** in our code. 
/cc @wansir

Issue #3843, opened 2021-05-06
Simplify provisioner and storageclass capabilities


stoneshi-yunify:
**Describe the Bug**
Currently ks has two crd to define the provisioner and storageclass capabilities:
```
[root@test-ks ~]# kubectl get crd | egrep capab
provisionercapabilities.storage.kubesphere.io         2021-04-28T08:17:53Z
storageclasscapabilities.storage.kubesphere.io        2021-04-28T08:17:54Z
```

And we have code to automatically detect the provisioner capabilities, however, it does not work for most provisioners. 

The proposed change is:
- simplify the  provisioner and storageclass capabilities, I don't think we need the storageclass capabilities, and I don't think we need export so many capabilities as we do now. 
- remove the automatic detection of provisioner capabilities, it does not work well.
- work with ks-console developer to change the frontend logic. The provisioner capabilities should be provided by user, except the qingstor-csi and qingcloud-csi which can be built-in with ks.

**Versions Used**
KubeSphere:
Kubernetes: (If KubeSphere installer used, you can skip this)

**Environment**
How many nodes and their hardware configuration:

For example: CentOS 7.5 / 3 masters:  8cpu/8g; 3 nodes: 8cpu/16g
(and other info are welcomed to help us debugging)

**How To Reproduce**

**Expected behavior**

xyz-li:
https://kubesphere.com.cn/forum/d/4505-31-ks-controller-manager

stoneshi-yunify:
for 3.1.1, we need to remove the auto detect part, but keep the storageclasscapabilities and provisionercapabilities, so ks-console does not need any change.

the ks-console part change will be tracked in https://github.com/kubesphere/kubesphere/issues/3921

PR #3947, opened 2021-06-09
remove storage capability auto detection

dkeven:
Signed-off-by: dkven <dkvvven@gmail.com>

**What type of PR is this?**
/kind bug

The current auto-detection of StorageClasses' capabilities, by probing its CSIDriver's gRPC socket, does not work most of the time.

And when probe fails, it outputs a huge amount of error logs repeatedly, making things even worse.

To be backward compatible, and to be compatible with current ks-console logic, this PR removes the probing logic from the storageCapabilityController, and uses the preset ProvisionerCapabilities as the value source of StorageClassCapabilities instead.

That is, when a StorageClass/ProvisionerCapability Object gets created/updated, the controller will find the corresponding PC, and, based on its values, creates(if not exists)/updates the relevant StorageClassCapability Object(s), as well as updates the SCs' annotations or creates VolumeSnapshotClass Object(s), if necessary.

Therefore, with the new version of this controller, users need to mannually create ProvisionerCapability or StorageClassCapability Objects for their custom StorageClasses. (like they do now)

Also, this PR adds a preset ProvisionerCapability yaml file for the qingcloud-csi and qingstor-csi driver.

This is a minor, backward compatible, temporary fix. In the future, we will change ProvisionerCapability, and discard StorageClassCapability entirely, to provide a simple version of storage capability, which requires co-working with ks-console.

Fixes (https://github.com/kubesphere/kubesphere/issues/3843)


config/default/provisonercapability/disk-csi-qingcloud-com.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: storage.kubesphere.io/v1alpha1
```
stoneshi-yunify, line 1: I suspect this file and the below neonsan-csi-qingstor-com.yaml will be used by ks-installer to preinstall the provisonercapabilities, if so, be sure to submit related PR to ks-installer or talk with @pixiake

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -141,12 +126,20 @@ func NewController(
...
+	// but don't overdo by cascade deleting the SCCs when a PC is deleted
+	// since the role of PCs is more like a template rather than owner to SCCs
+
+	// This is a backward compatible fix to remove the useless auto detection of SCCs
+	// in the future, we will only keep ProvisionerCapability and remove the StorageClassCapability CRD entirely
+	ksStorageInformer.ProvisionerCapabilities().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: controller.handleProvisionerCapability,
 		UpdateFunc: func(oldObj, newObj interface{}) {
-			return
+			controller.handleProvisionerCapability(newObj)
```
stoneshi-yunify, line 140: compare resourceversion of old and new object before this?

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -141,12 +126,20 @@ func NewController(
...
 		},
-		DeleteFunc: controller.handlerCSIDriver,
+		DeleteFunc: func(obj interface{}) {},
```
stoneshi-yunify, line 142: can this DeleteFunc be leaved unset?

config/default/provisonercapability/disk-csi-qingcloud-com.yaml
dkeven: Yes, that's true, I'll do that when this PR is approved

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -278,14 +270,16 @@ func (c *StorageCapabilityController) syncHandler(key string) error {
 	if err != nil {
 		return err
 	}
-	// No capability because csi-plugin not installed
+	// Can't get StorageCapability's ProvisionerCapability
```
stoneshi-yunify, line 273: Should be `StorageClass's ProvisionerCapability`?

stoneshi-yunify (changes requested):
Review done. Main logic is okay to me, thanks. Though need to address my questions.

dkeven (commented):
> Review done. Main logic is okay to me, thanks. Though need to address my questions.

Thanks for the review & suggestions, I've made the changes accordingly.

stoneshi-yunify (commented):
/lgtm

PR #3961, opened 2021-06-10
Make ks-apiserver be easier to run locally with kube config

LinuxSuRen:
**What type of PR is this?**
/kind feature

As I mentioned in the title, users can run ks-apiserver locally.

None.

None.

None.

### DIFF
--- a/pkg/simple/client/k8s/options.go
+++ b/pkg/simple/client/k8s/options.go
@@ -18,6 +18,7 @@ package k8s
 
 import (
 	"os"
+	"os/user"
 	"path"
 
 	"k8s.io/client-go/util/homedir"
@@ -54,8 +55,16 @@ func NewKubernetesOptions() (option *KubernetesOptions) {
 	}
 
 	// make it be easier for those who wants to run api-server locally
-	userHomeConfig := path.Join(homedir.HomeDir(), ".kube/config")
-	if _, err := os.Stat(userHomeConfig); !os.IsNotExist(err) {
+	homePath := homedir.HomeDir()
+	if homePath == "" {
+		// try os/user.HomeDir when $HOME is unset.
+		if u, err := user.Current(); err == nil {
+			homePath = u.HomeDir
+		}
+	}
+
+	userHomeConfig := path.Join(homePath, ".kube/config")
+	if _, err := os.Stat(userHomeConfig); err == nil {
 		option.KubeConfig = userHomeConfig
 	}
 	return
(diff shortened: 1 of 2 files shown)


wansir:
`clientcmd.DeferredLoadingClientConfig` may be a good choice.

https://github.com/kubesphere/kubesphere/blob/master/vendor/sigs.k8s.io/controller-runtime/pkg/client/config/config.go#L120-L137

pkg/simple/client/k8s/options.go
```diff
@@ -44,12 +47,18 @@ type KubernetesOptions struct {
...
-		Burst:      1e6,
+func NewKubernetesOptions() (option *KubernetesOptions) {
+	option = &KubernetesOptions{
+		QPS:   1e6,
+		Burst: 1e6,
+	}
+
+	// make it be easier for those who wants to run api-server locally
+	userHomeConfig := path.Join(homedir.HomeDir(), ".kube/config")
+	if _, err := os.Stat(userHomeConfig); !os.IsNotExist(err) {
```
yuswift, line 58: I'd prefer to judge if err==nil. Why do you only handle the `NotExist` err?

LinuxSuRen: Excellent suggestion!

LinuxSuRen:
> `clientcmd.DeferredLoadingClientConfig` may be a good choice.

Thanks for your suggestion. Here we only need the KubeConfig file path. I learned from your suggested code lines.

pkg/simple/client/k8s/options.go
```diff
@@ -44,12 +48,26 @@ type KubernetesOptions struct {
...
+		if u, err := user.Current(); err == nil {
+			homePath = u.HomeDir
+		}
+	}
+
+	userHomeConfig := path.Join(homePath, ".kube/config")
+	if _, err := os.Stat(userHomeConfig); err == nil {
+		option.KubeConfig = userHomeConfig
 	}
+	return
```
JohnNiang, line 70: IMO, the `return` line is unnecessary for here.

LinuxSuRen: image

JohnNiang: I have no problem as https://github.com/golang/go/issues/12168

LinuxSuRen: > I have no problem as [golang/go#12168](https://github.com/golang/go/issues/12168)

I didn't get your point. Are you trying to say that there's no error in your IDE?

JohnNiang: I mean that I've seen the [declined proposal](https://github.com/golang/go/issues/12168) about return statement at the very end of a function, so I have no problem of this return statement.

PR #3965, opened 2021-06-15
Improve the s3 uploader for better performance.

dkkb:
**What type of PR is this?**

/kind feature

For better performance, we can increase the concurrency to upload large objects.
We need to use the length of data to calculate concurrency, so I change the `Upload` interface method,
replace the `io.Reader` with `[]byte` to get the length.

As we can see, if the file size is greater than 100MB,  upload speed is much faster than before.

Fixes #

```
```

```docs

```


xyz-li:
@dkkb 
I think you need change this too.
https://github.com/kubesphere/kubesphere/blob/master/pkg/models/devops/s2ibinary_handler.go#L107
```
	err = s.s3Client.Upload(fmt.Sprintf("%s-%s", namespace, name), copy.Spec.FileName, binFile)
```

zheng1:
@dkkb Thank you for your contribution

/approve

PR #3966, opened 2021-06-15
Fix failed to assign an administrator to a existing project

### LINKED ISSUE #3819: Failed to allocate workspace and manager to a project
**Describe the Bug**

image

image

image

**Versions Used**
KubeSphere: nightly-20210427

/kind bug

wansir:

**What type of PR is this?**

/kind bug

We can assign an administrator to a project on the cluster management page, when `kubesphere.io/creator` annotation changes we need to create the corresponding role binding. 

Fixes #3819


RolandMa1986:
@wansir It looks like the behavior is changed by the commit. Previously the `kubesphere.io/creator` annotation persistent the user who created the project. So the ** creator** should never be changed. But it looks like an **owner** now. Please confirm whether it's the desired behavior change.

wansir:
> @wansir It looks like the behavior is changed by the commit. Previously the `kubesphere.io/creator` annotation persistent the user who created the project. So the ** creator** should never be changed. But it looks like an **owner** now. Please confirm whether it's the desired behavior change.

I agree with you, the creator and the administrator are not the same concepts, should create a role binding instead of modifying the creator.

PR #4034, opened 2021-07-01
Fix failed to push docker images with a specified tag

wansir:

### What type of PR is this?

/kind flake

Fix failed to push image with a wrong image tag. Whether the tag or a branch changes, the image should be pushed.

error logs: https://github.com/kubesphere/kubesphere/runs/2950932182?check_suite_focus=true#step:11:100

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,4 +64,4 @@ jobs:
         if: github.event_name == 'push'
         run: |
           echo ${{ secrets.DOCKER_PASSWORD }} | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
-          REPO=kubespheredev TAG=${GITHUB_REF#refs/heads/} make container-push
+          REPO=kubespheredev TAG=${GITHUB_REF#refs/*/} make container-push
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -64,4 +64,4 @@ jobs:
         if: github.event_name == 'push'
         run: |
           echo ${{ secrets.DOCKER_PASSWORD }} | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
-          REPO=kubespheredev TAG=${GITHUB_REF#refs/heads/} make container-push
+          REPO=kubespheredev TAG=${GITHUB_REF#refs/*/} make container-push
```
yuswift, line 67: please modify https://github.com/kubesphere/kubesphere/blob/master/.github/workflows/build-multiarch.yaml too and cherry-pick this pr

yuswift: BTW, what's the value of `TAG ` if we run the workflow in tag `v3.1.1-rc.1`?

zheng1:
```bash
GITHUB_REF='refs/tags/v3.1.1-rc.1'
echo ${GITHUB_REF#refs/*/}
```

wansir: > please modify https://github.com/kubesphere/kubesphere/blob/master/.github/workflows/build-multiarch.yaml too and cherry-pick this pr

done

PR #4103, opened 2021-07-29
feat: load configuration from Environment variables

RolandMa1986:

### What type of PR is this?
/kind feature
/kind flake

This PR adds support reading application configuration from environment variables. For example, the Redis password can be either configurated in CLI by the `--redis-password` flag, YAML file, or environment variable "KUBESPHERE_REDIS_PASSWORD".

For security reasons, users should not save the passwords in the ConfigMap or Deployment's CLI flags.  In Kubernetes practice, We should save security information into a Secret. Then either mount the Secret as a Volume or Environment variable.  
In the feature, all the security configurations used by ks-apiserver would be saved into Secret and mount as Environment variables.  

Fixes #

```
```

```release-note
NONE
```

```docs

```

### DIFF
--- a/config/ks-core/templates/ks-apiserver.yml
+++ b/config/ks-core/templates/ks-apiserver.yml
@@ -49,6 +49,10 @@ spec:
           name: kubesphere-config
         - mountPath: /etc/localtime
           name: host-time
+        env:
+        {{- if .Values.env }}
+        {{- toYaml .Values.env | nindent 8 }}
+        {{- end }}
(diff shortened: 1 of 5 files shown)


yuswift:
I'm curious where do we set env(e.g. in the ks-installer or somewhere else)?

RolandMa1986:
> I'm curious where do we set env(e.g. in the ks-installer or somewhere else)?
The env will be set in the ks-installer eventually.

PR #4106, opened 2021-08-02
Move the codebase of DevOps into a separate project

### LINKED ISSUE #3962: Make DevOps as a separate Application from KubeSphere core
## Backgroud

Currently, the DevOps backend is part of KubeSphere Core. It's impossible to upgrade the DevOps component independently if a specific version has a crucial bug exist. Also, users cannot install DevOps as an application on Kubernetes.

## Proposal

I'd like to suggest that make DevOps be a separate application. Please see also the following points:

* Install DevOps application as a Helm charts
* Has no dependency on KubeSphere
  *  Only as a regular k8s application
  * Advanced features could be an enhancement in KubeSphere

What things do we need to do?

* Move the DevOps controller-manager and APIServer into a separate repo
* Move the DevOps frontend into a separate repo
* Has a Helm chart for the DevOps application

/area devops
/kind proposal
/stage alpha

LinuxSuRen:
### What type of PR is this?

/kind cleanup

Fixes #3962

Considering there're too many files changed in this PR, please read the following suggestion.

Mostly, I delete the DevOps-related code lines. And move it to [a new project](https://github.com/kubesphere-sigs/ks-devops). About this part, you don't need to take too much time to review.

In order to make it be compatible with the DevOps API. I created [a proxy](pkg/apiserver/filters/devops.go) for the DevOps-related APIs. This might be important.

Why did I create the new project understand the organization kubesphere-sigs? It lasts a long time to make the DevOps be available. Before it comes stable, I prefer to put it to SIG organization. It can avoid making users feel confused about the quality of the project.

```release-note
Move the codebase of DevOps into a separate project
```

```docs
None.
```


JohnNiang:
hi @LinuxSuRen , there may be some remaining files related devops to be removed. please feel free to correct me if i miss anything.

image

pkg/apiserver/authorization/rbac/rbac.go
```diff
@@ -250,11 +250,11 @@ func (r *RBACAuthorizer) visitRulesFor(requestAttributes authorizer.Attributes,
 				}
 			}
 		} else if requestAttributes.GetResourceScope() == request.DevOpsScope {
-			if workspace, err = r.am.GetDevOpsControlledWorkspace(requestAttributes.GetDevOps()); err != nil {
-				if !visitor(nil, "", nil, err) {
-					return
-				}
-			}
+			//if workspace, err = r.am.GetDevOpsControlledWorkspace(requestAttributes.GetDevOps()); err != nil {
```
zryfish, line 253: Why do we keep those unused lines?

pkg/apiserver/filters/devops.go
```diff
@@ -0,0 +1,78 @@
...
+	"k8s.io/apimachinery/pkg/util/proxy"
+	"k8s.io/apiserver/pkg/endpoints/handlers/responsewriters"
+	"k8s.io/klog"
+
+	"kubesphere.io/kubesphere/pkg/apiserver/request"
+	"kubesphere.io/kubesphere/pkg/server/errors"
+)
+
+// WithDevOpsAPIServer proxy request to DevOps service if requests path has the APIGroup with devops.kubesphere.io
+func WithDevOpsAPIServer(handler http.Handler, config *jenkins.Options, failed proxy.ErrorResponder) http.Handler {
```
zryfish, line 36: For those business specific logic, it's better to use proxy, for example https://github.com/kubesphere/kubesphere/blob/master/pkg/kapis/notification/v1/register.go

pkg/simple/client/devops/jenkins/options.go
```diff
@@ -25,10 +25,14 @@ import (
...
-	Username       string `json:",omitempty" yaml:"username" description:"Jenkins admin username"`
-	Password       string `json:",omitempty" yaml:"password" description:"Jenkins admin password"`
-	MaxConnections int    `json:"maxConnections,omitempty" yaml:"maxConnections" description:"Maximum connections allowed to connect to Jenkins"`
+	Host                       string `json:",omitempty" yaml:"host" description:"Jenkins service host address"`
+	Username                   string `json:",omitempty" yaml:"username" description:"Jenkins admin username"`
+	Password                   string `json:",omitempty" yaml:"password" description:"Jenkins admin password"`
+	MaxConnections             int    `json:"maxConnections,omitempty" yaml:"maxConnections" description:"Maximum connections allowed to connect to Jenkins"`
+	DevOpsServiceAddress       string `json:"devopsServiceAddress" yaml:"devopsServiceAddress" description:"The separate DevOps Service address"`
+	DevOpsPluginServiceAddress string `json:"devopsPluginServiceAddress" yaml:"devopsPluginServiceAddress" description:"The KubeSphere DevOps plugin Service address"`
```
zryfish, line 33: What's the difference between `DevopsServiceAddress` and `DevOpsPluginServiceAddress`?

pkg/simple/client/devops/jenkins/options.go
```diff
@@ -25,10 +25,14 @@ import (
...
+	// K8sBearerToken does not take input from outside
+	K8sBearerToken string
```
zryfish, line 35: Why do we need `K8sBearerToken`?

pkg/simple/client/devops/jenkins/options.go
LinuxSuRen: `DevopsServiceAddress` is the address of ks-devops-apiserver from https://github.com/kubesphere-sigs/ks-devops/

`DevOpsPluginServiceAddress` is the address of ks-devops-plugin from https://github.com/kubesphere-sigs/ks-devops-plugin

Of course, this is just a temporary solution. We will remove them once the real plugin mechanism of KubeSphere is ready.

pkg/models/iam/am/am.go
```diff
@@ -106,13 +107,12 @@ type amOperator struct {
...
 	clusterRoleGetter          resourcev1alpha3.Interface
 	roleGetter                 resourcev1alpha3.Interface
-	devopsProjectLister        devopslisters.DevOpsProjectLister
 	namespaceLister            listersv1.NamespaceLister
 	ksclient                   kubesphere.Interface
 	k8sclient                  kubernetes.Interface
 }
 
-func NewReadOnlyOperator(factory informers.InformerFactory) AccessManagementInterface {
+func NewReadOnlyOperator(factory informers.InformerFactory, config *config.Config) AccessManagementInterface {
```
wansir, line 115: `config *config.Config` is not used

pkg/simple/client/devops/jenkins/options.go
LinuxSuRen: The purpose of the variable `K8sBearerToken`. https://github.com/kubesphere/kubesphere/pull/4106/files#diff-1f0f138b59d0cd61e116c34a0f0fdda8c23c48cd517d1f0c9946ea6d4d1d3930R112

This also is a temporary solution to pass the token.

pkg/apiserver/authorization/rbac/rbac.go
LinuxSuRen: Currently, we didn't move this part to any projects. And there's no good solution to solve it. In order to avoid forgetting it, I recommend leaving it until we have a solution for it.

wansir (changes requested):

pkg/models/iam/am/am.go
LinuxSuRen: Good catch!

JohnNiang:
@LinuxSuRen , i have a question for you: should we remove the DevOps APIs and generated DevOps clients?

pkg/apiserver/filters/devops.go
LinuxSuRen: It's a really good example. But we will replace the proxy with the KubeSphere plugin mechanism which is under the testing stage. I believe it will be very soon.

/cc @wansir

LinuxSuRen:
> @LinuxSuRen , i have a question for you: because we won't maintain those DevOps APIs in here in the future, should we remove the DevOps APIs and generated DevOps clients?

Yes. I prefer to remove all of them.

pkg/kapis/iam/v1alpha2/handler.go
```diff
@@ -1099,10 +1099,10 @@ func (h *iamHandler) DescribeNamespaceRole(request *restful.Request, response *r
 
 // resolve the namespace which controlled by the devops project
 func (h *iamHandler) resolveNamespace(namespace string, devops string) (string, error) {
-	if devops == "" {
-		return namespace, nil
-	}
-	return h.am.GetDevOpsRelatedNamespace(devops)
+	//if devops == "" {
+	return namespace, nil
```
JohnNiang, line 1103: Please pre-check if devops is empty, then return devops first if devops is not empty.

This fix could fix <https://github.com/kubesphere-sigs/ks-devops/issues/91>.

Test image: `johnniang/ks-apiserver:remove-devops-ctrl-iam`.

LinuxSuRen: Thanks for pointing this.

pkg/apiserver/filters/devops.go
wansir: Before that, we'd better follow [this example](https://github.com/kubesphere/kubesphere/blob/master/pkg/kapis/notification/v1/register.go.) register proxy API endpoint. The dynamic  API registration is a new feature of 4.0, which will develop in a new branch.

JohnNiang: > Before that, we'd better follow [this example](https://github.com/kubesphere/kubesphere/blob/master/pkg/kapis/notification/v1/register.go.) register proxy API endpoint. The dynamic API registration is a new feature of 4.0, which will develop in a new branch.

The broken link `this example` has one more period.

pkg/apiserver/filters/devops.go
```diff
@@ -0,0 +1,78 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		info, ok := request.RequestInfoFrom(req.Context())
+		if !ok {
+			err := errors.New("Unable to retrieve request info from request")
+			klog.Error(err)
+			responsewriters.InternalError(w, req, err)
+		}
+
+		if info.APIGroup == "devops.kubesphere.io" {
+			serverAsProxy(w, req, config.DevOpsServiceAddress, config.K8sBearerToken)
```
LinuxSuRen, line 55: About this part, it can adopt the following example.

https://github.com/kubesphere/kubesphere/blob/master/pkg/kapis/notification/v1/register.go

But others only have the partial DevOps APIs, I found out it cannot use that directly. For example, the APIs come from `tenant.kubesphere.io` and `resources.kubesphere.io`. Do you have any good ideas about this part?

So, I propose two potential solutions here:

* remain the directory: `/pkg/kapis/devops` which fully adopt the above example
  * pros: keep the DevOps API proxy be clean
  * cons: split the DevOps proxy code lines into serval places, increase the maintenance cost
* keep the filter and partially reuse [the generic proxy](https://github.com/kubesphere/kubesphere/blob/2aa55c5d6b14dbff458de3f257745abc3c0dcb0c/pkg/kapis/generic/generic.go#L95) 
  * pros: keep all the devops-related code lines together
  * cons: the code base is not beautiful

No matter which solution, all the DevOps proxy code lines will be removed.

/cc @wansir @zryfish

LinuxSuRen:
I'll close this PR. And replace it with #4129

PR #4108, opened 2021-08-02
Run makefile command fmt and goimports with test command

LinuxSuRen:
### What type of PR is this?

/kind cleanup

it is very easy to forget the format command after some code lines changed. but the comand is necessary for all contributors, it is not easy to forget

None existing issues.

```
None
```

```release-note
None
```

```docs
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,7 +133,7 @@ helm-uninstall:
 	kubectl delete -f https://raw.githubusercontent.com/kubesphere/ks-installer/master/roles/ks-core/prepare/files/ks-init/role-templates.yaml
 
 # Run tests
-test: vet
+test: vet fmt goimports
 	export KUBEBUILDER_CONTROLPLANE_START_TIMEOUT=2m; go test ./pkg/... ./cmd/... -covermode=atomic -coverprofile=coverage.txt
 
 .PHONY: clean


Makefile
```diff
@@ -133,7 +133,7 @@ helm-uninstall:
 	kubectl delete -f https://raw.githubusercontent.com/kubesphere/ks-installer/master/roles/ks-core/prepare/files/ks-init/role-templates.yaml
 
 # Run tests
-test: vet
+test: vet fmt goimports
```
zryfish, line 136: I have a second thought, we use `make test` in Prow, adding `fmt goimports` will slow our pipeline. We have presubmit checks to find those unformatted PRs. For developers, `make verify-all` before submitting is a good practice.

LinuxSuRen: `make verify-all` is a definitely good solution to do the check work. But this is my first time to know that makefile command. It's quite important that letting contributors know fewer things before they get started their contributions. How about the following solution:

```
test: vet verify-all
  xxx

test-without-verify: vet
  xxx
```

then let Prow runs the command `make test-without-verify`.

zryfish: I think it's better to make it more clear that contributors should run `make verify-all` before submitting in our development guide. Also, if they don't run `make verify-all`, our presubmit will complain and they will do it next time :)

LinuxSuRen: Ok than. But I still believe that the process should be as easy as possible. I'm going to close this PR until others want to simplify the process.

PR #4128, opened 2021-08-06
clean up app when workspace is deleted

### LINKED ISSUE #3915: After deleting workspace, the app templates  will not be deleted.
**Versions Used**
KubeSphere:  `kubesphere3.1`
Kubernetes:   `v1.20.4`

**Preset conditions**
There is a workspace named 'test-ws'

**How To Reproduce**
1. Go to workspace 'test-ws'
2. Create a template named 'app-test'
3. Delete workspace 'test-ws' 
4. Create a workspace named 'test-ws'
5. Go to workspace 'test-ws'
6. View App Templates

**Expected behavior**
There is no template in 'app templates' page

**Actual behavior**
Template 'app-test' in 'app templates' page

/kind bug
/milestone 3.1.1
/assign @xyz-li 
/priority medium

xyz-li:
### What type of PR is this?

/kind bug

Clean up app when workspace is delete.

When one workspace is being deleting, we can delete all the app which are not active or suspended of this workspace,
but when an app has been promoted to app store, we have to deal with it specially.
If we just delete that app, then this app will be deleted from app store too.
If we leave it alone, and user creates a workspace with the same name sometime, then this app will appear in this new workspace which confuses the user.
So we need to delete all the appversion which are not active or suspended first, then remove the workspace label from the app. And on the console of ks, we will show something like "(workspace deleted)" to user for this app.

Fixes #3915

```
```

```release-note
None

```

```docs

```

related issue https://github.com/kubesphere/console/issues/2145

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -41,7 +41,7 @@ const (
 	ChartApplicationIdLabelKey        = "application.kubesphere.io/app-id"
 	ChartApplicationVersionIdLabelKey = "application.kubesphere.io/app-version-id"
 	CategoryIdLabelKey                = "application.kubesphere.io/app-category-id"
-	DangingAppCleanupKey              = "application.kubesphere.io/app-cleanup"
+	DanglingAppCleanupKey             = "application.kubesphere.io/app-cleanup"
 	CreatorAnnotationKey              = "kubesphere.io/creator"
 	UsernameLabelKey                  = "kubesphere.io/username"
 	DevOpsProjectLabelKey             = "kubesphere.io/devopsproject"
@@ -71,8 +71,9 @@ const (
 	OpenpitrixAttachmentTag  = "Attachment"
 	OpenpitrixRepositoryTag  = "Repository"
 	OpenpitrixManagementTag  = "App Management"
-	CleanupDangingAppOngoing = "ongoing"
-	CleanupDangingAppDone    = "done"
+
+	CleanupDanglingAppOngoing = "ongoing"
+	CleanupDanglingAppDone    = "done"
 
 	DevOpsCredentialTag  = "DevOps Credential"
 	DevOpsPipelineTag    = "DevOps Pipeline"
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -41,6 +41,7 @@ const (
 	ChartApplicationIdLabelKey        = "application.kubesphere.io/app-id"
 	ChartApplicationVersionIdLabelKey = "application.kubesphere.io/app-version-id"
 	CategoryIdLabelKey                = "application.kubesphere.io/app-category-id"
+	DangingAppCleanupKey              = "application.kubesphere.io/app-cleanup"
 	CreatorAnnotationKey              = "kubesphere.io/creator"
 	UsernameLabelKey                  = "kubesphere.io/username"
 	DevOpsProjectLabelKey             = "kubesphere.io/devopsproject"
@@ -70,6 +71,8 @@ const (
 	OpenpitrixAttachmentTag  = "Attachment"
 	OpenpitrixRepositoryTag  = "Repository"
 	OpenpitrixManagementTag  = "App Management"
+	CleanupDangingAppOngoing = "ongoing"
+	CleanupDangingAppDone    = "done"
 
 	DevOpsCredentialTag  = "DevOps Credential"
 	DevOpsPipelineTag    = "DevOps Pipeline"
--- a/pkg/controller/openpitrix/helmapplication/helm_application_controller.go
+++ b/pkg/controller/openpitrix/helmapplication/helm_application_controller.go
@@ -22,6 +22,8 @@ import (
 	"strconv"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/labels"
+
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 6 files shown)


pkg/constants/constants.go
```diff
@@ -41,6 +41,7 @@ const (
 	ChartApplicationIdLabelKey        = "application.kubesphere.io/app-id"
 	ChartApplicationVersionIdLabelKey = "application.kubesphere.io/app-version-id"
 	CategoryIdLabelKey                = "application.kubesphere.io/app-category-id"
+	DangingAppCleanupKey              = "application.kubesphere.io/app-cleanup"
```
zryfish, line 44: What does `Danging` mean?

pkg/controller/openpitrix/helmapplication/helm_application_controller.go
```diff
@@ -181,3 +188,82 @@ func (r *ReconcileHelmApplication) SetupWithManager(mgr ctrl.Manager) error {
 func inAppStore(app *v1alpha1.HelmApplication) bool {
 	return strings.HasSuffix(app.Name, v1alpha1.HelmApplicationAppStoreSuffix)
 }
+
+// cleanupDangingApp delete the app when it is not active and not suspended,
```
zryfish, line 192: delete => deletes

pkg/controller/openpitrix/helmapplication/helm_application_controller.go
```diff
@@ -181,3 +188,82 @@ func (r *ReconcileHelmApplication) SetupWithManager(mgr ctrl.Manager) error {
...
+// sets the workspace label to empty and remove parts of the appversion when app state are active or suspended
+func (r *ReconcileHelmApplication) cleanupDangingApp(ctx context.Context, app *v1alpha1.HelmApplication) error {
+	if app.Annotations[constants.DangingAppCleanupKey] == constants.CleanupDangingAppOngoing {
```
zryfish, line 195: What if app.Annotations are nil?

pkg/controller/openpitrix/helmapplication/helm_application_controller.go
```diff
@@ -181,3 +188,82 @@ func (r *ReconcileHelmApplication) SetupWithManager(mgr ctrl.Manager) error {
...
+		// Just delete the app when the state is not active or not suspended.
+		if app.Status.State != v1alpha1.StateActive && app.Status.State != v1alpha1.StateSuspended {
```
zryfish, line 197: What if app.Status is nil?

zryfish:
I need more background information about this PR. The linked issue is not clear. Please add more info, and focus on **Why**.

pkg/controller/openpitrix/helmapplication/helm_application_controller.go
xyz-li: Status is a struct, not a pointer, it can not be nil.

pkg/constants/constants.go
xyz-li: Danging => Dangling

pkg/controller/openpitrix/helmapplication/helm_application_controller.go
xyz-li: A nil map behaves like an empty map when reading, but attempts to write to a nil map will cause a runtime panic.

xyz-li:
> I need more background information about this PR. The linked issue is not clear. Please add more info, and focus on **Why**.

When one workspace is being deleting, we can delete all the app which are not active or suspended of this workspace,
but when an app has been promoted to app store, we have to deal with it specially.
If we just delete that app, then this app will be deleted from app store too.
If we leave it alone, and user creates a workspace with the same name sometime, then this app will appear in this new workspace which confuses the user.
So we need to delete all the appversion which are not active or suspended first, then remove the workspace label from the app. And on the console of ks, we will show something like "(workspace deleted)" to user for this app.

zryfish:
Why only delete inactive and suspended apps?

xyz-li:
> Why only delete inactive and suspended apps?

Active apps are promoted to app store, then these apps are out of the scope of workspace. The user should delete them in app store.

zryfish:
Why don't we use owner reference? An app is owned by workspace, if an app is published to the app store, then we remove its owner reference. In this case, inactive or suspended apps will be deleted together with workspace by K8s GC.

xyz-li:
If we add owner reference to metadata, then we don't need to delete helm app specifically. But we need remove workspace label from helm app and add annotations to the app in app store. And we  also need to remove owner reference from metadata.

zryfish:
> But we need remove workspace label from helm app and add annotations to the app in app store.

What do you mean remove labels? Can you give some examples?

xyz-li:
```
apiVersion: application.kubesphere.io/v1alpha1
kind: HelmApplication
metadata:
  labels:
    application.kubesphere.io/app-category-id: ctg-yo579pmoopvq42
    kubesphere.io/workspace:  test-workspace  # the workspace label
  name: app-znwn5xp2z89q7o
  resourceVersion: "5364785"
  uid: abb3ca40-ab65-416c-8d92-fc78140c36df
spec: {}
```
All the apps have a label `kubesphere.io/workspace` which distinguishes the workspace of an app. If the workspace `test-workspace` is deleted, we must set the the label to empty. Or the user may create the workspace `test-workspace` at sometime, then this app will appear this new workspace suddenly.

Issue #4075, opened 2021-07-20
Manage storage capabilities in console


stoneshi-yunify:
**What's it about?**
Add new feature to manage storage capabilities in console

/area storage
/milestone v3.2
/kind feature-request

f10atin9:
I'll remove useless capabilityCRD and then update capability controller.The updated controller can watch the changes of storageClass and CSIDriver, and when they change, it can be judged whether it should be annotated. If the true controller will annotate "storageclass.kubesphere.io/support-snapshot: true" and "storageclass.kubesphere.io/support-clone: true". The console will judge whether the capability button can be clicked based on this value.

f10atin9:
The judgment process of the controller is as follows: 
When storageClass is added or updated, controller gets its Provisioner and matches it with CSIDriver. If the match is successful, it will be annotated, otherwise no operation. 
When CSIDriver is added or deleted, get related storageClasses, add or delete annotations for them.
In addition, if k8s version <1.17, cluster does not support snapshot, it will be annotated "storageclass.kubesphere.io/support-snapshot: false".
And the console will add page management capability. If the user closes the capability, the false annotation will be added at this time.

stoneshi-yunify:
I prefer the allow-xxx vs support -xxx annotation, just like the allowExpansion. The allow-xxx implies both "it is supported" and "I allow it".

For example, if one storage supports snapshot, but the admin don't want sc A to create snapshot, in this case, the allow-snapshot annotation is better, while the support-xxx is a little misleading.

f10atin9:
> I prefer the allow-xxx vs support -xxx annotation, just like the allowExpansion. The allow-xxx implies both "it is supported" and "I allow it".
> 
> For example, if one storage supports snapshot, but the admin don't want sc A to create snapshot, in this case, the allow-snapshot annotation is better, while the support-xxx is a little misleading.

Okay, i will update it. Thanks for your suggestion.

PR #4163, opened 2021-08-19
remove capability CRDs and update controller

f10atin9:
Signed-off-by: f10atin9 <f10atin9@kubesphere.io>

### What type of PR is this?

/kind feature
/area storage
Add new feature to manage storage capabilities in console

Fixes #
https://github.com/kubesphere/kubesphere/issues/4075
```
```
/cc @stoneshi-yunify @dkeven @kubesphere/sig-storage 

```release-note

```

```docs

```


pkg/controller/storage/capability/capability_controller.go
```diff
@@ -47,46 +47,38 @@ import (
...
+	csiDriverLister storagelistersv1beta1.CSIDriverLister
+	csiDriverSynced cache.InformerSynced
+
 	snapshotSupported   bool
 	snapshotClassClient snapshotclient.VolumeSnapshotClassInterface
 	snapshotClassLister snapshotlisters.VolumeSnapshotClassLister
 	snapshotClassSynced cache.InformerSynced
 
-	workQueue workqueue.RateLimitingInterface
+	workQueue    workqueue.RateLimitingInterface
```
stoneshi-yunify, line 72: rename this field so we know what queue it is.

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -126,24 +116,10 @@ func NewController(
...
-			newPC := newObj.(*ksstorage.ProvisionerCapability)
-			oldPC := oldObj.(*ksstorage.ProvisionerCapability)
-			if newPC.ResourceVersion == oldPC.ResourceVersion {
-				return
-			}
-			controller.handleProvisionerCapability(newObj)
-		},
+	csiDriverInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    controller.enqueueStorageClassByCSI,
+		UpdateFunc: nil,
```
stoneshi-yunify, line 121: remove it if it's nil

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -126,24 +116,10 @@ func NewController(
...
+		DeleteFunc: controller.enqueueStorageClassByCSI,
```
stoneshi-yunify, line 122: do we need handle the delete event? What would you do if a csi driver was deleted? Remove the allow-xxx annotations?

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -47,46 +47,38 @@ import (
...
 	snapshotClassLister snapshotlisters.VolumeSnapshotClassLister
 	snapshotClassSynced cache.InformerSynced
 
-	workQueue workqueue.RateLimitingInterface
+	workQueue    workqueue.RateLimitingInterface
+	csiWorkQueue workqueue.RateLimitingInterface
 }
 
 // This controller is responsible to watch StorageClass/ProvisionerCapability.
 // And then update StorageClassCapability CRD resource object to the newest status.
```
dkeven, line 77: This comment should be replaced by a new one that describes what this controller does.

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -182,29 +153,35 @@ func (c *StorageCapabilityController) Run(threadCnt int, stopCh <-chan struct{})
...
 	}
-	c.workQueue.Add(key)
+	objs, err = c.storageClassLister.List(labels.NewSelector())
+	if err != nil {
+		utilruntime.HandleError(err)
+		return
+	}
+	for _, obj := range objs {
+		if obj.Provisioner == key {
+			c.workQueue.Add(obj.Name)
```
stoneshi-yunify, line 181: should call `enqueueStorageClass` instead of add it to queue directly

pkg/controller/storage/capability/capability_controller.go
f10atin9: yes，the annotations will be removed.

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -257,40 +234,20 @@ func (c *StorageCapabilityController) syncHandler(key string) error {
 	storageClass, err := c.storageClassLister.Get(name)
 	if err != nil {
 		// StorageClass has been deleted, delete StorageClassCapability and VolumeSnapshotClass
-		if errors.IsNotFound(err) {
-			if c.snapshotAllowed() {
-				err = c.deleteSnapshotClass(name)
```
stoneshi-yunify, line 262: please split the snapshotclass handling to a separated controller. A controller should only do one thing. This is not your fault, but improve this in this PR.

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -47,46 +47,38 @@ import (
...
 	snapshotClassClient snapshotclient.VolumeSnapshotClassInterface
 	snapshotClassLister snapshotlisters.VolumeSnapshotClassLister
 	snapshotClassSynced cache.InformerSynced
 
-	workQueue workqueue.RateLimitingInterface
+	workQueue    workqueue.RateLimitingInterface
+	csiWorkQueue workqueue.RateLimitingInterface
 }
 
 // This controller is responsible to watch StorageClass/ProvisionerCapability.
```
stoneshi-yunify, line 76: review this and all other comments in this controller. Most of them need rewrite.

pkg/controller/storage/capability/capability_controller.go
f10atin9: What do you think of  storageClassWorkQueue？

pkg/controller/storage/capability/capability_controller.go
f10atin9: Copy that.

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -257,40 +234,20 @@ func (c *StorageCapabilityController) syncHandler(key string) error {
...
-			return err
-		}
-		// Don't delete the already created SCC
-		// as it might be created manually by user
-		return nil
-	}
-	klog.Infof("StorageClass %s has capability %v", name, capabilitySpec)
-
+	//Cloning and volumeSnapshot support only available for CSI drivers.
+	withCapability := c.supportCapability(storageClass)
```
stoneshi-yunify, line 247: `withCapability` is vague. Rename it to `isCSIStorage` is something like this.

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -306,63 +263,71 @@ func (c *StorageCapabilityController) syncHandler(key string) error {
...
-		return err
+		return nil
 	}
-	// If StorageClassCapability exist, update it.
-	storageClassCapabilityUpdate := storageClassCapabilityExist.DeepCopy()
-	storageClassCapabilityUpdate.Spec = *capabilitySpec
-	if !reflect.DeepEqual(storageClassCapabilityExist, storageClassCapabilityUpdate) {
-		klog.Info("Update StorageClassCapability: ", storageClassCapabilityUpdate)
-		_, err = c.storageClassCapabilityClient.Update(context.Background(), storageClassCapabilityUpdate, metav1.UpdateOptions{})
+	_, err = c.storageClassClient.Update(context.Background(), storageClass, metav1.UpdateOptions{})
```
dkeven, line 276: You should do a DeepCopy() before changing the object.

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -306,63 +263,71 @@ func (c *StorageCapabilityController) syncHandler(key string) error {
...
-			return err
+			storageClass.Annotations[annotationSupportSnapshot] = strconv.FormatBool(c.snapshotSupported)
+		}
+	} else {
+		if storageClass.Annotations != nil && c.snapshotSupported {
+			if _, ok := storageClass.Annotations[annotationSupportSnapshot]; ok {
+				delete(storageClass.Annotations, annotationSupportSnapshot)
+			}
 		}
 	}
```
stoneshi-yunify, line 310: this snippet code needs improvement. should just:
```
if annotations == nil {
    annotations = make(map)
}

annotations[snapshot] = snapshotsupported

return nil
```

same for the `addCloneVolumeAnnotation` function

stoneshi-yunify (changes requested):
review summary:
- split snapshotclass and capability in two controllers
- capability controller should only handle the storageclass add event, maybe the update event, but never the delete event.
- watch all storageclass, not just the csi storageclass. default snapshot/clone to true for csi storageclass and false for other storageclass. Put this true/false value to the storageclass annotation.

pkg/controller/storage/capability/capability_controller.go
f10atin9: I think this will modify the value set by the user.

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -306,63 +263,71 @@ func (c *StorageCapabilityController) syncHandler(key string) error {
...
+	if err != nil {
 		return err
 	}
 	return nil
 }
 
-func (c *StorageCapabilityController) updateStorageClassSnapshotSupported(storageClass *storagev1.StorageClass, snapshotSupported bool) error {
-	if storageClass.Annotations == nil {
-		storageClass.Annotations = make(map[string]string)
+func (c *StorageCapabilityController) supportCapability(storageClass *storagev1.StorageClass) bool {
```
dkeven, line 283: IMO, this func should not be named `supportCapability` as it simply checks whether a CSIDriver exists for this StorageClass, in which case we just ASSUME the SC has these capabilities.

A name like `hasCSIDriver` might be more precise.

pkg/controller/storage/capability/capability_controller.go
stoneshi-yunify: it's okay

pkg/controller/storage/capability/capability_controller.go
stoneshi-yunify: I did not see the remove annotation code. In my opinion, the annotation should not change even if the csi driver is deleted. just like the `allowExpansion`.  They are independent. `I allow it, but the driver does not exist`.

pkg/controller/storage/capability/capability_controller.go
f10atin9: Do we need to remove annotations when removing CSIDriver?

pkg/controller/storage/capability/capability_controller.go
f10atin9: > I did not see the remove annotation code. In my opinion, the annotation should not change even if the csi driver is deleted. just like the `allowExpansion`. They are independent. `I allow it, but the driver does not exist`.

Okay

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -306,63 +263,71 @@ func (c *StorageCapabilityController) syncHandler(key string) error {
...
-		if errors.IsNotFound(err) {
-			// If StorageClassCapability doesn't exist, create it
-			storageClassCapabilityCreate := &ksstorage.StorageClassCapability{ObjectMeta: metav1.ObjectMeta{Name: storageClass.Name}}
-			storageClassCapabilityCreate.Spec = *capabilitySpec
-			klog.Info("Create StorageClassCapability: ", storageClassCapabilityCreate)
-			_, err = c.storageClassCapabilityClient.Create(context.Background(), storageClassCapabilityCreate, metav1.CreateOptions{})
-			return err
-		}
-		return err
+		return nil
```
dkeven, line 274: Should it return `err` instead of `nil`?

pkg/controller/storage/capability/capability_controller.go
```diff
@@ -306,63 +263,71 @@ func (c *StorageCapabilityController) syncHandler(key string) error {
...
+		}
+		_, err := strconv.ParseBool(storageClass.Annotations[annotationSupportClone])
+		if err != nil {
+			storageClass.Annotations[annotationSupportClone] = strconv.FormatBool(cloneSupported)
+		}
+	} else {
+		if storageClass.Annotations != nil {
+			if _, ok := storageClass.Annotations[annotationSupportClone]; ok {
+				delete(storageClass.Annotations, annotationSupportClone)
+			}
```
dkeven, line 327: Would this logic remove the annotations set by the user manually, if the corresponding CSIDriver does not exist?

dkeven:
> @dkeven: changing LGTM is restricted to collaborators

Okay, I'll leave

pkg/controller/storage/snapshotclass/snapshotclass_controller.go
```diff
@@ -0,0 +1,216 @@
...
+	storageClassLister  storagelistersv1.StorageClassLister
+	storageClassSynced  cache.InformerSynced
+	snapshotClassClient snapshotclient.VolumeSnapshotClassInterface
+	snapshotClassLister snapshotlisters.VolumeSnapshotClassLister
+	snapshotClassSynced cache.InformerSynced
+
+	snapshotClassWorkQueue workqueue.RateLimitingInterface
+}
+
+//This controller is responseible to watch StorageClass
```
zryfish, line 59: responseible => responsible, and should leave spaces between `//` and real comment, the rest of this PR is OK

f10atin9: Okay, I have updated it.

PR #4164, opened 2021-08-19
Upgrade ingress api version to networking.k8s.io/v1

### LINKED ISSUE #4068: Upgrade Ingress API version to networking.k8s.io/v1
**What's it about?**
Upgrade Ingress API to networking.k8s.io/v1.  

Ingress is the core component that is used by the KubeSphere Routes feature. KubeSphere is still using the old api "extensions/v1beta1" both Frontend and Backend. However "extensions/v1beta1" Ingress  is deprecated in v1.14+, unavailable in v1.22+. So we need to upgrade it to the latest version to be compatible with the higher version Kubernetes cluster.

**Area Suggestion**
/area console
/area microservice
/area multicluster

/kind feature-request

/good-first-issue

wenhuwang:
Signed-off-by: wenhuwang <976400757@qq.com>

### What type of PR is this?
/kind api-change
/kind feature

Ingress is the core component that is used by the KubeSphere Routes feature. KubeSphere is still using the old api "extensions/v1beta1" both Frontend and Backend. However "extensions/v1beta1" Ingress is deprecated in v1.14+, unavailable in v1.22+. So we need to upgrade it to the latest version to be compatible with the higher version Kubernetes cluster.

Fixes #4068

```
```

```release-note
None
```

```docs

```

### DIFF
--- a/staging/src/kubesphere.io/api/types/v1beta1/federatedingress_types.go
+++ b/staging/src/kubesphere.io/api/types/v1beta1/federatedingress_types.go
@@ -17,7 +17,7 @@ limitations under the License.
 package v1beta1
 
 import (
-	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
+	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -45,7 +45,7 @@ type FederatedIngressSpec struct {
 }
 
 type IngressTemplate struct {
-	Spec extensionsv1beta1.IngressSpec `json:"spec,omitempty"`
+	Spec networkingv1.IngressSpec `json:"spec,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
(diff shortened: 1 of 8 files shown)


staging/src/kubesphere.io/api/types/v1beta1/federatedingress_types.go
```diff
@@ -45,7 +45,7 @@ type FederatedIngressSpec struct {
 }
 
 type IngressTemplate struct {
-	Spec extensionsv1beta1.IngressSpec `json:"spec,omitempty"`
+	Spec networkingv1.IngressSpec `json:"spec,omitempty"`
```
RolandMa1986, line 48: @yuswift @wenhuwang 
Needs to consider FederatedIngress API version conversion. Build-in Ingress resources could be handled with Kube-apiserver by default. But not for CRD.

yuswift: We don't use the file to generate crd files. So I think we can skip it.

wenhuwang: Subsequent ingress will use the network.k8s.io/v1 version, I think IngressTemplate should also be upgraded together

yuswift: Yes, it should be upgraded.

Issue #3055, opened 2020-11-02
[Proposal]Refactor KubeSphere gateway with CRD


zryfish:
KubeSphere gateway now is just a normal service with Nginx controller as its backend. It offers a declarative API to create/modify/delete. The biggest shortcoming of this approach is not flexible to add extra configurations to it, such as increase replicas, upgrade controller version etc. 

/assign @zryfish 
/kind feature

zryfish:
Planning features listed as the following:

- [ ] Current gateway features are all kept, including change gateway type( nodeport <=> loadbalancer), enable application governing, add extra annotations.
- [ ] enable gateway for Cluster scope
- [ ] enable gateway for Project scope
- [ ] nginx configuration, [limited support](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/)
- [ ] increase/decrease gateway replicas
- [ ] check gateway logs
- [ ] metrics of gateway, including request total volume/ success rate / latency [scrape config](https://github.com/kubernetes/ingress-nginx/blob/master/deploy/prometheus/prometheus.yaml)

Features are not included in this iteration:
- [ ] ~~enable gateway for Workspace scope~~

zryfish:
Moving it to BackLog, cause we are going to miss 3.1.0 release deadline

liangyuanpeng:
it is possible to support multiple ingress?

RolandMa1986:
The Ingress Controller is going to be replaceable with this refactor. We have designed a sample of Gateway Operator with Nginx Ingress controller. Every Ingress Controller can be easily converted to a KubeSphere Gateway with this mode. 
The POC can be found at https://github.com/RolandMa1986/gateway-operator.

RolandMa1986:
One of the issues that haven't metioned above is how do we upgrade from the old project gateway.
We can hardly perform a rolling upgrade without service interapton when upgrading from the older gateway to the Gateway Operator.  The Deployment in the Nginx Ingress Contoller's Helm chart is different from the old YAML definition(that are some immutable fields were changed). We had to delete the old deployment while upgrading.
So we may need to let the user self decide whether/when to upgrade the older gateway. There are 2 possible solutions:
1. Provider a Gateway switch in the Installer. Users can enable the Gateway feature to automatically upgrade from the old gateway. Or, users still can use the older gateway if the Gateway feature is not enabled. 
[...]

PR #4178, opened 2021-08-26
feat: integrate a helm sdk that use to install gateway helm chart

RolandMa1986:
### What type of PR is this?
/kind feature

As feature #3055 mentioned, we need a flexible way to deploy the kubesphere gateway. One of the solutions is using the Operator model to deploy the related gateway resources. 
In this PR:
1. Helm Operator SDK has been integrated into the ks-controller-manager, we can easily convert any helm chart into an Operator. 
2. An `watches.yaml` config file can be configurated in the gateway options,  which use to monitor the creation of Gateway resource and deployment gateway helm chart.
```
- group: gateway.kubesphere.io
  version: v1alpha1
  kind: Gateway
  chart: config/gateway
```
3. Added a gateway helm that use to render the Nginx Ingress Controller Helm chart.
3. Upgraded Helm package to helm 3.6.1

Fixes ##3055

```
```

```release-note
None
```

```docs

```


RolandMa1986:
@zryfish Please take a look at this PR.

pkg/apiserver/config/config_test.go
```diff
@@ -179,6 +180,10 @@ func newTestConfig() (*Config, error) {
 		MeteringOptions: &metering.Options{
 			RetentionDay: "7d",
 		},
+		GatewayOptions: &gateway.Options{
+			WatchesPath: "/etc/kubesphere/watches.yaml",
```
zryfish, line 184: What does the `watches.yaml` look like, and why do we need it?

pkg/apiserver/config/config.go
```diff
@@ -104,6 +105,7 @@ type Config struct {
 	NotificationOptions   *notification.Options                      `json:"notification,omitempty" yaml:"notification,omitempty" mapstructure:"notification"`
 	KubeEdgeOptions       *kubeedge.Options                          `json:"kubeedge,omitempty" yaml:"kubeedge,omitempty" mapstructure:"kubeedge"`
 	MeteringOptions       *metering.Options                          `json:"metering,omitempty" yaml:"metering,omitempty" mapstructure:"metering"`
+	GatewayOptions        *gateway.Options                           `json:"gateway,omitempty" yaml:"router,omitempty" mapstructure:"router"`
```
zryfish, line 108: Should we use `gateway` for the YAML as well? I am a little worried about the difference here

config/gateway/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+service:
+  ## annotations for Services, used for config Cloud LoadBalancer
+  annotations: {}
+  type: LoadBalancer
+
+## for nginx controller, same with controller
+deployment:
+  annotations: {}
+  replicas: 1
```
zryfish, line 25: Should add newline at the end of the file.

pkg/apiserver/config/config.go
RolandMa1986: Yes, it should be `gateway`. fixed.

pkg/apiserver/config/config_test.go
RolandMa1986: Please check the comments in the PR (2nd item). Generally, it's a mapping relation between CRD and the Helm Chart. The operator will monitor the CRD configured in the file, when CR is created and a corresponding Helm Release will be created too.
Currently, the Helm Operator SDK only supports file path-based parameters. We may replace it with CRD in the feature.

config/gateway/values.yaml
RolandMa1986: fixed.

PR #4194, opened 2021-08-30
[v3.2] Add grafana dashboard importing API

zhu733756:
Signed-off-by: zhu733756 <talonzhu@yunify.com>

Import Grafana template and convert it to KubeSphere dashboard

/kind design
/area monitoring
/area observability
/kind feature-request
/area console

### Background
Grafana is a successful and widely used observability tool. In the current release, we have supported importing dashboards from Grafana dashboard templates.

Refer to [here](https://github.com/kubesphere/monitoring-dashboard#converter-tool), we have finished features like below:
- a new version v1alpha2 for CRD has been provided, which supports more fields.
- the new version and the older version are compatible with each other.
- add a converter tool, aims to help for importing Grafana dashboard templates, and convert to kubesphere dashboard manifests.

### What this PR does / why we need it:

This pr aims to integrate these interfaces on the backend section.

For a detailed API description, the API integrates with a grouped version "monitoring.kubesphere.io/v1alpha3". 
The completed request API maybe like this as follows:
```
/kapis/monitoring.kubesphere.io/v1alpha3/dashboard/template
```
We can reach the backend by post request with three parameters:
- `grafanaDashboardName`:  the name of this Grafana template needed to convert.
- `grafanaDashboardUrl`: the link to download this Grafana template. 
- `grafanaDashboardContent`: the whole JSON content needed to convert.
> Note that the parameter `grafanaDashboardName` is indispensable, and the requested parameter `grafanaDashboardUrl` and `grafanaDashboardContent` cannot be empty at the same time.

### Which issue(s) this PR fixes:

fix #4195

### Special notes for reviewers:
Currently, this API only supports create cluster scope dashboard.
A sample can be described as follow:
```
curl -d '{"grafanaDashboardName":"test2","grafanaDashboardUrl":"https://grafana.com/api/dashboards/7362/revisions/5/download"}' -H "Content-Type: application/json" localhost:9090/kapis/monitoring.kubesphere.io/v1alpha3/dashboard/template
```

Issue #4195, opened 2021-08-30
[v3.2] Import Grafana Dashboard in custom monitoring


zhu733756:
## Background
Grafana is a successful and widely used observability tool. In the current release, we have supported importing dashboards from Grafana dashboard templates.

Refer to [here](https://github.com/kubesphere/monitoring-dashboard#converter-tool), we have finished features like below:
- a new version v1alpha2 for CRD has been provided, which supports more fields.
- the new version and the older version are compatible with each other.
- add a converter tool, aims to help for importing Grafana dashboard templates, and convert to kubesphere dashboard manifests.

In the near future, we will integrate these interfaces on the backend section.

## Plans
* [ ] Import grafana dashboard and convert it to KubeSphere dashboard

/kind design
/area monitoring
/area observability
/kind feature-request

PR #4203, opened 2021-09-01
fix API specification for #4194

zhu733756:
Signed-off-by: zhu733756 <talonzhu@yunify.com>

Import Grafana template and convert it to KubeSphere dashboard

/kind design
/area monitoring
/area observability
/kind feature-request
/area console

fixes #4194 for better API specification.

fixes #4194 for better API specification, see issue #4194  or pr #4195 for a detailed introduction.

Currently, this API only supports create clusterdashboards.

The API changes to:
```
/kapis/monitoring.kubesphere.io/v1alpha3/clusterdashboards/{grafanaDashboardName}/template
```

Add new field `Description` to the post body:
- Description:  the description of this dashboard.

A sample can be described as follow:
```
#! /bin/bash
curl -d '{"grafanaDashboardUrl":"https://grafana.com/api/dashboards/7362/revisions/5/download", "description":"this is a test dashboard."}' -H "Content-Type: application/json" localhost:9090/kapis/monitoring.kubesphere.io/v1alpha3/clusterdashboards/test1/template
```

```release-note
None
```

/cc @benjaminhuo @wansir

### DIFF
--- a/pkg/kapis/monitoring/v1alpha3/dashboard_template_test.sh
+++ b/pkg/kapis/monitoring/v1alpha3/dashboard_template_test.sh
@@ -1,3 +1,3 @@
 #! /bin/bash
 
-curl -d '{"grafanaDashboardUrl":"https://grafana.com/api/dashboards/7362/revisions/5/download", "description":"this is a test dashboard."}' -H "Content-Type: application/json" localhost:9090/kapis/monitoring.kubesphere.io/v1alpha3/clusterdashboard/test1/template
+curl -d '{"grafanaDashboardUrl":"https://grafana.com/api/dashboards/7362/revisions/5/download", "description":"this is a test dashboard."}' -H "Content-Type: application/json" localhost:9090/kapis/monitoring.kubesphere.io/v1alpha3/clusterdashboards/test1/template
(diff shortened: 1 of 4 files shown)


zhu733756:
As mentioned by @harrisonliu5. 
> todo:  add a description field to this API.

benjaminhuo:
`Does this PR introduced a user-facing change?` should be set to None because the same release notes already exist in the original issue, no need to add the same here again.

zhu733756:
> `Does this PR introduced a user-facing change?` should be set to None because the same release notes already exist in the original issue, no need to add the same here again.

fixed

zhu733756:
> /kapis/monitoring.kubesphere.io/v1alpha3/clusterdashboard/{grafanaDashboardName}/template
> =>
> /kapis/monitoring.kubesphere.io/v1alpha3/clusterdashboards/{grafanaDashboardName}/template

fixed

benjaminhuo:
@zhu733756 As I mentioned there is no need to add a release notes for this PR, because the one in the original PR is enough for this feature, no need to add something like `Since adding the field description to this API, the console would have a design for it.`, set it to `None` is just ok

zhu733756:
> @zhu733756 As I mentioned there is no need to add release notes for this PR, because the one in the original PR is enough for this feature, no need to add something like `Since adding the field description to this API, the console would have a design for it.`, set it to `None` is just ok

fixed

PR #4204, opened 2021-09-01
helm: allows ks-core to mount extra volumes

RolandMa1986:

### What type of PR is this?
/kind feature

This PR would allow the ks-core helm chart to mount extra volumes.  

Feature #3055 requires loading the gateway helm chart and configuration file from the local filesystem.  We can either package them into the container image or mount them when deploying. Mount those files as volumes would provide more flexibility, Users can upgrade the components independently. 

Fixes #

```
```

```release-note
None
```

```docs

```

### DIFF
--- a/config/ks-core/templates/ks-apiserver.yml
+++ b/config/ks-core/templates/ks-apiserver.yml
@@ -39,7 +39,7 @@ spec:
         - containerPort: 9090
           protocol: TCP
         resources:
-          {{- toYaml .Values.apiserverResources | nindent 12 }}
+          {{- toYaml .Values.apiserver.resources | nindent 12 }}
         volumeMounts:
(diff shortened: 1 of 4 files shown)


yuswift:
BTW, the ks-installer should be updated too.

zryfish:
> We can either package them into the container image or mount them when deploying. Mount those files as volumes would provide more flexibility, Users can upgrade the components independently.

Agreed. But I think we still need to package a default template into the container image while users can override by mounting other templates.

RolandMa1986:
> > We can either package them into the container image or mount them when deploying. Mount those files as volumes would provide more flexibility, Users can upgrade the components independently.
> 
> Agreed. But I think we still need to package a default template into the container image while users can override by mounting other templates.

Sure.

Issue #4158, opened 2021-08-18
Support setting host cluster name 


swiftslee:
Currently, the host cluster name is hard-coded to `host`. 
![image](https://user-images.githubusercontent.com/37265389/130204038-94badce0-81eb-47b6-a4ed-1ae4842bf02a.png)

We need to specify the host cluster name in the `cc` config and convert it into `ks-console-config` configmap.

yuswift:
Here is the configMap of ks-console-config and the ks-installer config. What about adding a `hostClusterName` field? @zryfish 
```
server:
  http:
    hostname: localhost
    port: 8000
    static:
      production:
        /public: server/public
        /assets: dist/assets
        /dist: dist
  redis:
    port: 6379
    host: redis.kubesphere-system.svc
  redisTimeout: 5000
  sessionTimeout: 7200000
client:
  version:
    kubesphere: v3.1.1
    kubernetes: v1.19.8
    openpitrix: v0.3.5
  enableKubeConfig: true
hostClusterName: host
```


ks-installer config
```
  multicluster:
    clusterRole: none  # host | member | none  # You can install a solo cluster, or specify it as the Host or Member Cluster.
    hostClusterName: host
```

zryfish:
hostClusterName is reasonable in multicluster setup, I think `defaultClusterName` is more appropriate in this scenario. Host cluster is created by cluster controller,  maybe we could pass `defaultClusterName` to cluster controller when convert single cluster to host cluster. ks-installer could do this for us, but first, maybe we need to add a host cluster name flag for the cluster controller .

zryfish:
@harrisonliu5  Any thoughts? Is it possible to change the default cluster name `default` to, for example, `gondor`?

yuswift:
> hostClusterName is reasonable in multicluster setup, I think `defaultClusterName` is more appropriate in this scenario. Host cluster is created by cluster controller, maybe we could pass `defaultClusterName` to cluster controller when convert single cluster to host cluster. ks-installer could do this for us, but first, we need to add a host cluster name flag for the cluster controller .

For console, it is complicated when handling different scenarios, and we need two params for ks-installer. IMO, one param is enough and it's convenient for console, although it will mislead the users that the host cluster name is still `host`.

harrisonliu5:
If you use the hostClusterName in the configMap of ks-console-config as the cluster name, there may be errors when some resources need the cluster name to obtain data. It is better to get the cluster name according to the backend data.

[...]

PR #4211, opened 2021-09-05
feature: support setting host cluster name

yuswift:
Signed-off-by: yuswift <yuswift2018@gmail.com>

### What type of PR is this?
/kind feature
We can set the host cluster name in the `kubesphere.yaml` now, e.g:
```
authentication:
  authenticateRateLimiterMaxTries: 10
  authenticateRateLimiterDuration: 10m0s
  loginHistoryRetentionPeriod: 168h
  maximumClockSkew: 10s
  multipleLogin: True
  kubectlImage: kubesphere/kubectl:v1.19.1
  jwtSecret: "qTcNlzhUlTa6bjo4lZaXG1Xr"
network:
  ippoolType: none
multicluster:
  enable: true
  agentImage: kubesphere/tower:v0.2.0
  proxyPublishService: tower.kubesphere-system.svc
  hostClusterName: control
```

part of #4158 
```
```

```release-note
the cluster controller now supports setting the host cluster name, default set to host
```

```docs

```

### DIFF
--- a/cmd/controller-manager/app/controllers.go
+++ b/cmd/controller-manager/app/controllers.go
@@ -190,7 +190,8 @@ func addControllers(
 			client.Config(),
 			kubesphereInformer.Cluster().V1alpha1().Clusters(),
 			client.KubeSphere().ClusterV1alpha1().Clusters(),
-			multiClusterOptions.ClusterControllerResyncSecond)
+			multiClusterOptions.ClusterControllerResyncPeriod,
+			multiClusterOptions.HostClusterName)
 	}
 
 	var nsnpController manager.Runnable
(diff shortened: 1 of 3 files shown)


zryfish:
What if the user wants to change the name of the existing host cluster?

yuswift:
> What if the user wants to change the name of the existing host cluster?

The name of the cluster can not be changed after being created. I think there is a workaround for the users `stop the cluster controller, delete the host cluster, update the kubesphere.yaml, restart the controller`

yuswift:
> > What if the user wants to change the name of the existing host cluster?
> 
> The name of the cluster can not be changed after being created. I think there is a workaround for the users `stop the cluster controller, delete the host cluster, update the kubesphere.yaml, restart the controller`

But the solution is a very bad idea. The users shouldn't change the cluster name, because it will affect the federation resources which are managed by kubfed.

PR #4214, opened 2021-09-06
Move scope-misplaced resource getters to the right place

f10atin9:
Signed-off-by: f10atin9 <f10atin9@kubesphere.io>

### What type of PR is this?

/kind bug

### What this PR does / why we need it:
It's semantically correct and more efficient to retrieve when it's amended.
### Which issue(s) this PR fixes:

Fixes #

### Special notes for reviewers:
```
```

### Does this PR introduced a user-facing change?

```release-note
None
```

### Additional documentation, usage docs, etc.:

```docs

```

PR #4215, opened 2021-09-06
Move scope-misplaced resource getters to the right place

dkeven:
Signed-off-by: dkeven <keven@kubesphere.io>

### What type of PR is this?
/kind cleanup

The resource getters for the `namespace` and `node` objects, which are cluster scoped objects, are misplaced under the `NamespacedResourceGetters` map:
https://github.com/kubesphere/kubesphere/blob/51955831a774d4d2b9132b638c9d1c64ac483f33/pkg/models/resources/v1alpha3/resource/resource.go#L105
https://github.com/kubesphere/kubesphere/blob/51955831a774d4d2b9132b638c9d1c64ac483f33/pkg/models/resources/v1alpha3/resource/resource.go#L109 
They should be moved to the `ClusterResourceGetters` map.

```release-note
None
```

### DIFF
--- a/pkg/kapis/resources/v1alpha3/handler_test.go
+++ b/pkg/kapis/resources/v1alpha3/handler_test.go
@@ -49,7 +49,7 @@ func TestResourceV1alpha2Fallback(t *testing.T) {
 	}{
 		{
 			description: "list namespaces",
-			namespace:   "default",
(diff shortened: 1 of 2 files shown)


yuswift:
duplicate #4214

PR #4227, opened 2021-09-10
feat: ingress metrics query apis

RolandMa1986:

### What type of PR is this?
/kind feature
/kind api-change

Metrics of the gateway that shows on the gateway page. The named querys includes:

- ingress_request_count",
- ingress_request_5xx_count
- ingress_request_4xx_count
- ingress_active_connections
- ingress_success_rate
- ingress_request_duration_average
- ingress_request_duration_50percentage
- ingress_request_duration_95percentage
- ingress_request_duration_99percentage
- ingress_request_volume
- ingress_request_volume_by_ingress
- ingress_request_network_sent
- ingress_request_network_received
- ingress_request_memory_bytes
- ingress_request_cpu_usage

For a **None ranged query** the query parameters like bellow:
kapis/monitoring.kubesphere.io/v1alpha3/namespaces/{namespace}/ingresses?metrics_filter={metrices}&time={currrent time}&job={matires-job}&step={duration}&pod={pod}

`step` specific the time windows of the Range Vector data that used to calculate the metric. default to 5 minutes.

Fixes ##3055

```
```

```release-note
None
```

```docs

```

### DIFF
--- a/pkg/kapis/monitoring/v1alpha3/helper.go
+++ b/pkg/kapis/monitoring/v1alpha3/helper.go
@@ -91,6 +91,7 @@ type reqParams struct {
 	ingress                   string
 	job                       string
 	services                  string
+	duration                  string
 	pvcFilter                 string
 	queryType                 string
 }
@@ -152,6 +153,7 @@ func parseRequestParams(req *restful.Request) reqParams {
 	r.componentType = req.PathParameter("component")
 	r.ingress = req.PathParameter("ingress")
 	r.job = req.QueryParameter("job")
+	r.duration = req.QueryParameter("duration")
 	r.expression = req.QueryParameter("expr")
 	r.metric = req.QueryParameter("metric")
 	r.queryType = req.QueryParameter("type")
@@ -345,12 +347,12 @@ func (h handler) makeQueryOptions(r reqParams, lvl monitoring.Level) (q queryOpt
 
 	case monitoring.LevelIngress:
 		q.identifier = model.IdentifierIngress
-		var step *time.Duration
-		// step param is reused in none Range Query to pass vector's time duration.
+		var du *time.Duration
+		// duration param is used in none Range Query to pass vector's time duration.
 		if r.time != "" {
-			s, err := time.ParseDuration(r.step)
+			s, err := time.ParseDuration(r.duration)
 			if err == nil {
-				step = &s
+				du = &s
 			}
 		}
 		q.option = monitoring.IngressOption{
@@ -359,7 +361,7 @@ func (h handler) makeQueryOptions(r reqParams, lvl monitoring.Level) (q queryOpt
 			Ingress:        r.ingress,
 			Job:            r.job,
 			Pod:            r.podName,
-			Step:           step,
+			Duration:       du,
 		}
 		q.namedMetrics = model.IngressMetrics
 
--- a/pkg/kapis/monitoring/v1alpha3/helper.go
+++ b/pkg/kapis/monitoring/v1alpha3/helper.go
@@ -88,6 +88,8 @@ type reqParams struct {
 	applications              string
 	openpitrixs               string
 	cluster                   string
+	ingress                   string
+	job                       string
 	services                  string
 	pvcFilter                 string
 	queryType                 string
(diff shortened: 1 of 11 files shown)


RolandMa1986:
/cc @junotx 
Could help review this PR? Thanks

pkg/simple/client/monitoring/prometheus/promql.go
```diff
@@ -165,6 +165,21 @@ var promQLTemplates = map[string]string{
...
+	"ingress_request_duration_average":      `sum_over_time(nginx_ingress_controller_request_duration_seconds_sum{$1,$2}[$3])/sum_over_time(nginx_ingress_controller_request_duration_seconds_count{$1,$2}[$3])`,
+	"ingress_request_duration_50percentage": `histogram_quantile(0.50, sum by (le) (rate(nginx_ingress_controller_request_duration_seconds_bucket{$1,$2}[$3])))`,
+	"ingress_request_duration_95percentage": `histogram_quantile(0.90, sum by (le) (rate(nginx_ingress_controller_request_duration_seconds_bucket{$1,$2}[$3])))`,
+	"ingress_request_duration_99percentage": `histogram_quantile(0.99, sum by (le) (rate(nginx_ingress_controller_request_duration_seconds_bucket{$1,$2}[$3])))`,
+	"ingress_request_volume":                `round(sum(irate(nginx_ingress_controller_requests{$1,$2}[$3])), 0.001)`,
+	"ingress_request_volume_by_ingress":     `round(sum(irate(nginx_ingress_controller_requests{$1,$2}[$3])) by (ingress), 0.001)`,
+	"ingress_request_network_sent":          `sum(irate(nginx_ingress_controller_response_size_sum{$1,$2}[$3]))`,
+	"ingress_request_network_received":      `sum(irate(nginx_ingress_controller_request_size_sum{$1,$2}[$3]))`,
+
```
junotx, line 182: The promql statements may have syntax problems when `$1` is empty or both `$1` and `$2` are empty (although they are not empty by the api's parameter settings), so it is recommended to combine them into one

pkg/simple/client/monitoring/prometheus/promql.go
```diff
@@ -448,3 +465,37 @@ func makePVCMetricExpr(tmpl string, o monitoring.QueryOptions) string {
...
+func makeIngressMetricExpr(tmpl string, o monitoring.QueryOptions) string {
+	var ingressSelector string
+	var jobSelector string
+	duration := "5m"
+
+	// parse Range Vector Selectors metric{key=value}[duration]
+	if o.MeterOptions != nil {
+		duration = o.MeterOptions.Step.String()
+	}
```
junotx, line 477: It is not really good that the `step` parameter in range query specify both the step of the range query and time window of range vector

junotx (changes requested):

pkg/simple/client/monitoring/prometheus/promql.go
RolandMa1986: > The promql statements may have syntax problems when `$1` is empty or both `$1` and `$2` are empty (although they are not empty by the api's parameter settings), so it is recommended to combine them into one

It's hard to merge those 2 parameters `ingress` and `job` into one param. Because both those 2 parameters are required in the API level(for security reasons). but in some of the promql only need one, because the `ingress `param is not saved in the data.

junotx: ok, it is fine as long as `$1` and `$2` are not empty when they are replaced

pkg/simple/client/monitoring/prometheus/promql.go
RolandMa1986: @junotx Refined. Added new parameter "duration" to the query string.

pkg/kapis/monitoring/v1alpha3/helper.go
```diff
@@ -337,6 +345,26 @@ func (h handler) makeQueryOptions(r reqParams, lvl monitoring.Level) (q queryOpt
...
+	case monitoring.LevelIngress:
+		q.identifier = model.IdentifierIngress
+		var du *time.Duration
+		// duration param is used in none Range Query to pass vector's time duration.
+		if r.time != "" {
+			s, err := time.ParseDuration(r.duration)
+			if err == nil {
+				du = &s
+			}
+		}
```
junotx, line 357: Why is the `duration` parameter restricted to instant query only

RolandMa1986: There's no place to set  `duration `in the chart controls and it would be confused for users how to set the value.
The following screenshot shows how the instant query used on the UI

junotx: So here the `duration` is set instead of `start` and `end` time, which looks different from other monitoring pages

junotx (approved):
/lgtm

PR #4228, opened 2021-09-10
Fix inconsistent user status in member cluster

### LINKED ISSUE #4167: The status of project members in multi cluster environment is incorrect
**Describe the Bug**

The user is active on the account management page, but it's disabled on the member cluster project members page. The user status is incorrect if login from another identity provider.

image

image

**Versions Used**
KubeSphere: v3.1.0

**Expected behavior**

The user's status is consistent.

wansir:
Signed-off-by: hongming <hongming@kubesphere.io>

### What type of PR is this?
/kind bug

The user label [will not be synchronized to the member cluster](https://github.com/kubesphere/kubesphere/blob/release-3.1/pkg/controller/user/user_controller.go#L335-L341), but the user controller in the member cluster will modify the user status according to the incorrect label. In fact, the user status is controlled by the user controller in the host cluster.

The user status is inconsistent due to the user label cannot be synchronized to the member cluster. The user-controller in the member cluster should not manage user spec and user status(which is managed by kubefed).
 
Fixes #4167

```release-note
NONE
```

### DIFF
--- a/pkg/controller/user/user_controller.go
+++ b/pkg/controller/user/user_controller.go
@@ -18,43 +18,37 @@ package user
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"reflect"
 	"time"
+
+	"github.com/go-logr/logr"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/types"
+	typesv1beta1 "kubesphere.io/api/types/v1beta1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	authoptions "kubesphere.io/kubesphere/pkg/apiserver/authentication/options"
 
 	"k8s.io/apimachinery/pkg/util/validation"
 
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
 
-	"kubesphere.io/kubesphere/pkg/controller/utils/controller"
-
 	"golang.org/x/crypto/bcrypt"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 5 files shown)


pkg/controller/user/user_controller.go
```diff
@@ -18,43 +18,37 @@ package user
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"reflect"
 	"time"
 
+	"github.com/go-logr/logr"
```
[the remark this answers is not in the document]
wansir: We use `klogr` as a delegate Logger(which is defined in  [controller-runtime/blob/master/pkg/log/log.go](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/log/log.go)).

`ctrl.Log` is the base logger used by controller-runtime.  It delegates to klog.

https://github.com/kubesphere/kubesphere/blob/master/cmd/controller-manager/app/server.go#L199

staging/src/kubesphere.io/api/types/v1beta1/federateduser_types.go
```diff
@@ -45,7 +45,10 @@ type FederatedUserSpec struct {
 }
 
 type UserTemplate struct {
-	Spec v1alpha2.UserSpec `json:"spec,omitempty"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              v1alpha2.UserSpec `json:"spec"`
+	// +optional
+	Status v1alpha2.UserStatus `json:"status,omitempty"`
```
RolandMa1986, line 51: @wansir  There are a lot of code refactoring in the commits which seems doesn't relate to the root cause.  Could summarize how does the `User` reconcile be skipped in the Member cluster?

wansir: I've refactored the user controller with controller runtime, controller runtime is simpler than resource client. The federated types defined in `federateduser_types.go` is incorrect.

wansir: The user label will not be synchronized to the member cluster, but the user controller in the member cluster will modify the user status according to the incorrect label. In fact, the user status is controlled by the user controller in the host cluster.

PR #4238, opened 2021-09-15
Support skip information reconfirm when using external IDP

### LINKED ISSUE #4177: Allow to configure ad/ldap identifier to map userid/domain to ks user account automatically
**What's it about?**

By configuring ldap/ad to KubeSphere, the console will show a confirm window when ldap/ad login for the first time, it's not user-friendly.
We'd better allow to configure ad/ldap identifier to map userid/domain to ks user account automatically.

* every user in ldap/ad has to confirm that user name, it's not user-friendly.

**Area Suggestion**

/kind feature-request
/area iam

wansir:
Signed-off-by: hongming <hongming@kubesphere.io>

### What type of PR is this?

/kind feature
/area auth

Skip information confirmation will reduce user login steps when using a custom identity provider.

Fixes #4177

```release-note
Support skip information confirmation when using external IDP
```

### DIFF
--- a/pkg/apiserver/authentication/oauth/oauth_options.go
+++ b/pkg/apiserver/authentication/oauth/oauth_options.go
@@ -37,7 +37,8 @@ const (
 	GrantHandlerPrompt GrantHandlerType = "prompt"
 	// GrantHandlerDeny auto-denies client authorization grant requests
 	GrantHandlerDeny GrantHandlerType = "deny"
-	// MappingMethodAuto  The default value.The user will automatically create and mapping when login successful.
+	// MappingMethodAuto  The default value.
+	// The user will automatically create and mapping when login successful.
 	// Fails if a user with that username is already mapped to another identity.
 	MappingMethodAuto MappingMethod = "auto"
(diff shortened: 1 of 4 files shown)


pkg/apiserver/authentication/oauth/oauth_options.go
```diff
@@ -141,6 +142,11 @@ type IdentityProviderOptions struct {
 	//  - mixed:  A user entity can be mapped with multiple identifyProvider.
 	MappingMethod MappingMethod `json:"mappingMethod" yaml:"mappingMethod"`
 
+	// ConfirmationNotRequired means that when the user login successfully,
+	// reconfirm the account information is not required.
+	// Username from IDP must math [a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*
+	ConfirmationNotRequired bool `json:"confirmationNotRequired"`
```
RolandMa1986, line 148: 1. A positive naming convention should be preferred.

wansir: Compatible with the default configuration of the old version. Confirmation is required by default.

PR #4243, opened 2021-09-15
feat: get gateway's workload and status

RolandMa1986:
### What type of PR is this?
/kind feature

1. This RP adds new APIs to retrieved pods and logs for the gateway. pod and log are treated as subresources of the gateways. So that we can simplify the authentication.

/namespaces/{namespace}/gateways/{gateway}/pods 
/namespaces/{namespace}/gateways/{gateway}/pods/{pod}/log

2. Append service's status to the Gateway's status. That is used to show IP and ports in the UI.

Fixes # #3055 

```
```

```release-note
None
```

```docs

```


pkg/kapis/gateway/v1alpha1/handler.go
```diff
@@ -17,30 +17,44 @@ limitations under the License.
...
 	gw      operator.GatewayOperator
+	factory informers.InformerFactory
 }
 
 //newHandler create an instance of the handler
-func newHandler(options *gateway.Options, cache cache.Cache, client client.Client) *handler {
+func newHandler(options *gateway.Options, cache cache.Cache, client client.Client, factory informers.InformerFactory, k8sClient kubernetes.Interface) *handler {
+	conversionsv1.RegisterConversions(scheme.Scheme)
 	// Do not register Gateway scheme globally. Which will cause conflict in ks-controller-manager.
 	v1alpha1.AddToScheme(client.Scheme())
```
wansir, line 53: Why not register the scheme in `pkg/apis/apis.go` 

https://github.com/kubesphere/kubesphere/blob/master/cmd/ks-apiserver/app/options/options.go#L227

RolandMa1986: We introduced an SDK in `ks-controller-manager` called `Helm Operator plugin`. Which will register an unstructured scheme `Gateway` firstly. If we register the typed `Gateway` scheme in `pkg/apis/apis.go`, the controller-runtime client will throw an exception.

PR #4250, opened 2021-09-17
Dynamically load helm app into app-store

### LINKED ISSUE #4244: Support import new Helm-based Apps into KubeSphere App Store dynamically
**What's it about?**

The built-in applications are tightly coupled with the KubeSphere version, this is not flexible for us to expand the application ecosystem of KubeSphere App Store and build the partnership with more ISVs and partners. 

I propose to support import new Helm-based apps into KubeSphere App Store dynamically. 

* A lot of ISVs or partners are applied to add their Helm charts to KubeSphere App Store, see an [example](https://github.com/kubesphere/helm-charts/pull/150).
* Users can deploy the applications in time when the related Pull Request has been merged, they don't need to wait for a new release of KubeSphere

Please leave your comments below if there's anyone who agrees with me. Or just give me a thumb up.

**Area Suggestion**

/kind feature-request
/area app-management

xyz-li:
### What type of PR is this?
/kind feature

Dynamically load helm app into app store.

As our partners need to add their Helm charts to KubeSphere App Store, I add one helm repo type which is called the `built-in` helm repo. The `built-in` repo's name starts with `builtin-` which differentiates it from the repo created by the user.

ks-controller-manager will periodically load apps in the built-in repo and updates the helm repo CRD.  
When the user lists the app in the app store, the ks-apiserver joins the apps in the build-in repo and apps of HelmApplication.

Fixes #4244

```
```

```release-note
Feature: Dynamically load the community developed helm charts into the app store
```

```docs

```

### DIFF
--- a/pkg/utils/mathutil/mathutil.go
+++ b/pkg/utils/mathutil/mathutil.go
@@ -1,0 +1,9 @@
+package mathutil
+
+// Max returns the larger of a and b.
+func Max(a, b int) int {
+	if a >= b {
+		return a
+	}
+	return b
+}
--- a/pkg/utils/mathutil/mathutil.go
+++ b/pkg/utils/mathutil/mathutil.go
@@ -1,9 +1,0 @@
-package mathutil
-
-// Max returns the larger of a and b.
-func Max(a, b int) int {
-	if a >= b {
-		return a
-	}
-	return b
-}
--- a/pkg/utils/reposcache/repo_cahes.go
+++ b/pkg/utils/reposcache/repo_cahes.go
@@ -19,7 +19,6 @@ package reposcache
 
 import (
 	"context"
-	"encoding/base64"
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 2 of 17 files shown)


pkg/utils/reposcache/repo_cahes.go
```diff
@@ -118,10 +145,47 @@ func loadBuiltinChartData(name, version string) ([]byte, error) {
...
+	defer c.Unlock()
+
 	c.deleteRepo(repo)
 	return nil
 }
 
+// CopyCategoryCount copies the internal map to avoid `concurrent map iteration and map write`.
+func (c *cachedRepos) CopyCategoryCount() map[string]int {
+	c.RLock()
+	c.RUnlock()
```
wansir, line 158:
```suggestion
	defer c.RUnlock()
```

pkg/utils/mathutil/mathutil.go
```diff
@@ -0,0 +1,9 @@
+package mathutil
+
+// Max returns the larger of a and b.
+func Max(a, b int) int {
```
zryfish, line 4: Why not using math.Max https://pkg.go.dev/math#Max

xyz-li: The parameters and return value of `math.Max` are all `float64`. `mathutil.Max(a, b)` should written as `int(math.Max(int(a), int(b)))`, there are too many type conversions.

zryfish: OK, if you prefer this way, I think it's unnecessary to put it into another package, just make it an inner function, it'll be easy to read.

PR #4284, opened 2021-09-26
update ingress version in ks-core chart

### LINKED ISSUE #4283: Failed to deploy sample App
**Describe the Bug**
When Gateway is enabled, Sample App can't be deployed. The Ingress version doesn't match with server side.

ingress

**Versions Used**
KubeSphere:

**Environment**

For example: CentOS 7.5 / 3 masters:  8cpu/8g; 3 nodes: 8cpu/16g
(and other info are welcomed to help us debugging)

**How To Reproduce**

**Expected behavior**

RolandMa1986:

### What type of PR is this?
/kind bug
/kind flake
Since #4068 we have upgraded the Ingress API versions to networking.k8s.io/v1. but the sample app still uses the old API version. So we have to upgrade it to the networking.k8s.io/v1 version.

Fixes #4283

```
```

```release-note
None
```

```docs

```

### DIFF
--- a/config/ks-core/templates/ks-apiserver.yml
+++ b/config/ks-core/templates/ks-apiserver.yml
@@ -41,9 +41,6 @@ spec:
         resources:
           {{- toYaml .Values.apiserver.resources | nindent 12 }}
         volumeMounts:
-        - mountPath: /var/run/docker.sock
-          name: docker-sock
-          readOnly: true
         - mountPath: /etc/kubesphere/ingress-controller
           name: ks-router-config
         - mountPath: /etc/kubesphere/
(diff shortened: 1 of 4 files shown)


yuswift:
ks-installer should be updated too.

config/ks-core/templates/ks-apiserver.yml
```diff
@@ -43,12 +43,14 @@ spec:
         volumeMounts:
         - mountPath: /var/run/docker.sock
```
zryfish, line 44: I think we can remove `docker.sock` in v3.2, since there are no components depending on this.

PR #4385, opened 2021-10-21
if error is not NULL,Maybe return error would be better.

liuan1986:

Fixes #

```
```

```release-note
None
```

```docs

```


zryfish:
Thanks for your contribution

PR #4394, opened 2021-10-25
OAuthClient with {client_id:kubesphere} must be config,if not, use De…

liuan1986:
### What type of PR is this?
/kind bug

use  kubesphere as the client_id is kubesphere internal characteristics，user do not know how to config it,
so we need to init a default oauth client, use kubesphere as its name.

1. without this RP, you must config auth clients, if not ,will cause the following problem ：
image

with this RP, you do not need config auth clients, this RP use kubesphere as default client

2.without this RP ,from v3.3.1 upgrade to the latest version without the kubesphere.yaml congfig changed, will cause the problem:“the OAuth client was not found” ,and this RP can fix this problem.
```release-note
Add kubsphere as default OAuthClient.
```

### DIFF
--- a/cmd/ks-apiserver/app/options/options.go
+++ b/cmd/ks-apiserver/app/options/options.go
@@ -245,6 +245,11 @@ func (s *ServerRunOptions) NewAPIServer(stopCh <-chan struct{}) (*apiserver.APIS
 		klog.Fatalf("unable to create issuer: %v", err)
 	}
 
+	err = s.AuthenticationOptions.OAuthOptions.CheckDefaultOAuthClient()
+	if err != nil {
+		return nil, fmt.Errorf("Check default OAuth client error: %v", err)
+	}
+
 	apiServer.Server = server
 
 	return apiServer, nil
--- a/pkg/apiserver/authentication/oauth/options.go
+++ b/pkg/apiserver/authentication/oauth/options.go
@@ -23,6 +23,8 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/klog/v2"
+
 	"kubesphere.io/kubesphere/pkg/utils/sliceutil"
 )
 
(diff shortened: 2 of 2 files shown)


wansir:
Hard code is not good choice. You'd better add the default oauth client in the [configuration file](https://github.com/kubesphere/kubesphere/blob/master/config/ks-core/templates/kubesphere-config.yaml). 

Configure the default OAuth client in `kubesphere-config`, https://github.com/kubesphere/ks-installer/blob/master/roles/ks-core/config/templates/kubesphere-config.yaml.j2#L25-L29.

For developer, please use the basic auth for quick look.

liuan1986:
> Hard code is not a good choice. You'd better add the default OAuth client in the [configuration file](https://github.com/kubesphere/kubesphere/blob/master/config/ks-core/templates/kubesphere-config.yaml).
> 
> Here is an example: https://github.com/kubesphere/ks-installer/blob/master/roles/ks-core/config/templates/kubesphere-config.yaml.j2#L25-L29.
> 
> For developer, please use the basic auth for quick look.

Adding the default OAuth client to the configuration file is a good way. I also considered this method before pulling this request, but this will affect the compatibility of the latest version with the old version of the configuration file. When upgrading from an old version to this version, if you do not modify the configuration file, an error will be reported, because the configuration file of the previous version does not have the configuration of the default OAuth client, and the configuration file of the old version like this:
image

So I added the default OAuth client by hard coding, but if tthe default OAuth client is configured in the configuration file, it will be based on the configuration in the configuration file. The code is as follows:
image
I want to improve the reliability and compatibility of the program in this way to avoid errors caused by not modifying the configuration file after the upgrade.

There is no problem with your proposal to add the default OAuth client to the configuration file, but please consider again to see if you need to add a piece of code to ensure that the configuration is compatible? Or directly through the configuration file?

wansir:
> > Hard code is not a good choice. You'd better add the default OAuth client in the [configuration file](https://github.com/kubesphere/kubesphere/blob/master/config/ks-core/templates/kubesphere-config.yaml).
> > Here is an example: https://github.com/kubesphere/ks-installer/blob/master/roles/ks-core/config/templates/kubesphere-config.yaml.j2#L25-L29.
> > For developer, please use the basic auth for quick look.
> 
> Adding the default OAuth client to the configuration file is a good way. I also considered this method before pulling this request, but this will affect the compatibility of the latest version with the old version of the configuration file. When upgrading from an old version to this version, if you do not modify the configuration file, an error will be reported, because the configuration file of the previous version does not have the configuration of the default OAuth client, and the configuration file of the old version like this: image
> 
> So I added the default OAuth client by hard coding, but if tthe default OAuth client is configured in the configuration file, it will be based on the configuration in the configuration file. The code is as follows: image I want to improve the reliability and compatibility of the program in this way to avoid errors caused by not modifying the configuration file after the upgrade.
> 
> There is no problem with your proposal to add the default OAuth client to the configuration file, but please consider again to see if you need to add a piece of code to ensure that the configuration is compatible? Or directly through the configuration file?

Hard code is obscure, not too long ago we removed it. 
The `ks-installer` will add the default OAuth client configuration after the upgrade.

liuan1986:
Adding the default OAuth client to the configuration file~

PR #4434, opened 2021-11-11
Support query pods by status

iawia002:
### What type of PR is this?

/kind feature

This PR adds status filtering to the `kapis/resources.kubesphere.io/v1alpha3/pods` API, which now allows you to filter pods with a specific status by the `status` query parameter, eg: 

```
/kapis/resources.kubesphere.io/v1alpha3/pods?sortBy=startTime&limit=10&status=Running
```

Fixes #

```release-note
Support query pods by status.
```

/cc @zryfish

### DIFF
--- a/pkg/models/resources/v1alpha3/pod/pods.go
+++ b/pkg/models/resources/v1alpha3/pod/pods.go
@@ -28,9 +28,10 @@ import (
 )
 
 const (
-	filedNameName    = "nodeName"
-	filedPVCName     = "pvcName"
-	filedServiceName = "serviceName"
+	fieldNodeName    = "nodeName"
+	fieldPVCName     = "pvcName"
+	fieldServiceName = "serviceName"
+	fieldStatus      = "status"
 )
 
 type podsGetter struct {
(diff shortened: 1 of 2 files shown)


pkg/models/resources/v1alpha3/pod/pods.go
```diff
@@ -28,9 +28,10 @@ import (
 )
 
 const (
-	filedNameName    = "nodeName"
-	filedPVCName     = "pvcName"
-	filedServiceName = "serviceName"
+	fieldNameName    = "nodeName"
```
JohnNiang, line 31: fieldNameName -> fieldNodeName

JohnNiang (changes requested):

pkg/models/resources/v1alpha3/pod/pods.go
iawia002: Nice catch!

iawia002:
@zryfish Do we need to cherry-pick this to the release-3.2 branch?

zryfish:
Yeah, go ahead. just comment with `/cherrypick [branch name]`

PR #4446, opened 2021-11-15
Support importing Grafana templates to the workspace level.

### LINKED ISSUE #4433: [Enhance] Support importing Grafana templates to the workspace level.
Support importing Grafana templates to the workspace level. Currently, this function only can be accessed by the cluster lever. 

# Solution

See https://github.com/kubesphere/kubesphere/blob/511d0d02fffc766c0ee218dab15d399888681672/pkg/kapis/monitoring/v1alpha3/handler.go#L404 for detail. Maybe we need to add a namespace parameter as an optional to determine to convert it to Dashboard struct or not.

Related pr https://github.com/kubesphere/kubesphere/pull/4194.

/good-first-issue 
/area observability
/kind feature-request
/assign @zhu733756

zhu733756:
Signed-off-by: zhu733756 <zhu733756@kubesphere.io>

/kind feature
/kind api-change

Fixes #4433

Namespaced API paths can be described as follows:
```
#! /bin/bash
curl -d '{"grafanaDashboardUrl":"https://grafana.com/api/dashboards/7362/revisions/5/download", "description":"this is a test dashboard."}' -H "Content-Type: application/json" localhost:9090/kapis/monitoring.kubesphere.io/v1alpha3/namespaces/default/dashboards/test2/template
```

API paths for the Cluster-level resources are not changed.

```
None
```

```release-note
This feature will support importing Grafana templates to namespaced scope dashboards.
```

```docs

```


pkg/kapis/monitoring/v1alpha3/handler.go
```diff
@@ -392,57 +396,89 @@ func (h handler) handleGrafanaDashboardImport(req *restful.Request, resp *restfu
...
+		}
 
-	// create this dashboard
-	_, err = h.k.Discovery().RESTClient().
-		Post().
-		AbsPath("/apis/monitoring.kubesphere.io/v1alpha2/clusterdashboards").
-		Body(jsonDashbaord).
-		DoRaw(ctx)
+		// create this dashboard
+		err = h.rtClient.Create(ctx, &clusterdashboard)
```
junotx, line 437: If err is not empty here, it may return directly rather than perform the subsequent creation.

pkg/kapis/monitoring/v1alpha3/register.go
```diff
@@ -555,6 +557,16 @@ func AddToContainer(c *restful.Container, k8sClient kubernetes.Interface, monito
 		Returns(http.StatusOK, respOK, monitoringdashboardv1alpha2.ClusterDashboard{})).
 		Produces(restful.MIME_JSON)
 
+	ws.Route(ws.POST("/namespaces/{namespace}/dashboards/{grafanaDashboardName}/template").
+		To(h.handleGrafanaDashboardImport).
+		Doc("Convert Grafana templates to KubeSphere clusterdashboards.").
```
junotx, line 562: `clusterdashboards` in the doc should be renamed to `dashboards`

junotx (changes requested):

pkg/kapis/monitoring/v1alpha3/handler.go
zhu733756: See below it already handled the not nil case. 
```
err = h.rtClient.Get(ctx, objKey, &clusterdashboard)
// can get it will be went failed.
if err == nil {
	api.HandleBadRequest(resp, nil, errors.New("dashboards with the same name already exists."))
	return
}

err = h.rtClient.Create(ctx, &clusterdashboard)
// returns with creation failed reason if cannot create
if err != nil {
	api.HandleBadRequest(resp, nil, err)
	return
}
// if err == nil , can create it, returns to the client the completed dashboard.
resp.WriteAsJson(clusterdashboard)
```

pkg/kapis/monitoring/v1alpha3/register.go
zhu733756: will be fixed

pkg/kapis/monitoring/v1alpha3/handler.go
junotx: The err below should be the creation error. Of course, ignoring the get error causes no problem for now.

zhu733756: Yeah. The design is what the codes described.  If we return the get error, the creation process will not take an effect.

pkg/kapis/monitoring/v1alpha3/dashboard_template_test.sh
```diff
@@ -1,3 +1,4 @@
 #! /bin/bash
 
-curl -d '{"grafanaDashboardUrl":"https://grafana.com/api/dashboards/7362/revisions/5/download", "description":"this is a test dashboard."}' -H "Content-Type: application/json" localhost:9090/kapis/monitoring.kubesphere.io/v1alpha3/clusterdashboards/test1/template
\ No newline at end of file
+curl -d '{"grafanaDashboardUrl":"https://grafana.com/api/dashboards/7362/revisions/5/download", "description":"this is a test clusterdashboard."}' -H "Content-Type: application/json" localhost:9090/kapis/monitoring.kubesphere.io/v1alpha3/clusterdashboards/test1/template
+curl -d '{"grafanaDashboardUrl":"https://grafana.com/api/dashboards/7362/revisions/5/download", "description":"this is a test dashboard."}' -H "Content-Type: application/json" localhost:9090/kapis/monitoring.kubesphere.io/v1alpha3/namespaces/default/dashboards/test2/template
```
wansir, line 4: Why not use a test framework?

zhu733756: All right, need enhance to implement.

zhu733756: > Why not use a test framework?

@wansir It is hard to add this test framework. Because it is a shell script for debug.  If we want to do that,  we can add tests in repo: kubesphere/monitoring-dashboard.   This handler imports that repo.

PR #4471, opened 2021-11-24
fix groupbinding controller unittest

wansir:
### What type of PR is this?

/kind failing-test

Fix unexpected unit test failure.

Fix the failed unit test, https://prow.kubesphere.io/view/s3/prow-logs/pr-logs/pull/kubesphere_kubesphere/4455/pull-kubesphere-unit-test/1463412604385890304.

Filter out all read action before comparing.

```release-note
None
```

### DIFF
--- a/pkg/controller/groupbinding/groupbinding_controller_test.go
+++ b/pkg/controller/groupbinding/groupbinding_controller_test.go
@@ -63,8 +63,8 @@ type fixture struct {
 	fedgroupBindingLister []*fedv1beta1types.FederatedGroupBinding
 	userLister            []*v1alpha2.User
 	// Actions expected to happen on the client.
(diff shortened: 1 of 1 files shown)


RolandMa1986:
@wansir Is there a potential race condition issue？

wansir:
> @wansir Is there a potential race condition issue？

Do you mean that unit tests are parallel? We create `fixture` for each test.

RolandMa1986:
> > @wansir Is there a potential race condition issue？
> 
> Do you mean that unit tests are parallel? We create `fixture` for each test.

Ok, I think it is because the Informers was started in another goroutine. So the Watch Action will be added to the k8s actions asynchronously. So we should filter all the watch actions.

/lgtm

PR #4475, opened 2021-11-25
WIP: fix: Account password settings can have no capital letters

### LINKED ISSUE #4242: Account password settings can  have no capital letters
**Describe the Bug**
When changing the account password on the ks-console page, we must include uppercase and lowercase letters and numbers
But if we use the following command in the terminal, we can set the password without capital letters
```
kubectl patch users <user> -p '{"spec":{"password":"<password>"}}' --type='merge' && kubectl annotate users <user> iam.kubesphere.io/password-encrypted-
```
image

**Versions Used**
KubeSphere: 3.1.0
Kubernetes: 1.19.8

/kind bug

live77:
### What type of PR is this?

/kind bug

When changing the account password on the ks-console page, we must include uppercase and lowercase letters and numbers
But if we use the following command in the terminal, we can set the password without capital letters

Fixes #4242 

```
```

```release-note

```

```docs
Add a ValidatingWebhookConfiguration to check the password.
The Validator is implemented by a straight forward methods.
Not using regexp.Match because the Go regexp lib does not support backtracking. 

Valid password should contains:
- at least one capital letter,
- at least one small letter,
- at least one digit.
- and length is between 6 and 64.

This rule is written in kubesphere/console password pattern:
export const PATTERN_PASSWORD = /^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)[^]{6,64}$/
```

```release-note
Enable user password pattern validation in backends.
```


wansir:
Using [CRD validation](https://book.kubebuilder.io/reference/markers/crd-validation.html) is the best way. [Regular expression](https://stackoverflow.com/questions/31417387/regular-expression-to-find-bcrypt-hash) of the encrypted password must be included.

live77:
> Using [CRD validation](https://book.kubebuilder.io/reference/markers/crd-validation.html) is the best way. [Regular expression](https://stackoverflow.com/questions/31417387/regular-expression-to-find-bcrypt-hash) of the encrypted password must be included.

I see, CRD validation is pretty cool. Thanks for the guidance. I'll try this way.

Issue #4170, opened 2021-08-23
Edge node not only kubeege


Imy123:
**What's it about?**

In kubesphere if want use this page 
![image](https://user-images.githubusercontent.com/17736269/130405151-9730501f-3cc9-433f-90ff-a4753ff5720f.png)
I must open the kubeedge config.But kubeedge is not only, such as openyurt.

> /area edge

/kind feature-request

liangyuanpeng:
KubeSphere not support OpenYurt yet.

Issue #4153, opened 2021-08-17
Will kubesphere support SuperEdge in the near future?


nuoyaneasy:
**What's it about?**

**What's the reason why we need it?**

I believe this is an important feature for Kubesphere. There're a few use cases:

* case one
* case two
* ...

Please leave your comments below if there's anyone agrees with me. Or just give me a thumb up.

**Area Suggestion**

/kind feature-request

liangyuanpeng:
Also have CNCF project of OpenYurt, i believe that support all of these is not easy.  

I think you can join the meeting in next time for talk about this and you can see detail from https://github.com/kubesphere/community/tree/master/sig-edge .

benjaminhuo:
It depends on how many users request this feature, we'll support this if there are enough feature requests.

PR #4478, opened 2021-11-26
Convert compoent  kubeedge to edgeruntime

zhu733756:
Signed-off-by: zhu733756 <zhu733756@kubesphere.io>

### What type of PR is this?

/kind feature
/kind api-change

To support more edge frameworks as optional.  The former component kubeedge needs to be converted to EdgeRuntime. Also, the proxy service has a change:
```
edgeruntime:
      endpoint: http://edgeservice.kubeedge.svc/api/
```

See #https://github.com/kubesphere/ks-installer/pull/1829 for a detailed introduction.

```
None
```

```release-note
KubeEdge proxy service changes to http://edgeservice.kubeedge.svc/api/
```

As for console, API group changes from kubeedge.kubesphere.io to edgeruntime.kubesphere.io, no other breaks.

### DIFF
--- a/pkg/apiserver/config/config_test.go
+++ b/pkg/apiserver/config/config_test.go
@@ -38,6 +38,7 @@ import (
 	"kubesphere.io/kubesphere/pkg/simple/client/auditing"
 	"kubesphere.io/kubesphere/pkg/simple/client/cache"
 	"kubesphere.io/kubesphere/pkg/simple/client/devops/jenkins"
+	edgeruntime "kubesphere.io/kubesphere/pkg/simple/client/edgeruntime"
 	"kubesphere.io/kubesphere/pkg/simple/client/events"
 	"kubesphere.io/kubesphere/pkg/simple/client/gateway"
 	"kubesphere.io/kubesphere/pkg/simple/client/gpu"
@@ -182,6 +183,9 @@ func newTestConfig() (*Config, error) {
 		KubeEdgeOptions: &kubeedge.Options{
 			Endpoint: "http://edge-watcher.kubeedge.svc/api/",
 		},
+		EdgeRuntimeOptions: &edgeruntime.Options{
+			Endpoint: "http://edgeservice.kubeedge.svc/api/",
+		},
 		MeteringOptions: &metering.Options{
 			RetentionDay: "7d",
 		},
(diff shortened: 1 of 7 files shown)


zhu733756:
Hi, guys, after this pr merged, you can bring your own computing framework, call #4170 #4153

pkg/simple/client/edgeruntime/options.go
```diff
@@ -43,6 +43,6 @@ func (o *Options) Validate() []error {
 }
 
 func (o *Options) AddFlags(fs *pflag.FlagSet, c *Options) {
-	fs.StringVar(&o.Endpoint, "edge-watcher-endpoint", c.Endpoint,
```
zhu733756, line 46: maybe need to add a namespace flag.

pkg/apiserver/config/config_test.go
```diff
@@ -177,8 +177,8 @@ func newTestConfig() (*Config, error) {
 			IndexPrefix: "ks-logstash-auditing",
 			Version:     "6",
 		},
-		KubeEdgeOptions: &kubeedge.Options{
-			Endpoint: "http://edge-watcher.kubeedge.svc/api/",
```
zhu733756, line 181: should be configured as a namespace pramameter

zryfish:
Does this PR introduce API breaking change?

zhu733756:
> Does this PR introduce API breaking change?

Yeah, see the related link in ks-installer  #kubesphere/ks-installer#1829 .  As to console, group change from kubeedge.kubesphere.io to edgeruntime.kubesphere.io, no other breaks.

zryfish:
That's big, I suggest keeping both API groups at the moment, mark `kubeedge.kubesphere.io` as deprecated, and remove it in the later version.  Always be careful about API break changes.

zhu733756:
> That's big, I suggest keeping both API groups at the moment, mark `kubeedge.kubesphere.io` as deprecated, and remove it in the later version. Always be careful about API break changes.

All right. Will fix it by another commit later.

benjaminhuo:
Make sure you handle the deprecation well in the edgeruntime service

zhu733756:
> Make sure you handle the deprecation well in the edgeruntime service because the old API and the new API will all be proxied to the backend edge runtime service

Just make a temp fixed.

zhu733756:
> Make sure you handle the deprecation well in the edgeruntime service because the old API and the new API will all be proxied to the backend edge runtime service

Have tested the method in the restful-go framework named Deprecate(), but seems to make no effect.  See https://pkg.go.dev/github.com/emicklei/go-restful#RouteBuilder.Deprecate.

PR #4481, opened 2021-11-27
fix: Account password settings can have no capital letters

### LINKED ISSUE #4242: Account password settings can  have no capital letters

live77:
### What type of PR is this?

/kind bug

Fixes #4242

```
```

```release-note
Enable user password pattern checking. (as same as the console does)
```

```docs
Valid password(as same as the console) should contains:
- at least one capital letter,
- at least one small letter,
- at least one digit.
- and length is between 6 and 64.

This rule is written in kubesphere/console password pattern:
export const PATTERN_PASSWORD = /^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)[^]{6,64}$/

Use CRD validation pattern to implement this.

Password pattern is tricky here.
The regexp in console(javascript) is quite straightforward: ^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)[^]{6,64}$
But in Go, we don't have ?= (back tracking) capability in regexp (also in CRD validation pattern)

So we have to adopted an alternative scheme to achieve.
We use 6 different regexp to combine to achieve the same effect.

These six schemes enumerate the arrangement of numbers, uppercase letters, and lowercase letters that appear for the first time.

- ^(.*[a-z].*[A-Z].*[0-9].*)$ stands for lowercase letter comes first, then followed by an uppercase letter, then a digit.
- ^(.*[a-z].*[0-9].*[A-Z].*)$ stands for lowercase letter comes first, then followed by a digit, then an uppercase leeter.
- ^(.*[A-Z].*[a-z].*[0-9].*)$
- ^(.*[A-Z].*[0-9].*[a-z].*)$
- ^(.*[0-9].*[a-z].*[A-Z].*)$
- ^(.*[0-9].*[A-Z].*[a-z].*)$

Last but not least, the bcrypt string is also included to match the encrypted password. ^(\$2[ayb]\$.{56})$
```

### DIFF
--- a/config/crds/iam.kubesphere.io_users.yaml
+++ b/config/crds/iam.kubesphere.io_users.yaml
@@ -58,6 +58,9 @@ spec:
                 type: string
               password:
                 description: password will be encrypted by mutating admission webhook
+                maxLength: 64
+                minLength: 6
+                pattern: ^(.*[a-z].*[A-Z].*[0-9].*)$|^(.*[a-z].*[0-9].*[A-Z].*)$|^(.*[A-Z].*[a-z].*[0-9].*)$|^(.*[A-Z].*[0-9].*[a-z].*)$|^(.*[0-9].*[a-z].*[A-Z].*)$|^(.*[0-9].*[A-Z].*[a-z].*)$|^(\$2[ayb]\$.{56})$
                 type: string
             required:
             - email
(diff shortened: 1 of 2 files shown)


live77:
@wansir

zryfish:
Thanks for your contribution! 

/ok-to-test

zryfish:
@wansir Do we need to update crds of `ks-installer` ?

Issue #4486, opened 2021-11-29
NGINX Ingress controller 0.48.1 does not support Kubernetes 1.22.3


ljmict:
I installed Kubersphere 3.2.0. When I enable the project gateway, Kubesphere will create an Nginx Ingress Controller. When I check the log, the error message is as follows:
```
 -------------------------------------------------------------------------------

 NGINX Ingress controller

   Release:       v0.48.1

   Build:         30809c066cd027079cbb32dccc8a101d6fbffdcb

   Repository:    https://github.com/kubernetes/ingress-nginx

   nginx version: nginx/1.20.1

 

 -------------------------------------------------------------------------------

 

 I1129 09:18:44.077622       7 flags.go:211] "Watching for Ingress" class="nginx"

 W1129 09:18:44.077807       7 flags.go:216] Ingresses with an empty class will also be processed by this Ingress controller

 W1129 09:18:44.079148       7 client_config.go:614] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.

 I1129 09:18:44.079885       7 main.go:241] "Creating API client" host="https://10.96.0.1:443"

 I1129 09:18:44.098439       7 main.go:285] "Running in Kubernetes cluster" major="1" minor="22" git="v1.22.3" state="clean" commit="c92036820499fedefec0f847e2054d824aea6cd1" platform="linux/amd64"

 I1129 09:18:44.286102       7 main.go:105] "SSL fake certificate created" file="/etc/ingress-controller/ssl/default-fake-certificate.pem"

 I1129 09:18:44.288720       7 main.go:115] "Enabling new Ingress features available since Kubernetes v1.18"

 W1129 09:18:44.292505       7 main.go:127] No IngressClass resource with name nginx found. Only annotation will be used.

 I1129 09:18:44.358765       7 nginx.go:254] "Starting NGINX Ingress controller"

 I1129 09:18:44.408676       7 event.go:282] Event(v1.ObjectReference{Kind:"ConfigMap", Namespace:"kubesphere-controls-system", Name:"kubesphere-router-kubesphere-system", UID:"5af19877-6ca7-49cd-ad57-3c8288574c87", APIVersion:"v1", ResourceVersion:"7526978", FieldPath:""}): type: 'Normal' reason: 'CREATE' ConfigMap kubesphere-controls-system/kubesphere-router-kubesphere-system

[...]

PR #4551, opened 2021-12-17
upgrade ingress nginx version

### LINKED ISSUE #4548: Gateway deployment failure due to K8s v1.22+ api removal
**Describe the Bug**
In Kubernetes v1.22, there are some API groups completely removed, https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/#api-changes. After KubeSphere v3.2.0 release, we found there are still some compatibility issues left. In this article, https://kubesphere.com.cn/forum/d/6322-kubesphere, Gateway deployment still use `v1beta1/Ingress`, need to migrate to new API group `networking.k8s.io/v1`.

**Versions Used**
KubeSphere: v3.2.0
Kubernetes:  v1.22.1

**How To Reproduce**
1.  Enable gateway for a project.
2. Check the error log of the gateway

**Expected behavior**

/kind bug

chaunceyjiang:
### What type of PR is this?

/kind bug

ingress nginx controller deployment failure due to K8s v1.22+ api removal

Fixes #4548 #4486 

```
```

```release-note
None
```

```docs
upgrade ingress nginx version
https://github.com/kubernetes/ingress-nginx/blob/helm-chart-4.0.13/Changelog.md
```


RolandMa1986:
@chaunceyjiang It seems like the nginx-ingress 1.1.0 don't compatible with the old nginx 0.48.x. We can't simply change the downloaded helm chart version.

RolandMa1986:
@chaunceyjiang Please check the [doc](https://github.com/kubernetes/ingress-nginx/blob/main/docs/index.md#i-have-only-one-instance-of-the-ingress-nginx-controller-in-my-cluster-what-should-i-do-).  in the v1.x release. we need add the following flags to let ingress controller monitoring all ingress.

>In this case, you need to make your controller aware of the objects. If you have any Ingress objects that don't yet have either the .spec.ingressClassName field set in their manifest, or the ingress annotation (kubernetes.io/ingress.class), then you should start your Ingress-NGINX controller with the flag --watch-ingress-without-class=true.

> You can configure your Helm chart installation's values file with .controller.watchIngressWithoutClass: true.

chaunceyjiang:
Nginx Ingress - breaking change, Ingress.class now required
image

chaunceyjiang:
> @chaunceyjiang Please check the [doc](https://github.com/kubernetes/ingress-nginx/blob/main/docs/index.md#i-have-only-one-instance-of-the-ingress-nginx-controller-in-my-cluster-what-should-i-do-). in the v1.x release. we need add the following flags to let ingress controller monitoring all ingress.
> 
> > In this case, you need to make your controller aware of the objects. If you have any Ingress objects that don't yet have either the .spec.ingressClassName field set in their manifest, or the ingress annotation (kubernetes.io/ingress.class), then you should start your Ingress-NGINX controller with the flag --watch-ingress-without-class=true.
> 
> > You can configure your Helm chart installation's values file with .controller.watchIngressWithoutClass: true.

ok

chaunceyjiang:
> Nginx Ingress - breaking change, Ingress.class now required image

Turns out the fix was very simple, you just needed to ensure that your ingress annotations included the line kubernetes.io/ingress.class: "nginx" to specify the IngressClass. 

image
image

RolandMa1986:
@chaunceyjiang if users are upgradding from an older version, all the ingresses needed to be updated manually. So it would be better to add ".controller.watchIngressWithoutClass: true" to let it works like the old way.

chaunceyjiang:
> 

ok
you are right . I will add the parameter flag `--watch-ingress-without-class=true`

chaunceyjiang:
> > @chaunceyjiang if users are upgradding from an older version, all the ingresses needed to be updated manually. So it would be better to add ".controller.watchIngressWithoutClass: true" to let it works like the old way.
> 
> ok you are right . I will add the parameter flag `--watch-ingress-without-class=true`

image

image

chaunceyjiang:
so `kubesphere/nginx-ingress-controller:v1.1.0` is required

config/watches.yaml
```diff
@@ -3,6 +3,8 @@
   version: v1alpha1
   kind: Nginx
   chart: /var/helm-charts/ingress-nginx
+  overrideValues:
```
RolandMa1986, line 6: I would recommend changing the Nginx CR directly, the file located under:
https://github.com/kubesphere/kubesphere/blob/master/config/gateway/templates/nginx-ingress.yaml
The  OverrideValues are used when overriding values with environment variables.

chaunceyjiang:
image
image

RolandMa1986:
/lgtm
@chaunceyjiang btw, The docker image version will be overridden by the ks-installer:
https://github.com/kubesphere/ks-installer/blob/master/roles/download/defaults/main.yml

chaunceyjiang:
> /lgtm @chaunceyjiang btw, The docker image version will be overridden by the ks-installer: https://github.com/kubesphere/ks-installer/blob/master/roles/download/defaults/main.yml

ok

https://github.com/kubesphere/ks-installer/pull/1872

Issue #4426, opened 2021-11-08
[Feature] Support change member cluster kubeconfig 


zryfish:
**What's it about?**
Multicluster is widely used since we introduced it from version v3.0. There is one problem with these member clusters,  there is no way to replace expired `kubeconfig`. You can use the following command to check expiration date of member cluster kubeconfig:
```
kubectl get cluster [cluster name] -o jsonpath='{.spec.connection.kubeconfig}' | base64 -d | grep "client-certificate-data" | awk '{print $2}' | base64 -d | openssl x509 -enddate -noout -in -
```
After the expiration date, the user either needs to unbind cluster and rejoin cluster or manually change `kubeconfig` in cluster cr. Both are inconvenient and complicated. 

**What's the reason why we need it?**

Users can replace member cluster kubeconfig through UI wich is less risky.
/area multicluster
/kind feature-request

/cc @kubesphere/sig-multicluster

zryfish:
Please take a look at this @iawia002

zryfish:
We may also need to find a way to notify users when cluster kubeconfig approaching the expiration date.

zryfish:
@iawia002  Any update on this?

iawia002:
> @iawia002 Any update on this?

I will submit a PR in a few days to add the corresponding API

iawia002:
> We may also need to find a way to notify users when cluster kubeconfig approaching the expiration date.

If we want to achieve this, we may need to add the openssl packages to the base image of ks-apiserver. Is this acceptable?

zryfish:
Just go ahead, we can review your solution once it's more concrete.

zryfish:
/kind design
/area console

@iawia002 Could you add some documentation about the new API?

iawia002:
> @iawia002 Could you add some documentation about the new API?

Yes, where should I put this documentation?

zryfish:
Need to validate the old cluster with the new kubeconfig.

VioZhang:
Is it would be better that we can also support uploading the yaml of kubeconfig and the token ，not only the text？

wenxin-01:
Issue fixed in `v3.3.0-rc.0`

PR #4562, opened 2021-12-21
Add update cluster kubeconfig API

iawia002:
### What type of PR is this?

/kind feature

Part of #4426

This patch adds a new API that allows users to update the member cluster's kubeconfig:

```
PUT /kapis/cluster.kubesphere.io/v1alpha1/clusters/{name}/kubeconfig

{
  "kubeconfig": "base64 encoded data"
}
```

This API will only allow to update member clusters that use direct mode, for member clusters that use proxy mode, we don't need to update the kubeconfig, if the certs expired, just restart the tower component in the host cluster, it will renew the cert.

Fixes #

The API to detect whether the cluster kubeconfig certificate has expired will be submitted in the next PR.

```release-note
Add update cluster kubeconfig API.
```

```docs

```

/cc @zryfish

### DIFF
--- a/pkg/kapis/cluster/v1alpha1/handler.go
+++ b/pkg/kapis/cluster/v1alpha1/handler.go
@@ -46,7 +46,9 @@ import (
 	"kubesphere.io/api/cluster/v1alpha1"
 
 	"kubesphere.io/kubesphere/pkg/api"
+	clusterv1alpha1 "kubesphere.io/kubesphere/pkg/api/cluster/v1alpha1"
 	"kubesphere.io/kubesphere/pkg/apiserver/config"
+	kubesphere "kubesphere.io/kubesphere/pkg/client/clientset/versioned"
 	"kubesphere.io/kubesphere/pkg/client/informers/externalversions"
 	clusterlister "kubesphere.io/kubesphere/pkg/client/listers/cluster/v1alpha1"
 	"kubesphere.io/kubesphere/pkg/version"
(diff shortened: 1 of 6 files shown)


pkg/kapis/cluster/v1alpha1/handler.go
```diff
@@ -251,6 +254,73 @@ func (h *handler) generateDefaultDeployment(cluster *v1alpha1.Cluster, w io.Writ
...
+	cluster := obj.DeepCopy()
+	if _, ok := cluster.Labels[v1alpha1.HostCluster]; ok {
+		api.HandleBadRequest(response, request, fmt.Errorf("update kubeconfig of the host cluster is not allowed"))
+		return
+	}
+	// For member clusters that use proxy mode, we don't need to update the kubeconfig,
+	// if the certs expired, just restart the tower component in the host cluster, it will renew the cert.
+	if cluster.Spec.Connection.Type == v1alpha1.ConnectionTypeProxy {
+		api.HandleBadRequest(response, request, fmt.Errorf(
+			"update kubeconfig of member clusters that use proxy mode is not allowed, if the certs expired, just restart the tower component in the host cluster, it will renew the cert",
```
zryfish, line 280: Maybe we should automatically renew proxy mod cluster certs in tower, not by restarting tower, it's risky. 
```suggestion
			"update kubeconfig of member clusters which using proxy mode is not allowed, their certs are managed and will be renewed by tower.",
```

iawia002: Sure

Issue #4569, opened 2021-12-24
Add a function to login to the edge host in the kubesphere edge node interface


lynxcat:
**What's it about?**

Add a function to log in to the edge host in the kubesphere edge node interface

**What's the reason why we need it?**

This function can make it easier for us to manage the edge host, such as
* Download Image before updating edge application
* Update edgecore or docker at the edge
* Modify the edge host configuration
* ...

** Solution **
- Option One：

> - 1.Add a ssh-client to this image [kubectl](https://github.com/kubesphere/kubectl)
> - 2.Use edgemesh's ssh proxy tunnel to forward traffic to the edge
- Option Two：
> - 1.Run a container on the edge to mount the root directory of the host into the container, and then use chroot to manage the host
> - 2.Use kubectl exec to connect to the container of the corresponding edge node

The Option One is relatively simple to implement.We only need to update the kubeedge version, then turn on edgemesh, and make a few changes to the kubesphere front-end code.
like this:
![1fdb3543285ff5732fed38c5f86ba53](https://user-images.githubusercontent.com/2222701/147314230-25f00378-6db6-4008-b395-49dccb7087b1.png)

The Option Two  will not need to change the code but will occupy more resources on the edge, and the security will be worse. Because kubectl exec has no security mechanism, but ssh and edgemesh do some security measures

Please leave your comments below if there's anyone agrees with me. Or just give me a thumb up.

**Area Suggestion**

/area edge

/kind feature-request

zhu733756:
It seems that option One will rely on the kubeedge edgemesh, option two can save one container resource for edge measurement, and more safety?

zryfish:
I found an article addressing this issue https://alexei-led.github.io/post/k8s_node_shell/, I think it'll help a lot. IMO, the second option is more Kuberenetes and general could be used on any node of the cluster not only the Kubeedge node. 

Also, quote from the article above
> But this approach still requires from you to manage access to the bastion servers and protect SSH keys. IMHO, managing supporting SSH infrastructure, is a high price to pay.

lynxcat:
> I found an article addressing this issue https://alexei-led.github.io/post/k8s_node_shell/, I think it'll help a lot. IMO, the second option is more Kuberenetes and general could be used on any node of the cluster not only the Kubeedge node.
> 
> Also, quote from the article above
> 
[...]

PR #4579, opened 2021-12-28
Add a function to shell access to the node in the kubesphere

lynxcat:
### What type of PR is this?

/kind documentation
/kind feature

### Why we need it:

Add a function to shell access to the node in the kubesphere

see #4569 
```release-note
Add a function to shell access to the node in the kubesphere.
```
[the proposal](https://github.com/lynxcat/kubesphere/blob/master/docs/proposal/shell-access-to-node-proposal.md)


docs/proposal/shell-access-to-node-proposal.md
```diff
@@ -0,0 +1,53 @@
...
+
+At first, I planned to use edgemesh's SSH tunnel proxy, but it can't work on k8s cluster nodes.
+Later, at the [suggestion](https://github.com/kubesphere/kubesphere/issues/4569) of the community I use nsenter to do this. 
+The nsenter is a small program from util-linux package, that can run program with namespaces (and cgroups) of other processes. Exactly what we need!
+
+## Workflow
+![shell-access-to-node](../images/shell-access-to-node.png)
+
+**step0:**  
+The client initiates a terminal link request to the api-server
```
zhu733756, line 20: You mean ks-apiserver?

pkg/models/terminal/terminal.go
```diff
@@ -224,3 +233,84 @@ func (t *terminaler) HandleSession(shell, namespace, podName, containerName stri
...
+							Privileged: &privileged,
+						},
+					},
+				},
+			},
+		}
+
+		pod, err = t.client.CoreV1().Pods(namespace).Create(context.Background(), p, metav1.CreateOptions{})
+		if err != nil {
+			klog.Warning(err)
```
zhu733756, line 276: Please add a prefix description here, like "xx create pod failed: " + err.Error()

pkg/models/terminal/terminal.go
```diff
@@ -224,3 +233,84 @@ func (t *terminaler) HandleSession(shell, namespace, podName, containerName stri
...
+	t.WatchPodStatusBeRunning(pod, func() {
+		mu.Lock()
+		if _, ok := NodeSessionCounter[nodename]; !ok {
+			var index int64 = 0
+			NodeSessionCounter[nodename] = &index
+		}
+		mu.Unlock()
+
+		atomic.AddInt64(NodeSessionCounter[nodename], 1)
+		t.HandleSession(shell, namespace, podName, containerName, conn)
```
zhu733756, line 290: Do we need to add a timeout logic to each session?

pkg/models/terminal/terminal.go
```diff
@@ -224,3 +233,84 @@ func (t *terminaler) HandleSession(shell, namespace, podName, containerName stri
...
+
+	t.WatchPodStatusBeRunning(pod, func() {
+		mu.Lock()
+		if _, ok := NodeSessionCounter[nodename]; !ok {
+			var index int64 = 0
+			NodeSessionCounter[nodename] = &index
+		}
+		mu.Unlock()
+
+		atomic.AddInt64(NodeSessionCounter[nodename], 1)
```
zhu733756, line 289: Would NodeSessionCounter be safe here?  Try to use value of NodeSessionCounter[nodename] to addint64.

Use "idx := atomic.LoadInt64(NodeSessionCounter[nodename])" in the lock logic.

pkg/models/terminal/terminal.go
```diff
@@ -224,3 +233,84 @@ func (t *terminaler) HandleSession(shell, namespace, podName, containerName stri
...
+		}
+
+		pod, err = t.client.CoreV1().Pods(namespace).Create(context.Background(), p, metav1.CreateOptions{})
+		if err != nil {
+			klog.Warning(err)
+			return
+		}
+	}
+
+	t.WatchPodStatusBeRunning(pod, func() {
```
zhu733756, line 281: I recommend here to use goroutine to watch, and the main process could be added to select channel signal for it,

zryfish:
@lynxcat Please submit your proposal to https://github.com/kubesphere/community, leave only code in your PR.

pkg/kapis/terminal/v1alpha2/register.go
```diff
@@ -50,6 +50,14 @@ func AddToContainer(c *restful.Container, client kubernetes.Interface, authorize
 		Metadata(restfulspec.KeyOpenAPITags, []string{constants.TerminalTag}).
 		Writes(models.PodInfo{}))
 
+	//Add new Route to support shell access to the node
+	webservice.Route(webservice.GET("/nodename/{nodename}/exec").
```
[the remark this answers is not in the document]
lynxcat: ok, i'll modify it

pkg/models/terminal/terminal.go
```diff
@@ -53,6 +57,10 @@ type TerminalSession struct {
 	sizeChan chan remotecommand.TerminalSize
 }
 
+var NodeSessionCounter = make(map[string]*int64)
```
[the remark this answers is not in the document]
lynxcat: thanks, i'll modify it and use sync.Once to init

lynxcat:
@zryfish OK, I'll delete proposal doc in this PR. Which directory should I put the proposal in community repo

pkg/models/terminal/terminal.go
lynxcat: because I can't detect whether the session is active, it is hard to add timeout logic. If this is necessary, an active detection mechanism needs to be added to complete this function.

pkg/models/terminal/terminal.go
lynxcat: ok, i'll modify it

pkg/models/terminal/terminal.go
lynxcat: thanks, i'll and it

pkg/kapis/terminal/v1alpha2/handler.go
```diff
@@ -89,3 +89,38 @@ func (t *terminalHandler) handleTerminalSession(request *restful.Request, respon
 
 	t.terminaler.HandleSession(shell, namespace, podName, containerName, conn)
 }
+
+func (t *terminalHandler) handleShellAccessToNode(request *restful.Request, response *restful.Response) {
```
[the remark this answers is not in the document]
lynxcat: i think this method don't need to be exposed. if we want to use it in another package, the method terminaler.HandleShellAccessToNode will be better

pkg/models/terminal/terminal.go
zhu733756: > because I can't detect whether the session is active, it is hard to add timeout logic. If this is necessary, an active detection mechanism needs to be added to complete this function.

OK. Let's talk about this topic tomorrow online. And welcome to share!

pkg/models/terminal/terminal.go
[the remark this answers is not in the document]
zhu733756: Also add a cap to this map, since the nodes can be listed.  We can get the max nodes num.

pkg/models/terminal/terminal.go
zhu733756: > > because I can't detect whether the session is active, it is hard to add timeout logic. If this is necessary, an active detection mechanism needs to be added to complete this function.
> 
> OK. Let's talk about this topic tomorrow online. And welcome to share!

Maybe we need to list all nodes and pre-set the nodename to this map ?

pkg/models/terminal/terminal.go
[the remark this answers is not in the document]
zhu733756: > Also add a cap to this map, since the nodes can be listed. We can get the max nodes num.

Maybe we need to list all nodes and pre-set the nodename to this map ?

zryfish:
> @lynxcat Please submit your proposal to https://github.com/kubesphere/community, leave only code in your PR.

I think we can put it into `sig/edge` https://github.com/kubesphere/community/tree/master/sig-edge now.

pkg/models/terminal/terminal.go
lynxcat: OK, but when considering this function, we also need to consider the FE code. Because when the connection is disconnected, the FE will be automatically reconnected

pkg/models/terminal/terminal.go
```diff
@@ -140,15 +151,135 @@ func (t TerminalSession) Close(status uint32, reason string) {
...
+	node, err := n.client.CoreV1().Nodes().Get(context.Background(), n.Nodename, metav1.GetOptions{})
+
+	if err != nil {
+		return n, fmt.Errorf("node cannot exist. nodename:%s, err: %v", n.Nodename, err)
+	}
+
+	flag := false
+	for _, condition := range node.Status.Conditions {
+		if condition.Type == v1.NodeReady && condition.Status == v1.ConditionTrue {
+			flag = true
```
zhu733756, line 200: Need a break here? Even if it doesn't seem to affect anything.

pkg/models/terminal/terminal.go
```diff
@@ -224,3 +355,49 @@ func (t *terminaler) HandleSession(shell, namespace, podName, containerName stri
 
 	session.Close(1, "Process exited")
 }
+
+func (t *terminaler) HandleShellAccessToNode(nodename string, conn *websocket.Conn) {
+	ch := make(chan bool, 1)
+	once.Do(func() {
```
zhu733756, line 361: once.Do had better add to func init section.

zhu733756, line 361: once.Do had better add to func init section.

pkg/models/terminal/terminal.go
```diff
@@ -224,3 +355,49 @@ func (t *terminaler) HandleSession(shell, namespace, podName, containerName stri
...
+		if pod.Status.Phase == v1.PodRunning {
+			atomic.AddInt64(NodeSessionCounter[n.Nodename], 1)
+			ch <- true
+			break
+		}
+
+		pod, err = n.client.CoreV1().Pods(pod.ObjectMeta.Namespace).Get(context.Background(), pod.ObjectMeta.Name, metav1.GetOptions{})
+		if err != nil {
+			klog.Warning(err)
+			ch <- false
```
zhu733756, line 399: You can use close(ch) to close the unbuffered channel and notify the main process. Since this practice is common.

pkg/models/terminal/terminal.go
```diff
@@ -140,15 +151,135 @@ func (t TerminalSession) Close(status uint32, reason string) {
...
+		}
+	}
+
+	return pod, nil
 }
 
-func NewTerminaler(client kubernetes.Interface, config *rest.Config) Interface {
-	return &terminaler{client: client, config: config}
+func (n NodeTerminaler) CleanUpNSEnterPod() {
+	atomic.AddInt64(NodeSessionCounter[n.Nodename], -1)
```
zhu733756, line 275: If at one second before,  another session comes in and adds mutex lock in its session, do many sessions be blocked to wait for that session? So do we need to weaken the lock to write/read lock?  Or have you tested the sync.Map?

pkg/kapis/terminal/v1alpha2/handler.go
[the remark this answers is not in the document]
zhu733756: OK

zhu733756:
Totally LGTM. See my comments above. /cc

pkg/models/terminal/terminal.go
lynxcat: Only a simple variable assignment is made in the lock, which will not take too long. I think there shouldn't cause a long wait

pkg/models/terminal/terminal.go
lynxcat: if I use close(ch) here, I cannot perform error detection.

pkg/models/terminal/terminal.go
lynxcat: thanks, i'll modify it

pkg/models/terminal/terminal.go
lynxcat: If there is no shell access request, this variable will not be used, so I put it here

zhu733756: > If there is no shell access request, this variable will not be used, so I put it here

I think one variable doesn't take up much resources.

pkg/models/terminal/terminal.go
zhu733756: OK. IMO, I think sync.Map will improve the read performance. You can have a try.

pkg/models/terminal/terminal.go
zhu733756: OK. Adding two unbuffered channels is another way to solve this case.

pkg/models/terminal/terminal.go
zhu733756: > OK. IMO, I think sync.Map will improve the read performance. You can have a try.

Sync.map is thread-safe, with constant time complexity for reads, inserts, and deletions.  You can remove lock here.

lynxcat: OK.I'll modify it

pkg/models/terminal/terminal.go
lynxcat: OK.I'll modify it

pkg/models/terminal/terminal.go
lynxcat: OK.I'll move it to init section

pkg/models/terminal/terminal.go
zhu733756: > OK.I'll modify it

No need to change. If it can work properly, let it go.

pkg/models/terminal/terminal.go
```diff
@@ -224,3 +351,48 @@ func (t *terminaler) HandleSession(shell, namespace, podName, containerName stri
...
+func (n *NodeTerminaler) WatchPodStatusBeRunning(pod *v1.Pod, succ chan bool, fail chan bool) {
+	var err error
+	for i := 0; i < 5; i++ {
+		if pod.Status.Phase == v1.PodRunning {
+			idx, _ := NodeSessionCounter.Load(n.Nodename)
+			atomic.AddInt64(idx.(*int64), 1)
+			close(succ)
+			return
+		}
+		time.Sleep(time.Second)
```
zhu733756, line 389: I think time.Sleep(time.Second) move to the end would be better.

pkg/models/terminal/terminal.go
```diff
@@ -140,15 +149,133 @@ func (t TerminalSession) Close(status uint32, reason string) {
...
+		Shell:         "sh",
+		Privileged:    true,
+		Config:        options,
+		client:        client,
+	}
+
+	node, err := n.client.CoreV1().Nodes().Get(context.Background(), n.Nodename, metav1.GetOptions{})
+
+	if err != nil {
+		return n, fmt.Errorf("node cannot exist. nodename:%s, err: %v", n.Nodename, err)
```
zryfish, line 192: `Nodes().Get()` returns err doesn't always means `node not exist`, sometimes it's network issue or internal error, so I suggest change the error message to
```suggestion
		return n, fmt.Errorf("getting node error, nodename:%s, err: %v", n.Nodename, err)
```

pkg/models/terminal/terminal.go
zryfish: I think we need to filter out unschedulable nodes, if user cordon nodes, it's unlikely to create pod on these nodes.

pkg/models/terminal/terminal.go
```diff
@@ -140,15 +149,133 @@ func (t TerminalSession) Close(status uint32, reason string) {
...
+			flag = true
+			break
+		}
+	}
+	if !flag {
+		return n, fmt.Errorf("node status error. node: %s", n.Nodename)
+	}
+
+	idx := int64(0)
+	NodeSessionCounter.LoadOrStore(nodename, &idx)
```
zryfish, line 207: Correct me if I am wrong. `NodeSessionCounter` is used to store user sessions, if there is more than one session attached to the exec pod,  we don't clean the exec pod. Since `NodeSessionCounter` is stored in memory, isn't data lost between ks-apiserver restarts?

pkg/models/terminal/terminal.go
```diff
@@ -224,3 +351,48 @@ func (t *terminaler) HandleSession(shell, namespace, podName, containerName stri
...
+		return
+	}
+
+	pod, err := nodeTerminaler.getNSEnterPod()
+	if err != nil {
+		klog.Warning("get nsenter pod error: ", err)
+		return
+	}
+
+	go nodeTerminaler.WatchPodStatusBeRunning(pod, succ, fail)
```
zryfish, line 370: there is a package in k8s `https://pkg.go.dev/k8s.io/apimachinery@v0.21.4/pkg/util/wait`, I think we can use `wait.Poll`.

pkg/models/terminal/terminal.go
zhu733756: Maybe need to check the pod on this node is not existed before setting it to zero.

pkg/models/terminal/terminal.go
```diff
@@ -378,21 +379,33 @@ func (t *terminaler) HandleShellAccessToNode(nodename string, conn *websocket.Co
...
+			atomic.AddInt64(idx.(*int64), 1)
+			return true, nil
 		}
+
+		return false, nil
+	})
+
+	if err != nil {
+		klog.Warning("watching pod status error: ", err)
+		close(fail)
```
zhu733756, line 407: Last time, I recommend using two channels and one goroutine to select signal in order to consider adding `time.After` to do timeout logic to cancel the context,  so far I think these two channels and goroutine can be removed, we can directly call the following logic here, the codes would be more clear and simple:
```
t.HandleSession(nodeTerminaler.Shell, nodeTerminaler.Namespace, nodeTerminaler.PodName, nodeTerminaler.ContainerName, conn)
		defer nodeTerminaler.CleanUpNSEnterPod()
```
What do you think? @lynxcat @zryfish

pkg/models/terminal/terminal.go
```diff
@@ -378,21 +379,33 @@ func (t *terminaler) HandleShellAccessToNode(nodename string, conn *websocket.Co
...
+	} else {
+		close(succ)
```
zhu733756, line 409: Last time, I recommend using two channels and one goroutine to select signal in order to consider adding `time.After` to do timeout logic to cancel the context,  so far I think these two channels and goroutine can be removed, we can directly call the following logic here, the codes would be more clear and simple:
```
t.HandleSession(nodeTerminaler.Shell, nodeTerminaler.Namespace, nodeTerminaler.PodName, nodeTerminaler.ContainerName, conn)
defer nodeTerminaler.CleanUpNSEnterPod()
```
What do you think? @lynxcat @zryfish

lynxcat: I agree with you.

pkg/models/terminal/terminal.go
zhu733756: Does this logic can be removed ?  And combines the initial logic in here:
```
idx, _ := NodeSessionCounter.Load(n.Nodename)
atomic.AddInt64(idx.(*int64), 1)
close(succ)
```
If we load successfully, let's add one to the value.  Or we directly set it to 1. This will save once loadorstore in memory.
What do you think @lynxcat @zryfish

pkg/models/terminal/terminal.go
```diff
@@ -224,3 +349,60 @@ func (t *terminaler) HandleSession(shell, namespace, podName, containerName stri
 
 	session.Close(1, "Process exited")
 }
+
+func (t *terminaler) HandleShellAccessToNode(nodename string, conn *websocket.Conn) {
+
```
wansir, line 354: https://github.com/kubesphere/kubesphere/blob/f0210193c1e9d986690e3bd8835ca444e401688d/pkg/kapis/terminal/v1alpha2/handler.go#L61-L82

Check permission to create pod/exec is required. Use the WebSocket protocol, only the GET verb will be received before the upgrade.

It's similar to `pods/exec`, `kubectl` using `SPDY/3.1` protocol connect to the pod, we grant pod exec permission by RBAC rule: `create pods/exec`

lynxcat: I checked the permissions in ```terminalHandler.handleShellAccessToNode()```, same as you quoted. if the current user doesn't have correct permission, he can't enter this function.

pkg/kapis/terminal/v1alpha2/handler.go
```diff
@@ -89,3 +89,38 @@ func (t *terminalHandler) handleTerminalSession(request *restful.Request, respon
...
+
+	createPodsExec := authorizer.AttributesRecord{
+		User:            user,
+		Verb:            "create",
+		Resource:        "pods",
+		Subresource:     "exec",
+		Namespace:       "kubesphere-controls-system",
+		ResourceRequest: true,
+		ResourceScope:   requestctx.NamespaceScope,
+	}
```
wansir, line 106:
```suggestion
	createNodesExec := authorizer.AttributesRecord{
		User:            user,
		Verb:            "create",
		Resource:        "nodes",
		Subresource:     "exec",
		ResourceRequest: true,
		ResourceScope:   requestctx.ClusterScope,
	}
```

According to `/kapis/terminal.kubesphere.io/v1alpha2/nodes/{node}/exec`.

PR #4582, opened 2021-12-30
replace the kubespheredev into kubesphere

hurricane1988:

Fixes #

```
```

```release-note

```

```docs

```

### DIFF
--- a/hack/deploy-kubesphere.sh
+++ b/hack/deploy-kubesphere.sh
@@ -30,9 +30,9 @@ function wait_for_installation_finish() {
     done < <(timeout 900 kubectl logs -n kubesphere-system deploy/ks-installer -f)
 }
 
-# Use kubespheredev and latest tag as default image
+# Use kubesphere and latest tag as default image
 TAG="${TAG:-latest}"
-REPO="${REPO:-kubespheredev}"
+REPO="${REPO:-kubesphere}"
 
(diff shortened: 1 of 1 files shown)


FeynmanZhou:
Have you raised a GitHub issue for this PR? If already, please help to link it in this PR, thanks.

JohnNiang (commented):
Hi @hurricane1988 , thanks for your contribution, but I left two question for you:

1. What was the problem you encountered? Is there any reference?
1. How did you confirm the root cause is it?

If you are resolving the problem in https://kubesphere.com.cn/forum/d/6428-v321 [In Chinese], I would like to say that the last change of those lines was 10 months ago, so I guess it is not the root cause of your problem.

hurricane1988:
Hi@JohnNiang the details as below:
1. when I upgraded from 3.1.1 to 3.2.1 used offline method, I found the image will be auto added "kubespheredev"
2. I find key word "kubespheredev" in ks-installer repo, and replace it with "kubesphere"
3. I'am very  interested in kubesphere project, and hope to be a contributor.

zryfish:
@hurricane1988 We use `hack/deploy-kubesphere.sh` to deploy a development KubeSphere cluster, all dev images are pushed to repo `kubespheredev`, so it's not a mistake, it's intentional. Can you provide more details about the problem you have?

PR #4584, opened 2021-12-30
Sync the expiration time of kubeconfig cert file of the cluster

iawia002:
### What type of PR is this?

/kind feature

Part of #4426

This patch adds a new status condition `KubeConfigCertExpiresInSevenDays` for a cluster:

```yaml
status:
  conditions:
  - lastTransitionTime: "2022-01-05T08:46:14Z"
    lastUpdateTime: "2022-01-05T08:46:14Z"
    message: 2022-10-25 08:24:09 +0000 UTC
    reason: KubeConfigCertExpiresInSevenDays
    status: "False"
    type: KubeConfigCertExpiresInSevenDays
  - lastTransitionTime: "2022-01-05T08:51:47Z"
    lastUpdateTime: "2022-01-05T08:51:47Z"
    message: Cluster has joined federation control plane successfully
    status: "True"
    type: Federated
  - lastTransitionTime: "2022-01-05T08:51:47Z"
    lastUpdateTime: "2022-01-05T08:51:47Z"
    message: Cluster is available now
    reason: Ready
    status: "True"
    type: Ready
```

The front end can determine whether the cluster certificate is about to expire based on the `KubeConfigCertExpiresInSevenDays` condition status, and time in `message` is the expiration time, we can show the certificate expiration note on the page.

Fixes #

```
```

```release-note
Sync the expiration time of kubeconfig cert file of the cluster.
```

```docs

```

/cc @zryfish

### DIFF
--- a/pkg/controller/cluster/cluster_controller.go
+++ b/pkg/controller/cluster/cluster_controller.go
@@ -19,7 +19,9 @@ package cluster
 import (
 	"bytes"
 	"context"
+	"crypto/x509"
 	"encoding/json"
+	"encoding/pem"
 	"fmt"
 	"net/http"
 	"reflect"
@@ -52,6 +54,7 @@ import (
 	clusterclient "kubesphere.io/kubesphere/pkg/client/clientset/versioned/typed/cluster/v1alpha1"
 	clusterinformer "kubesphere.io/kubesphere/pkg/client/informers/externalversions/cluster/v1alpha1"
 	clusterlister "kubesphere.io/kubesphere/pkg/client/listers/cluster/v1alpha1"
+	"kubesphere.io/kubesphere/pkg/utils/k8sutil"
 	"kubesphere.io/kubesphere/pkg/version"
 )
 
@@ -614,6 +617,11 @@ func (c *clusterController) syncCluster(key string) error {
 	}
 	c.updateClusterCondition(cluster, readyConditon)
 
+	if err = c.updateKubeConfigExpirationDateCondition(cluster); err != nil {
+		klog.Errorf("sync KubeConfig expiration date for cluster %s failed: %v", cluster.Name, err)
+		return err
+	}
+
 	if !reflect.DeepEqual(oldCluster, cluster) {
 		_, err = c.clusterClient.Update(context.TODO(), cluster, metav1.UpdateOptions{})
 		if err != nil {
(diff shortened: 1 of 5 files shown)


zryfish:
I think it will be more user-friendly if we check expiration in the controller and update it into cluster status.

pkg/kapis/cluster/v1alpha1/handler.go
```diff
@@ -321,6 +322,49 @@ func (h *handler) updateKubeConfig(request *restful.Request, response *restful.R
...
+	if err != nil {
+		api.HandleBadRequest(response, request, err)
+		return
+	}
+	if config.CertData == nil {
+		api.HandleBadRequest(response, request, fmt.Errorf("no cert data"))
+		return
+	}
+
+	cmd := exec.Command("openssl", "x509", "-enddate", "-noout")
```
zryfish, line 342: why not use the go package `crypto/tls`? I think that's more convenient.

iawia002:
> I think it will be more user-friendly if we check expiration in the controller and update it into cluster status.

Reasonable, I will update it

pkg/kapis/cluster/v1alpha1/handler.go
iawia002: I didn't know this package could do that, I'll do some research

zryfish: My bad, I mean `https://pkg.go.dev/crypto/x509`, K8s use it to generate kubeconfig, so we do in tower `https://github.com/kubesphere/tower/blob/master/pkg/certs/cert.go`.

pkg/controller/cluster/cluster_controller.go
```diff
@@ -232,6 +235,9 @@ func (c *clusterController) Run(workers int, stopCh <-chan struct{}) error {
 
 	}, c.resyncPeriod, stopCh)
 
+	// sync cluster's KubeConfig cert expiration date per day.
+	go wait.Until(c.syncKubeConfigExpirationDate, time.Hour*24, stopCh)
```
zryfish, line 239: If we check the certificate every 24 hrs, what if the user renewed kubeconfig, the condition of the cluster will not be updated immediately.

pkg/controller/cluster/cluster_controller.go
```diff
@@ -823,3 +829,69 @@ func (c *clusterController) unJoinFederation(clusterConfig *rest.Config, unjoini
...
+	for _, cluster := range clusters {
+		if _, ok := cluster.Labels[clusterv1alpha1.HostCluster]; ok {
+			continue
+		}
+		// we don't need to check member clusters which using proxy mode, their certs are managed and will be renewed by tower.
+		if cluster.Spec.Connection.Type == clusterv1alpha1.ConnectionTypeProxy {
+			continue
+		}
+
+		klog.Infof("sync KubeConfig expiration date for cluster %s", cluster.Name)
```
zryfish, line 867: infof with level.

zryfish:
Overall of this PR looks promising to me, just a few comments.

pkg/controller/cluster/cluster_controller.go
iawia002: Nice question, I wanted to sync every time I reconcile the cluster, but would it be too frequent?

> the condition of the cluster will not be updated immediately.

If we want to solve this problem perfectly, we need to sync the expiration time every time we reconcile the cluster, WDYT? Any other ideas?

zryfish: I think it is ok since checking kubeconfig expiration is not time-consuming.

iawia002: Updated, PTAL.

pkg/controller/cluster/cluster_controller.go
iawia002: Updated.

iawia002:
All comments addressed, PTAL @zryfish

zryfish:
@iawia002 Please create a feature-request for `kubesphere/console` to add cert expiration warning to the clusters

iawia002:
> @iawia002 Please create a feature-request for `kubesphere/console` to add cert expiration warning to the clusters

Done, https://github.com/kubesphere/console/issues/2898

Issue #4357, opened 2021-10-13
ks-apiserver may crash because of concurrent map read and map write


xushuxun:
**Describe the Bug**
ks-apiserver may crash

Here is the panic goroutine stack:
```
fatal error: concurrent map read and map write

goroutine 1764 [running]:
runtime.throw(0x34a39a6, 0x21)
        /usr/local/go/src/runtime/panic.go:1117 +0x72 fp=0xc001e3e218 sp=0xc001e3e1e8 pc=0x438072
runtime.mapaccess2(0x2f82040, 0xc0015bc660, 0xc000a25410, 0xc000a25410, 0xc002a802df)
        /usr/local/go/src/runtime/map.go:469 +0x255 fp=0xc001e3e258 sp=0xc001e3e218 pc=0x40f675
reflect.mapaccess(0x2f82040, 0xc0015bc660, 0xc000a25410, 0x3485d1b)
        /usr/local/go/src/runtime/map.go:1318 +0x3f fp=0xc001e3e290 sp=0xc001e3e258 pc=0x46837f
reflect.Value.MapIndex(0x2f82040, 0xc0000383b8, 0x195, 0x2e41080, 0xc000a25410, 0x98, 0x2e41080, 0xc000a254e0, 0x98)
        /usr/local/go/src/reflect/value.go:1189 +0x16e fp=0xc001e3e308 sp=0xc001e3e290 pc=0x4a02ee
encoding/json.mapEncoder.encode(0x3580f88, 0xc002050700, 0x2f82040, 0xc0000383b8, 0x195, 0x100)
        /usr/local/go/src/encoding/json/encode.go:813 +0x317 fp=0xc001e3e4b0 sp=0xc001e3e308 pc=0x5d1ff7
encoding/json.mapEncoder.encode-fm(0xc002050700, 0x2f82040, 0xc0000383b8, 0x195, 0x100)
        /usr/local/go/src/encoding/json/encode.go:779 +0x6d fp=0xc001e3e4f0 sp=0xc001e3e4b0 pc=0x5de74d
encoding/json.structEncoder.encode(0xc001c4e000, 0x10, 0x10, 0xc00225b4a0, 0xc002050700, 0x33ac260, 0xc000038320, 0x199, 0x100)
        /usr/local/go/src/encoding/json/encode.go:761 +0x2ab fp=0xc001e3e588 sp=0xc001e3e4f0 pc=0x5d1a8b
[...]

Issue #3469, opened 2021-03-15
Concurrent map writes cause panic


zryfish:
**Describe the Bug**
```
fatal error: concurrent map writes

goroutine 942 [running]:
runtime.throw(0x32384b4, 0x15)
	/opt/hostedtoolcache/go/1.13.15/x64/src/runtime/panic.go:774 +0x72 fp=0xc001cbc7d8 sp=0xc001cbc7a8 pc=0x42f272
runtime.mapassign_faststr(0x2ccb020, 0xc0014737a0, 0x3253b2b, 0x1f, 0xc001037740)
	/opt/hostedtoolcache/go/1.13.15/x64/src/runtime/map_faststr.go:291 +0x3fe fp=0xc001cbc840 sp=0xc001cbc7d8 pc=0x41316e
kubesphere.io/kubesphere/pkg/models/resources/v1alpha3/node.nodesGetter.annotateNode(0x39336e0, 0xc0000dba90, 0xc0002d8000, 0xc000c30900, 0x1b, 0x20)
	/home/runner/work/kubesphere/kubesphere/pkg/models/resources/v1alpha3/node/nodes.go:177 +0x506 fp=0xc001cbcdb0 sp=0xc001cbc840 pc=0x1809c76
kubesphere.io/kubesphere/pkg/models/resources/v1alpha3/node.nodesGetter.List(0x39336e0, 0xc0000dba90, 0x0, 0x0, 0xc001d2f700, 0x2030107, 0x4, 0xfffffffffffffff
f)
	/home/runner/work/kubesphere/kubesphere/pkg/models/resources/v1alpha3/node/nodes.go:120 +0x786 fp=0xc001cbcf50 sp=0xc001cbcdb0 pc=0x1809216
kubesphere.io/kubesphere/pkg/models/resources/v1alpha3/node.(*nodesGetter).List(0xc0003aa210, 0x0, 0x0, 0xc001d2f700, 0xc0003aa210, 0xc001d72ea8, 0xfffffffffff
fffff)
	<autogenerated>:1 +0x67 fp=0xc001cbcfa0 sp=0xc001cbcf50 pc=0x180bde7
kubesphere.io/kubesphere/pkg/models/resources/v1alpha3/resource.(*ResourceGetter).List(0xc00000e070, 0xc0015e05dc, 0x5, 0x0, 0x0, 0xc001d2f700, 0x4, 0x4, 0xc00
07344e0)
[...]

PR #4599, opened 2022-01-07
Fix: deepcopy before mutating shared objects

dkeven:
Signed-off-by: dkeven <keven@kubesphere.io>

### What type of PR is this?
/kind bug

Do a `DeepCopy()` before modifying shared objects from the informers to avoid mutaing the shared objects & avoid 
panic caused by concurrent map reading and writing.

Refer: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/controllers.md

```
Never mutate original objects! Caches are shared across controllers, this means that if you mutate your "copy" (actually a reference or shallow copy) of an object, you'll mess up other controllers (not just your own).

The most common point of failure is making a shallow copy, then mutating a map, like Annotations. Use api.Scheme.Copy to make a deep copy.
```

Fixes https://github.com/kubesphere/kubesphere/issues/4357

Fixes https://github.com/kubesphere/kubesphere/issues/3469

See also https://kubesphere.com.cn/forum/d/6466-fatal-error-concurrent-map-writes

```release-note
None
```

### DIFF
--- a/pkg/models/resources/v1alpha2/persistentvolumeclaim/persistentvolumeclaims.go
+++ b/pkg/models/resources/v1alpha2/persistentvolumeclaim/persistentvolumeclaims.go
@@ -123,6 +123,7 @@ func (s *persistentVolumeClaimSearcher) Search(namespace string, conditions *par
 
 	r := make([]interface{}, 0)
 	for _, i := range result {
+		i = i.DeepCopy()
 		inUse := s.countPods(i.Name, i.Namespace)
 		isSnapshotAllow := s.isSnapshotAllowed(i.GetAnnotations()["volume.beta.kubernetes.io/storage-provisioner"])
 		if i.Annotations == nil {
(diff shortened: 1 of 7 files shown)


zryfish:
@dkeven I see what you mean here. It seems we all ignore the fact informer returns references, not the values. The PR looks good to me. Could you please add comments to explain why we need deepcopy so others would avoid the same mistake in the future.

dkeven:
> @dkeven I see what you mean here. It seems we all ignore the fact informer returns references, not the values. The PR looks good to me. Could you please add comments to explain why we need deepcopy so others would avoid the same mistake in the future.

Do you mean add comments in the code, or in this PR's comments here?

zryfish:
> 

In the code.

dkeven:
Okay, I've added comments under the `pkg/models/resources/v1alpha3/node/nodes.go` for reference.

In this PR, I‘ve fixed the `node` module and some modules I've known, other modules should also be checked by corresponding maintainers for this issue.

PR #4605, opened 2022-01-10
Cleanup cluster controller and remove unused code

iawia002:
### What type of PR is this?

/kind cleanup

* Deepcopy the cluster object before we mutate it.
* Remove unused code.

Fixes #

```
```

```release-note
NONE
```

```docs

```

/cc @zryfish

### DIFF
--- a/pkg/controller/cluster/cluster_controller.go
+++ b/pkg/controller/cluster/cluster_controller.go
@@ -309,24 +309,6 @@ func buildClusterData(kubeconfig []byte) (*clusterData, error) {
 	}, nil
 }
 
-func (c *clusterController) syncStatus() error {
-	clusters, err := c.clusterLister.List(labels.Everything())
-	if err != nil {
-		return err
-	}
-
-	for _, cluster := range clusters {
-		key, err := cache.MetaNamespaceKeyFunc(cluster)
-		if err != nil {
-			return err
-		}
-
-		c.queue.AddRateLimited(key)
-	}
-
-	return nil
-}
-
 // reconcileHostCluster will create a host cluster if there are no clusters labeled 'cluster-role.kubesphere.io/host'
 func (c *clusterController) reconcileHostCluster() error {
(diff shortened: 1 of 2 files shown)


pkg/controller/cluster/cluster_controller.go
```diff
@@ -622,12 +590,10 @@ func (c *clusterController) syncCluster(key string) error {
 		return err
 	}
 
-	if !reflect.DeepEqual(oldCluster, cluster) {
```
zryfish, line 625: Why do we remove `DeepEqual`, will skip updating object if they are `deep equaled`

iawia002: This judgment is meaningless, the time in the condition is always changing, two objects can not be the same

zryfish: Which condition keeps changing?

iawia002: https://github.com/kubesphere/kubesphere/blob/365924e76b9b9fbe00eb8650678678db6dc35f33/pkg/controller/cluster/cluster_controller.go#L610-L618

wansir: Reconcile will be triggered after the update,  if not compare conditions, it will keep looping, right?

iawia002: Yes, I missed that, but again, the old `DeepEqual` method doesn't really work

zryfish: @wansir Any more comments?

iawia002: Maybe I should revert the status part change, reorganize that part and submit a separate PR.

zryfish: Any update? What do we do with this PR next?

iawia002: > Maybe I should revert the status part change, reorganize that part and submit a separate PR.

I'm currently working on the workspace API, I'll update this soon.

PR #4613, opened 2022-01-12
fix: Using defer in loops and end of loop call fileData.Close() can lead to resource leaks

gaffeyQiu:
### What type of PR is this?
/kind bug

may lead to resource leaks
1. Using defer in loops
2. call fileData.Close() is too late

Fixes #


LinuxSuRen (commented):
hi @gaffeyQiu , thanks for your time. DevOps has already been moved into a new project [ks-devops](https://github.com/kubesphere/ks-devops). Please feel free to file a new PR on that.

Issue #4580, opened 2021-12-30
Feat: "resources.kubesphere.io" Kapis group refactor


RolandMa1986:
**What's it about?**

Refactor the "resources.kubesphere.io" Kapis group with a generalized implementation. And obsolete this APIs Group in the feature. 
In the feature, all the CRDs(or CRDs with a kapis server label) can be register to the ks-apiserver automatically, as same as the Kubernetes APIs(including the URI), but still can provider the  frond-end friendly API style. 

**What's the reason why we need it?**

"resources.kubesphere.io" Kapis group is a common List/Get APIs for the Custom Resources, which provided a powerful API with filter, sort and pagging abilities than the origin Kubernetes APIs. However, a "resources/v1alpha3.Interface" need to be implement for the CRDs which need to be regiested to the API group.  If we digger into the Interface implementation, we can found that more than 60% code are duplicated, Many CRDs don't need implictent the **Filter**, **Comparer** for itself. 

Another disadvantage of this group is that we need an other GRV for this Group, which means we had to grant the permission seperatetlly. If we can register the API GRV as same as the CRD definenation, we can save a lot of work during grant  permission. 

I believe this is an important feature for Kubesphere. There're a few use cases:

* case one
* case two
* ...

Please leave your comments below if there's anyone agrees with me. Or just give me a thumb up.

**Area Suggestion**
/area api
/area apiserver

/kind feature-request

VioZhang:
finished

PR #4617, opened 2022-01-13
feat: Serving CRD in ks apiserver

RolandMa1986:
### What type of PR is this?
/kind feature

KS API server provided frontend friendly queries which provided a powerful API with filter, sort, and pegging abilities than the origin Kubernetes APIs.  So that many newly added CRDs need to be implemented in the old design.

In this PR, We will register CRDs that need to be served in the ks-apiserver with default Filters, Comparers. 

**Example:** 
```
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  annotations:
    kubesphere.io/resource-scope: workspaced
  labels:
    kubesphere.io/resource-served: "true"
spec:
  scope: Cluster
  group: cert-manager.io
  name:
...
```
After applied the label and annotation, the KS API server starts to serve each enabled CRD at an HTTP REST endpoint. In the above example, the API versions are available at `/kapis/cert-manager.io/v1`.

All the endpoints list in the bellow tables:

| URL | Scope | comments |
| -------- | -------- | -------- |
| kapis/{group}/{version}/{kind}/  | Cluster  | Query cluster resources |
| kapis/{group}/{version}/{kind}/{name}  | Cluster  | Get a cluster resource |
| kapis/{group}/{version}/{kind}/  | Namespace  | Query namespaced resources in all namspaces |
| kapis/{group}/{version}/namespaces/{namespace}/{kind}/ | Namespace |Query resources in the namespace|
| kapis/{group}/{version}/namespaces/{namespace}/{kind}/{name} | Namespace | Get a namespace resource |
| kapis/{group}/{version}/workspaces/{workspace}/{kind}/ | Workspace | Query cluster resources that belong to workspace |
| kapis/{group}/{version}/workspaces/{workspace}/{kind}/{name} | Workspace | Get a cluster resource that belong to workspace |

Fixes # #4580

This is a breakdown of PR https://github.com/kubesphere/kubesphere/pull/4608

```release-note
None
```

```docs

```

### DIFF
--- a/pkg/apiserver/apiserver.go
+++ b/pkg/apiserver/apiserver.go
@@ -25,17 +25,8 @@ import (
 	"strconv"
 	"time"
 
-	"kubesphere.io/kubesphere/pkg/utils/iputil"
-
-	"kubesphere.io/kubesphere/pkg/apiserver/authentication/token"
-
-	"kubesphere.io/kubesphere/pkg/apiserver/authorization"
-
-	"kubesphere.io/api/notification/v2beta1"
-
-	openpitrixv2alpha1 "kubesphere.io/kubesphere/pkg/kapis/openpitrix/v2alpha1"
-
 	"github.com/emicklei/go-restful"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	urlruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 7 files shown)


zryfish:
@RolandMa1986  I don't see any authorization code here. Are these APIs available for all users?

RolandMa1986:
> @RolandMa1986 I don't see any authorization code here. Are these APIs available for all users?

We don't need a specific authorization code here, since all the newly added APIs satisfied the GVR format.  the default RBAC authorization will be used ref: https://github.com/kubesphere/kubesphere/blob/master/pkg/apiserver/authorization/rbac/rbac.go

wansir:
> > @RolandMa1986我在这里看不到任何授权码。这些 API 是否可供所有用户使用？
> 
> 我们这里不需要特定的授权码，因为所有新添加的 API 都满足 GVR 格式。将使用默认的 RBAC 授权参考：[https ://github.com/kubesphere/kubesphere/blob/master/pkg/apiserver/authorization/rbac/rbac.go](https://github.com/kubesphere/kubesphere/blob/master/pkg/apiserver/authorization/rbac/rbac.go)

That's right, all of standard API will use RBACAuthorizer to verify permissions.

Issue #4604, opened 2022-01-10
 Optimize data retention policies for logging, auditing, and events


wanjunlei:
**What's it about?**

There are some issues with the current data retention policy for logging, auditing, and events.

1. The data retention period for logging, auditing, and events is the same and cannot be set individually.
2. Excessive cleaning when cleaning expired data. For example, when multiple clusters share an ES cluster, cluster a will also clear the expiration time of cluster b when it cleans up expired data. Problems can arise if clusters a and b have different data retention periods.

So, we need to optimize the data retention policies from the following aspects.

1. Support to set the separate data retention periods for logging, auditing, and events. We can add a parameter to `ClusterConfig` like this.

```
spec:
  common:
    es:
      logMaxAge: 7
  logging:
    logMaxAge: 15
  auditing:
    auditingMaxAge: 30
  events:
    eventMaxAge: 60
```

When `logging.logMaxAge` or `auditing.auditingMaxAge` or `evnets.eventMaxAge` is not set, `common.es.logMaxAge` will be used.

2. When cleaning expired data, the index needs to be filtered, and only the index created by the cluster can be deleted.
The config file for the curator may be like this.

```
1:
    action: delete_logging_indices
    description: "Clean up ES by deleting old indices"
    options:
      timeout_override:
      continue_if_exception: False
      disable_action: False
      ignore_empty_list: True
    filters:
    - filtertype: age
      source: name
      direction: older
      timestring: '%Y.%m.%d'
      unit: days
      unit_count: 7
      field:
      stats_result:
      epoch:
      exclude: False
    - filtertype: pattern
      kind: prefix
      value: ks-logstash-log
      exclude: False  
```

**What's the reason why we need it?**

I believe this is an important feature for Kubesphere. There're a few use cases:

* case one
* case two
* ...

Please leave your comments below if there's anyone agrees with me. Or just give me a thumb up.

**Area Suggestion**

/kind feature-request
/area observability

benjaminhuo:
It's better to make it more clear in the example that the prefix should have cluster id/name :
```
    - filtertype: pattern
      kind: prefix
      value: cluster-xxx-ks-logstash-log
      exclude: False  
```

PR #4631, opened 2022-01-19
Optimize the error message returned by the logging, auditing and events

wenchajun:
Signed-off-by: chengdehao <dehaocheng@yunify.com>

### What type of PR is this?

/kind feature

Users often locate information based on errors. In the current ks-apiserver the error messages in elasticsearch are also handled. ks-apiserver returns error messages in `es.reason`, but the error messages may  not be specific enough to pinpoint the error message. So now we returns `error.root_casue[0].reason` information.

Fixes https://github.com/kubesphere/kubesphere/issues/4604

I have introduced the elasticsearch package to wrap the error messages.Please see https://pkg.go.dev/github.com/dzlab/elastic-go#Error .  Also in testing it was found that the return message from elasticsearch may not have information about `error.root_casue[0]` in it. So the error message was judged so that the error message obtained was not empty.

```release-note
NONE
```


go.mod
```diff
@@ -26,6 +26,7 @@ require (
 	github.com/docker/docker v20.10.7+incompatible
 	github.com/docker/go-metrics v0.0.0-20181218153428-b84716841b82 // indirect
 	github.com/docker/libtrust v0.0.0-20160708172513-aabc10ec26b7 // indirect
+	github.com/dzlab/elastic-go v0.0.0-20160522173446-355f7c99ab99
```
zryfish, line 29: Is this package compatible with es v5/v6/v7? If so, could we just replace all dependencies `github.com/elastic/go-elasticsearch`?

wenchajun: `github.com/elastic/go-elasticsearch` belongs to the official dependencies. So it's more comprehensive. I referenced `github.com/dzlab/elastic-go` because of the error structure of the elasticsearch.  So I don't think it can replace that official package.

wanjunlei:
It's not a good idea to use a package that is no longer maintained. It's a very simple modification, you only need to define a struct to solve this problem, it is no need to introduce an unreliable package.

wenchajun:
> It's not a good idea to use a package that is no longer maintained. It's a very simple modification, you only need to define a struct to solve this problem, it is no need to introduce an unreliable package.

Thanks, I will close this pr and modify him.

PR #4633, opened 2022-01-19
Query workspace users with groupbindings

### LINKED ISSUE #4630: Users that are invited to the Workspace through Group should also be able to invite to Project
**What's it about?**
Currently, when a user is added to a Workspace through Group, the User can't be added to a sub-project of the workspace directly.  The user must be invited to the Workspace Members,  so that the user can be added to the sub-projects. Which doesn't make sense.

Users that are invited to the Workspace through groups should also be able to invite to Project too.

* case one
* case two
* ...

**Area Suggestion**
/area iam
/kind feature-request

RolandMa1986:

### What type of PR is this?
/kind feature
/kind api-change

Include users that are added to the Workspace through the Group when querying users by workspace.

Fixes #4630

```
```

```release-note
Users that are invited to Workspace Group will be included when querying users by workspace.
```

```docs

```

### DIFF
--- a/pkg/models/resources/v1alpha3/user/users.go
+++ b/pkg/models/resources/v1alpha3/user/users.go
@@ -164,6 +164,37 @@ func (d *usersGetter) listAllUsersInWorkspace(workspace, role string) ([]*iamv1a
 		}
 	}
 
+	groupBindings, err := d.ksInformer.Iam().V1alpha2().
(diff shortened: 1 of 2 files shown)


wansir:
/hold

There are some differences between workspace member management and organizational management. How to remove the organization member from the workspace member list if we mixed the two parts. I think a query condition should be used to filter the kind of workspace member, distinguish the entry of member management and organization management.

Issue #4603, opened 2022-01-10
Optimize the error message returned by the logging, auditing and events


wanjunlei:
**What's it about?**

This is an error message returned by ElasticSearch, and the `ks-apiserve` return the `error.reason` to the caller. 
This return value cannot tell the caller what caused the error. The correct thing to do is to return `error.root_casue[0].reason` to the caller.

```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "query_shard_exception",
        "reason" : "No mapping found for [date] in order to sort on",
        "index_uuid" : "ru8-IZ99RZeR2J-iDr1mDQ",
        "index" : "ks-logstash-log-2022.01.07"
      },
      {
        "type" : "query_shard_exception",
        "reason" : "No mapping found for [date] in order to sort on",
        "index_uuid" : "klv0PTipQ9Gi0RzdcxfwNA",
        "index" : "ks-logstash-log-2022.01.10"
      }
    ],
    "type" : "search_phase_execution_exception",
    "reason" : "all shards failed",
    "phase" : "query",
    "grouped" : true
  },
  "status" : 400
}

```

**What's the reason why we need it?**

I believe this is an important feature for Kubesphere. There're a few use cases:

* case one
* case two
* ...

Please leave your comments below if there's anyone agrees with me. Or just give me a thumb up.

**Area Suggestion**

/kind feature-request
/area observability

PR #4643, opened 2022-01-25
Optimize the error message

wenchajun:
Signed-off-by: chengdehao <dehaocheng@yunify.com>

### What type of PR is this?

/kind feature

Users often locate information based on errors. In the current ks-apiserver the error messages in elasticsearch are also handled. ks-apiserver returns error messages in `es.reason`, but the error messages may not be specific enough to pinpoint the error message. So now we returns `error.root_casue[0].reason` information.

Fixes https://github.com/kubesphere/kubesphere/issues/4603

I have introduced the elasticsearch package to wrap the error messages.Please see https://pkg.go.dev/github.com/olivere/elastic/v7#Error . Also in testing it was found that the return message from elasticsearch may not have information about `error.root_casue[0]` in it. So the error message was judged so that the error message obtained was not empty.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/simple/client/es/versions/error.go
+++ b/pkg/simple/client/es/versions/error.go
@@ -1,0 +1,19 @@
+package versions
+
+type Error struct {
+	Status  int           `json:"status"`
+	Details *ErrorDetails `json:"error,omitempty"`
+}
+
+type ErrorDetails struct {
+	Type         string                   `json:"type"`
+	Reason       string                   `json:"reason"`
+	ResourceType string                   `json:"resource.type,omitempty"`
+	ResourceId   string                   `json:"resource.id,omitempty"`
+	Index        string                   `json:"index,omitempty"`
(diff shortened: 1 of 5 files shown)


pkg/simple/client/es/versions/error.go
```diff
@@ -0,0 +1,32 @@
+package versions
+
+type Error struct {
+	Status  int           `json:"status"`
+	Details *ErrorDetails `json:"error,omitempty"`
+}
+
+type ErrorDetails struct {
```
wanjunlei, line 8: Useless fields can be deleted.

wenchajun: fixed, I have removed some fields by referring to the structure of other `error` packages.

wanjunlei (approved):

benjaminhuo:
This is a good enhancement , thanks @wenchajun @wanjunlei 
/lgtm

PR #4651, opened 2022-01-27
Use the kube-system UID to identify if the member cluster already exists

### LINKED ISSUE #4625: The host cluster cannot be removed when I add it as a member cluster.
**Describe the Bug**
When I mistakenly managed the host cluster itself as a member cluster, I tried to remove it but could not do so.

**Versions Used**
KubeSphere:v3.2.1
Kubernetes:v1.20

**Environment**
Centos 7/3 master ,4 nodes

**How To Reproduce**
1. In the host cluster, click `Add Cluster`, fill in the related information, and select the `Direct Connection` mode
2. Paste the host cluster's own `Kubeconfig` file into the text box
3. Try to remove the newly managed cluster
image
image
There are no associated action buttons in the red box

**Expected behavior**
I think I cannot add myself as a member cluster,this is unreasonable.

iawia002:
### What type of PR is this?

/kind bug

See background info and our previous discussion at #4625, We now use the UID of the target cluster's `kube-system` namespace as the unique identifier of the cluster to determine if the cluster already exists.

Screen Shot 2022-01-27 at 11 35 03 AM

Fixes #4625

This PR also does some code cleanup work, reducing a lot of unnecessary config parsing and client creation.

```release-note
NONE
```

```docs

```

/cc @zryfish

### DIFF
--- a/pkg/kapis/cluster/v1alpha1/handler.go
+++ b/pkg/kapis/cluster/v1alpha1/handler.go
@@ -301,13 +301,13 @@ func (h *handler) updateKubeConfig(request *restful.Request, response *restful.R
 		return
 	}
 
-	_, err = validateKubeSphereAPIServer(cluster.Spec.Connection.KubeSphereAPIEndpoint, cluster.Spec.Connection.KubeConfig)
+	_, err = validateKubeSphereAPIServer(config)
 	if err != nil {
 		api.HandleBadRequest(response, request, fmt.Errorf("unable validate kubesphere endpoint, %v", err))
 		return
 	}
 
-	err = h.validateMemberClusterConfiguration(cluster.Spec.Connection.KubeConfig)
+	err = h.validateMemberClusterConfiguration(clientSet)
 	if err != nil {
 		api.HandleBadRequest(response, request, fmt.Errorf("failed to validate member cluster configuration, err: %v", err))
 	}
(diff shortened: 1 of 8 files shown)


pkg/kapis/cluster/v1alpha1/handler.go
```diff
@@ -372,60 +381,42 @@ func (h *handler) validateKubeConfig(kubeconfig []byte) error {
 		return err
 	}
 
-	// clusters with the exactly same KubernetesAPIEndpoint considered to be one
+	// clusters with the exactly same kube-system namespace UID considered to be one
 	// MUST not import the same cluster twice
-	for _, cluster := range clusters {
-		if len(cluster.Spec.Connection.KubernetesAPIEndpoint) != 0 && cluster.Spec.Connection.KubernetesAPIEndpoint == config.Host {
-			return fmt.Errorf("existing cluster %s with the exacty same server address, MUST not import the same cluster twice", cluster.Name)
+	for _, existedCluster := range clusters {
```
zryfish, line 386: Could we add the clusterid to the status of Cluster object? I think we can use it in other scenarios too, it would also be easier to check clusterid here.

iawia002: I'm cool with this idea, but I think we should add it into an annotation, it's not a status message.

zryfish: Status usually set by controller, means watched state, which is exactly our case, I think it's fine to set it into status.

iawia002: I added a new `UID` field to status, PTAL

zryfish: Seen it, but there are too many changes on CRD yamls which shouldn't be. It's better to make changes only related to this RP.

iawia002: I have submitted a PR to update the automatically generated files separately, we need to merge that first: #4669

iawia002: Updated, PATL

PR #4659, opened 2022-02-08
feat: live-reload when configuration changed

### LINKED ISSUE #4504: [Feature Request] Reload configuration without restarting
**What's it about?**
Currently, when configuration changed, need to redeploy ks-apiserver/ks-controller-manager to apply the changes. It could cause inconsistencies when KubeSphere deployed without HA.  Also, It will be more user-friendly if user could enable components directly from the console without editing [ClusterConfiguration](https://github.com/kubesphere/ks-installer#enable-pluggable-components) manually.

I believe we need this configuration live reload because

- No need to redeploy `ks-apiserver/ks-controller-manager` obviously, it would help make KubeSphere more robust.

From my perspective, we can break this down into the following:

- [ ] Refactor current configuration loading code, make sure every option has a default value.
- [ ] Watch configuration changes, live-reload when configuration changed.

**Area Suggestion**
/area apiserver
/kind feature-request

x893675:
Signed-off-by: x893675 <x893675@icloud.com>

### What type of PR is this?
/kind feature

Currently, when configuration changed, need to redeploy ks-apiserver/ks-controller-manager to apply the changes. It could cause inconsistencies when KubeSphere deployed without HA. Also, It will be more user-friendly if user could enable components directly from the console without editing [ClusterConfiguration](https://github.com/kubesphere/ks-installer#enable-pluggable-components) manually.

Fixes #4504

```
None
```

```release-note
Support live-reload when config file changed.
```

```docs

```


cmd/controller-manager/app/server.go
```diff
@@ -114,6 +114,37 @@ func NewControllerManagerCommand() *cobra.Command {
...
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			ictx.CancelContext()
+			return nil
+		case cfg := <-configCh:
+			ictx.CancelContext()
+			s.MergeConfig(&cfg)
```
zryfish, line 135: Do we need to validate the new config before merging? Or just merge to a fast fail.

pkg/apiserver/config/config.go
```diff
@@ -79,6 +85,11 @@ import (
 // mysql-host is missing in command line flags, all other mysql command line flags
 // will be ignored.
 
+var (
+	once        sync.Once
+	cfgChangeCh = make(chan Config)
```
zryfish, line 90: Why not create cfgChangeCh as a local variable, I see nowhere else use it.

cmd/controller-manager/app/server.go
```diff
@@ -114,6 +114,37 @@ func NewControllerManagerCommand() *cobra.Command {
 	return cmd
 }
 
+func Run(s *options.KubeSphereControllerManagerOptions, configCh <-chan controllerconfig.Config, ctx context.Context) error {
+	ictx := controllerconfig.Context{}
```
zryfish, line 118: Why do we need to create a new Context{}? Can we just use default context with cancel function?
```
ictx, cancel := context.WithCancel(context.Background)
```

zryfish:
@x893675  The overall looks promising, just a few questions. And it would be better to add more comments to make it more straightforward for others.

cmd/controller-manager/app/server.go
x893675: When misconfigured, I think the app should just crash directly

pkg/apiserver/config/config.go
x893675: Now the `viper` initialization of the `TryLoadFromDisk` function is only executed once, and there are some problems with defining it as a local variable. I am going to optimize the cofnig package in another PR

cmd/controller-manager/app/server.go
x893675: The default context (`signals.SetupSignalHandler()`) is to control the entire program life cycle,  
A new context is created here to control the life cycle of the controller-manager(all controllers, sharedInformer, webhook etc.)

x893675:
> 

I'll add some comments later

cmd/controller-manager/app/server.go
zryfish: Yes, I understand the necessity of a new context, my concern is why not use `context.WithCancel` rather than create a new type?

x893675: I encapsulated the context so that ks-apiserver and controller-manager use the same context API. This is really confusing, I will remove it and add a comment

pkg/apiserver/config/config.go
x893675: The once in TryLoadFromDisk is added by me. The initialization of viper only needs to be executed once, especially the initialization of OnConfigChange. If the variable is defined in once of TryLoadFromDisk, according to the current function signature, it will not be accessible outside of once. If placed outside of once, multiple calls to TryLoadFromDisk will initialize multiple channels. 

I think the initialization of the configuration and the opening of the watch config channel should be separated from TryLoadFromDisk and exposed as a separate api. Or config is a singleton

I will optimize the config package in this pr

zryfish:
@x893675  Thanks. Looks good to me.  @wansir Any comments?
/lgtm

wansir:
> @x893675 Thanks. Looks good to me. @wansir Any comments? /lgtm

I'll test this pr later.

zryfish:
@x893675 I think we are ready to merge this PR, this only one thing left, please add the release node in the PR description. You can find section like below, add your release note in the `release-note`, you can reference other PRs.


```
For more information on release notes see: https://github.com/kubernetes/community/blob/master/contributors/guide/release-notes.md
-->
```release-note

\```
```

x893675:
> @x893675 I think we are ready to merge this PR, this only one thing left, please add the release node in the PR description. You can find section like below, add your release note in the `release-note`, you can reference other PRs.
> 
> ```
> For more information on release notes see: https://github.com/kubernetes/community/blob/master/contributors/guide/release-notes.md
> -->
> ```release-note
> 
> \```
> ```

@zryfish  done

zryfish:
@x893675  Thanks.
/approve

Issue #4185, opened 2021-08-27
Automatic resizing of PVC before out of space


stoneshi-yunify:
## Background
Many users want their PVCs to be automatically resized when certain contains met (e.g. usage > 90%), to prevent workload unavailability due to out of space.

## Proposal
- use crd/controller to monitor the pvc and do resize
- may integrate with Prometheus

/area storage
/kind feature-request

f10atin9:
These few project may be helpful：
https://github.com/topolvm/pvc-autoresizer
https://github.com/opstree/dynamic-pv-scaler

PR #4660, opened 2022-02-09
add pvc-autoresizer controller to ks-controller-manager

f10atin9:
### What type of PR is this?

/kind feature

Resizing pvc automatically is a good idea, and the additional installation of pvc-autoresizer is not convenient enough.
So I add pvc-autoresizer's controller to ks. There is no need to install pvc-autoresizer additionally.

Fixes #
#4185 
```
None
```

```release-note
When the user needs to use these functions, they just need to add the same annotations as before with pvc-autoresizer.
The console will support control it soon.
```

```docs
None
```


cmd/controller-manager/app/controllers.go
```diff
@@ -348,6 +352,36 @@ func addAllControllers(mgr manager.Manager, client k8s.Client, informerFactory i
 		addController(mgr, "volumesnapshot", volumeSnapshotController)
 	}
 
+	// "pvc-autoresizer"
+	if cmOptions.IsControllerEnabled("pvc-autoresizer") {
+		if err := runners.SetupIndexer(mgr, false); err != nil {
+			return err
+		}
+
+		promClient, err := runners.NewPrometheusClient("http://prometheus-k8s.kubesphere-monitoring-system:9090")
```
stoneshi-yunify, line 361: is the url `http://prometheus-k8s.kubesphere-monitoring-system:9090` fixed? hard code is not a good idea. at least it should be constructed with some constants.

stoneshi-yunify (changes requested):
overall looks good. good job!

cmd/controller-manager/app/controllers.go
f10atin9: Okay，I‘ll fix it.

stoneshi-yunify (approved):
/lgtm

cmd/controller-manager/app/controllers.go
```diff
@@ -348,6 +354,37 @@ func addAllControllers(mgr manager.Manager, client k8s.Client, informerFactory i
 		addController(mgr, "volumesnapshot", volumeSnapshotController)
 	}
 
+	// "pvc-autoresizer"
+	if cmOptions.IsControllerEnabled("pvc-autoresizer") {
+		if err := runners.SetupIndexer(mgr, false); err != nil {
+			return err
+		}
+
+		prometheusUrl := "http://" + constants.PrometheusSvcName + "." + constants.KubeSphereMonitoringNamespace + ":" + constants.PrometheusSvcPort
```
zryfish, line 363: Shouldn't create prometheusUrl using default service name, the endpoint of Prometheus could be different. Please use the address of  monitoring endpoint, you can reference https://github.com/kubesphere/kubesphere/blob/master/cmd/ks-apiserver/app/options/options.go#L123 for example.

cmd/controller-manager/app/controllers.go
```diff
@@ -348,6 +352,38 @@ func addAllControllers(mgr manager.Manager, client k8s.Client, informerFactory i
...
+				mgr.GetClient(),
+				ctrl.Log.WithName("pvc-autoresizer"),
+				1*time.Minute,
+				mgr.GetEventRecorderFor("pvc-autoresizer"),
+			)
+			addController(mgr, "pvcautoresizer", pvcAutoResizerController)
+		}
+	}
+
+	if cmOptions.IsControllerEnabled("workload-Restart") {
```
zryfish, line 377: `workload-Restart` is a bad controller name, and its style is inconsistent, maybe choose another name.

f10atin9: How about `expansion-restarter` or`autorestarter`?  Or any other better suggestions?

zryfish: yeah, sounds better. This controller is only used when PVC size changes, right? So, how about `pvc-workload-restarter`?

f10atin9: Yes it does, `pvc-workload-restarter` sounds great, I'll update it, and thanks for the suggestion.

PR #4679, opened 2022-02-18
Set the name of the current cluster into the kubesphere-config configmap

iawia002:
### What type of PR is this?

/kind feature

### What this PR does / why we need it:

See #4672 for background info, after this patch, the cluster name will be added into the `kubesphere-config` configmap:

before:

```yaml
data:
  kubesphere.yaml: |
    authentication:
      authenticateRateLimiterMaxTries: 10
      authenticateRateLimiterDuration: 10m0s
      loginHistoryRetentionPeriod: 168h
      maximumClockSkew: 10s
      multipleLogin: True
      kubectlImage: kubesphere/kubectl:v1.21.0
      jwtSecret: "LvHOc5ATkJy8amuW0NVZU6jNPrYG2qlj"
      oauthOptions:
        clients:
        - name: kubesphere
          secret: kubesphere
          redirectURIs:
          - '*'
    network:
      ippoolType: none
    openpitrix:
      s3:
        endpoint: http://minio.kubesphere-system.svc:9000
        region: us-east-1
        disableSSL: True
        forcePathStyle: True
        accessKeyID: openpitrixminioaccesskey
        secretAccessKey: openpitrixminiosecretkey
        bucket: app-store
    multicluster:
      enable: true
      agentImage: kubesphere/tower:v0.2.0
      proxyPublishService: tower.kubesphere-system.svc
      proxyPublishAddress: http://103.61.38.60:30361
    monitoring:
      endpoint: http://prometheus-operated.kubesphere-monitoring-system.svc:9090
      enableGPUMonitoring: false
    gpu:
      kinds:
      - resourceName: nvidia.com/gpu
        resourceType: GPU
        default: True
    notification:
      endpoint: http://notification-manager-svc.kubesphere-monitoring-system.svc:19093
    logging:
      host: http://elasticsearch-logging-data.kubesphere-logging-system.svc:9200
      indexPrefix: ks-logstash-log
    alerting:
      prometheusEndpoint: http://prometheus-operated.kubesphere-monitoring-system.svc:9090
      thanosRulerEndpoint: http://thanos-ruler-operated.kubesphere-monitoring-system.svc:10902
      thanosRuleResourceLabels: thanosruler=thanos-ruler,role=thanos-alerting-rules
    gateway:
      watchesPath: /var/helm-charts/watches.yaml
      namespace: kubesphere-controls-system
```

after:

```yaml
data:
  kubesphere.yaml: |
    authentication:
      authenticateRateLimiterMaxTries: 10
[...]

PR #4681, opened 2022-02-21
Unify the omitempty configuration of YAML annotation

iawia002:
### What type of PR is this?

/kind cleanup

Part of #4679, see https://github.com/kubesphere/kubesphere/pull/4679#issuecomment-1046563697 for more info.

Fixes #

```
```

```release-note
NONE
```

```docs

```

/cc @zryfish

### DIFF
--- a/pkg/apiserver/authentication/oauth/options.go
+++ b/pkg/apiserver/authentication/oauth/options.go
@@ -63,10 +63,10 @@ type Options struct {
 	Issuer string `json:"issuer,omitempty" yaml:"issuer,omitempty"`
 
 	// RSA private key file used to sign the id token
-	SignKey string `json:"signKey,omitempty" yaml:"signKey"`
+	SignKey string `json:"signKey,omitempty" yaml:"signKey,omitempty"`
 
 	// Raw RSA private key. Base64 encoded PEM file
-	SignKeyData string `json:"-,omitempty" yaml:"signKeyData"`
+	SignKeyData string `json:"-,omitempty" yaml:"signKeyData,omitempty"`
 
 	// Register identity providers.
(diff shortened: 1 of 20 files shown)


pkg/apiserver/config/config.go
```diff
@@ -167,7 +167,7 @@ func TryLoadFromDisk() (*Config, error) {
 		}
 	}
 
-	conf := New()
+	conf := &Config{}
```
zryfish, line 170: `New()` creates a default config, why change this?

iawia002: Many fields have incorrect (or arbitrary) default values but they are not `omitempty`, and I think we should parse the YAML values into an empty structure so as not to create previously unconfigured, redundant values.

zryfish: But, for example, if there is no monitoring configuration in the configuration file, we still want to use the default endpoint `prometheus.kubesphere-monitoring-system.svc`, by changing to `&Config{}`, is that possible?

iawia002: Nope, but does this scenario really exist, where the configmap configuration doesn't have it but we need to use it? I always think that we should load the YAML configuration without modifying it too much, to ensure that the whole YAML remains unchanged after `Marshal` and `Unmarshal`.

zryfish: I think yes, https://github.com/kubesphere/kubesphere/blob/f50de9af65abfdcc37faa832f1960128139363ae/pkg/apiserver/authorization/options.go#L35 we use `RBAC` as the default authorization method without explicitly configured in kubesphere-config. I think default options are necessary, especially when developing. It would be difficult to prepare a complete configuration file before running your program for a newbie.

iawia002:
@zryfish Can we merge this? I need this PR to be merged so that I can go back on #4679

PR #4686, opened 2022-02-23
Proxy DevOps APIs with group name and version

### LINKED ISSUE #4684: Request to proxy DevOps APIs with version v1alpha1
**What's it about?**

Request to make APIServer to proxy DevOps APIs with version v1alpha1.

At present, we have separated DevOps application into independent project and let APIServer proxy DevOps APIs with version v1alpha2 and v1alpha3, please see below:

https://github.com/kubesphere/kubesphere/blob/9fbb69faa42b133d60065c4527754de6ba3e65f6/pkg/kapis/devops/v1alpha2/register.go#L33
https://github.com/kubesphere/kubesphere/blob/9fbb69faa42b133d60065c4527754de6ba3e65f6/pkg/kapis/devops/v1alpha3/register.go#L35

But recently, we added a little DevOps APIs with version v1alpha1, please see blow:

- https://github.com/kubesphere/ks-devops/pull/468
- https://github.com/kubesphere/ks-devops/pull/467
- https://github.com/kubesphere/ks-devops/pull/460

I believe this is an important step for KubeSphere to add some new features in DevOps area, like ArgoCD integration and Pipeline Template.

**Area Suggestion**

/area devops

/kind feature-request
/cc @kubesphere/sig-devops 
/assign

JohnNiang:
### What type of PR is this?

/kind api-change
/kind cleanup
/area devops

1. Proxy DevOps APIs with group name and version
2. Refactor old DevOps API registers in `kapis` package

Fixes #4684 

Docker image for test: `johnniang/ks-apiserver:proxy-devops-v1alpha1`.

```release-note
None
```

```docs

```

/cc @kubesphere/sig-devops 
/cc @zryfish 
/cc @wansir


LinuxSuRen (approved):
/lgtm
/approve

zryfish:
I don't understand, since we've already moved to v1alpha2 v1alpha3, why do go back v1alpha1 again?

JohnNiang:
> I don't understand, since we've already moved to v1alpha2 v1alpha3, why do go back v1alpha1 again?

Please pay attention to #4684 , which I mentioned the three fresh CRDs with version v1alpha1. Please see:

- https://github.com/kubesphere/ks-devops/pull/468
- https://github.com/kubesphere/ks-devops/pull/467
- https://github.com/kubesphere/ks-devops/pull/460

They were new to us, so we defined the version of those CRDs into `v1alpha1`. Please correct me if I missed something.

zryfish:
That's the question. I thought we've already deprecated and removed v1alpha1, and since it's no longer existed in the prior version, why do we add new features to v1alpha1 instead move to a newer version? 
And in this PR, we brought back v1alpha1, and remove v1alpha2, which is an API break change. IMO, we should be more careful when we remove an API version, at least deprecate it first, and then remove it in the next release.

LinuxSuRen:
> And in this PR, we brought back v1alpha1, and remove v1alpha2

We didn't remove any API version in this PR.

> why do we add new features to v1alpha1 instead of moving to a newer version

What is your suggestion when we need to add a new CRD?

JohnNiang:
Hi, @zryfish , you may have misunderstood me. I didn't deprecate any APIs. On the contrary, I just refactored them for further proxy, like `v1beta1`, `v1beta2`, `v1` and so on.

IMO, we should deprecate APIs, like `/v1alpha1/resources/resource`, instead of `/v1alpha1`. See 
- https://kubernetes.io/docs/reference/using-api/deprecation-policy/ 
- https://kubernetes.io/docs/reference/using-api/

for more.

zryfish:
My bad, I missed the `v1alpha2`. So, no API versions deprecated, but we do still bring `v1alpha1` back, right? I think we should add new features to the newer API version. Right now, DevOps has three different API versions, it's really confusing.

JohnNiang:
Imagine if all DevOps features are stable, that is, all API versions have reached `v1`. At this time, we need to develop new features. Should the version of the new API be `v1alpha1` or `v2alpha1`, or upgrade all APIs to `v2alpha1`?

LinuxSuRen:
Kill me or not. Start a new feature with version 2? Cannot understand that. Born from a bigger family?

zryfish:
I'm not saying we should create a new version since there are v1alpha2 v1alph3, why can't we add new features to v1alpha3?

LinuxSuRen:
>  since there are v1alpha2 v1alph3, why can't we add new features to v1alpha3?

For the whole project, it's true. But for a particular CRD, it is just started. And those CRDs might be able to be an independent component in the future. It's not reasonable to start a version from the latest one.

For example, what should we do if we already have a version v2alpha3. The new CRD version should be v2alpha3? Or v1alpha1?

zryfish:
> For the whole project, it's true. But for a particular CRD, it is just started. And those CRDs might be able to be an independent component in the future. It's not reasonable to start a version from the latest one.
> 
> For example, what should we do if we already have a version v2alpha3. The new CRD version should be v2alpha3? Or v1alpha1?

OK, let's go back a little. Do you mean  /v1alpha1 is actually proxied to DevOps CRD, while /v1alpha2 /v1alpha3 are not?

LinuxSuRen:
> Do you mean /v1alpha1 is actually proxied to DevOps CRD, while /v1alpha2 /v1alpha3 are not?

Most of the Versions belong to the CRDs.

zryfish:
IMO, the api version may upgrade like the following, the beta versions could be negotiable. 
```
v1alpha1 -> v1alpha2 -> v1beta1 -> v1 -> v2alpha1 -> v2alpha2 -> v2alpha3 -> v2beta1 -> v2
```
So, since there are already v1alpha2 v1alpha3, we should develop features in the latest version. Also, v1alpha3 is semantic(https://semver.org/) greater than v1alpha1.

zryfish:
I think the documentation from K8s https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions is worth reading. 

In short, the expected API evolution is as follows:

```
newapigroup/v1alpha1 -> ... -> newapigroup/v1alphaN ->
newapigroup/v1beta1 -> ... -> newapigroup/v1betaN ->
newapigroup/v1 ->
newapigroup/v2alpha1 -> ...
```
While in alpha we expect to move forward with it, but may break it.

Once in beta we will preserve forward compatibility, but may introduce new versions and delete old ones.

v1 must be backward-compatible for an extended length of time.

JohnNiang:
Hi @zryfish , as you described above, what you mean is the version of API has nothing to do with the version of CRD. Even if we have a CRD with `v1alpha`, the API definitions we provide should be followed the rule you mentioned at https://github.com/kubesphere/kubesphere/pull/4686#issuecomment-1049578463. Is that correct?

pkg/apiserver/apiserver.go
```diff
@@ -65,8 +65,7 @@ import (
 	clusterkapisv1alpha1 "kubesphere.io/kubesphere/pkg/kapis/cluster/v1alpha1"
 	configv1alpha2 "kubesphere.io/kubesphere/pkg/kapis/config/v1alpha2"
 	crd "kubesphere.io/kubesphere/pkg/kapis/crd"
-	devopsv1alpha2 "kubesphere.io/kubesphere/pkg/kapis/devops/v1alpha2"
-	devopsv1alpha3 "kubesphere.io/kubesphere/pkg/kapis/devops/v1alpha3"
+	kapisdevops "kubesphere.io/kubesphere/pkg/kapis/devops"
```
zryfish, line 68: Personally, I prefer to organize APIs in the different groups like the reset, it's easy to tell which versions we are using.

LinuxSuRen: +1 put all the DevOps-related things into one place.

LinuxSuRen (commented):
/lgtm

PR #4695, opened 2022-03-01
Fix cannot change user status to disabled

### LINKED ISSUE #4442: Add account enalbe or disable function
**What's it about?**

The user account should be able to disable or enabled by admin for some reason.

* case one
* case two
* ...

**Area Suggestion**

/kind feature-request

wansir:
### What type of PR is this?

/kind bug

Currently, due to the limitation of the user controller, the user status cannot be changed to disabled manually. We need to support disabled or enabled accounts.

Fixes #4442

```release-note
Support disable/enable account manually.
```

```docs

```


zryfish:
@wansir Does the manifest of the User need to change accordingly?

wansir:
> @wansir Does the manifest of the User need to change accordingly?

There is no change that will affect the User CRD. Already remake the manifest.

Issue #4610, opened 2022-01-12
Feat: improve the user experience of member cluster console


wansir:
**What's it about?**

Improve the user experience and render the console according to different cluster roles.

**What's the reason why we need it?**

Currently, the console of the member cluster is not stable, the incorrect operation will cause data loss or resource conflict.

Here is a list of things we wish to do.

- [x] User/Workspace is managed by kubefed in host cluster, user and workspace management needs to be disabled in member cluster's console.
- [x] Currently, it is not possible to enter workspace from the member cluster due to API differences.
- [x] Pay attention to federated resources and global resources to avoid conflicts.

**Area Suggestion**

 /area console
/kind feature-request

harrisonliu5:
By judging the member cluster to shield the access control platform role creation operation and the creation operation in the enterprise space setting.

iawia002:
I will add the role of the cluster into the `kubesphere-config` configmap:

```diff
multicluster:
  enable: true
  agentImage: kubesphere/tower:v0.2.0
  proxyPublishService: tower.kubesphere-system.svc
  proxyPublishAddress: http://103.61.38.60:30361
+ clusterRole: host/member
```

wenxin-01:
/assign @harrisonliu5 
There is no workspace created in host cluster on the console page of member cluster.

host-cluster console: 
`/kapis/tenant.kubesphere.io/v1alpha2/workspaces?sortBy=createTime&limit=10`
![image](https://user-images.githubusercontent.com/68640256/168262077-dc6d2dfe-bc24-4e14-b471-7cdf39acd950.png)

member-cluster console:
`/kapis/tenant.kubesphere.io/v1alpha2/workspaces?sortBy=createTime&limit=10`
![image](https://user-images.githubusercontent.com/68640256/168262169-1ca98134-4924-4116-9cda-714d0b0c5b10.png)

wenxin-01:
Issue fixed in `kubespheredev/ks-installer:nightly-20220524`. But still have some issue in member-cluster console.
https://github.com/kubesphere/kubesphere/issues/4913

PR #4710, opened 2022-03-07
Set the role of the current cluster into the kubesphere-config configmap

iawia002:
### What type of PR is this?

/kind feature

Part of #4610, after this patch, the cluster role will be added into the kubesphere-config configmap:

```diff
multicluster:
  enable: true
  agentImage: kubesphere/tower:v0.2.0
  proxyPublishService: tower.kubesphere-system.svc
  proxyPublishAddress: http://103.61.38.60:30361
+ clusterRole: host/member
```

Fixes #

```
```

```release-note
NONE
```

```docs

```

/cc @wansir


wansir:
Cluster role implied multicluster enabled status. What about using `clusterRole: none` instead `enabled: false`.

iawia002:
I've changed the `clusterRole` values to `host` and `none`, but I'm not going to remove the previous `enable` field, we should be consistent with `ks-installer`

iawia002:
> `clusterRole: none`

This is not right, I see the ks-installer is using `clusterRole: member`, I'm gonna revert my change.

wansir:
> > `clusterRole: none`
> 
> This is not right, I see the ks-installer is using `clusterRole: member`, I'm gonna revert my change.

`clusterRole: none` means multicluster disabled.

https://github.com/kubesphere/ks-installer/blob/0f81d30827d84a2b1262742d58fcbba70efa8318/deploy/cluster-configuration.yaml#L131-L132

iawia002:
> Cluster role implied multicluster enabled status. What about using `clusterRole: none` instead `enabled: false`.
>
> `clusterRole: none` means multicluster disabled.
> 

I don't get it, what do you want me to change? The configmap will only be updated if multicluster is enabled, what does it have to do with the value `none`

iawia002:
> I think is able to resolve multicluster enable status from the cluster role. Field multicluster.enabled can be removed from multicluster options.

Nope, `clusterRole` is set after the cluster controller is started, enable is used before that to determine if some multi-cluster operation is needed. But the unused `EnableFederation` field can be removed.

wansir:
> > I think is able to resolve multicluster enable status from the cluster role. Field multicluster.enabled can be removed from multicluster options.
> 
> Nope, `clusterRole` is set after the cluster controller is started, `enable` is used before that to determine if some multi-cluster operation is needed. But the unused `EnableFederation` field can be removed.

Why not set up the cluster role before the controller started? used to instead `multicluster.enabled`, it's already set up cluster role before joining a member cluster to host cluster.

iawia002:
> Why not set up the cluster role before the controller started?

I still don't get it, it is now set by the cluster controller, how could I set it before the controller started? Looks like a deadlock.

> We can copy the cluster role from clusterconfiguration.

I don't know how and where we could read this `clusterconfiguration`

wansir:
> > Why not set up the cluster role before the controller started?
> 
> I still don't get it, it is now set by the cluster controller, how could I set it before the controller started? Looks like a deadlock.
> 
> > We can copy the cluster role from clusterconfiguration.
> 
> I don't know how and where we could read this `clusterconfiguration`

Cluster role is already set up before kubesphere launched.

https://github.com/kubesphere/ks-installer/blob/master/roles/ks-core/config/templates/kubesphere-config.yaml.j2#L114-L128
https://github.com/kubesphere/ks-installer/blob/master/deploy/cluster-configuration.yaml#L132

PR #4718, opened 2022-03-09
Add "snapshot-count" annotation for volumesnapshotClass

f10atin9:
Signed-off-by: f10atin9 <f10atin9@kubesphere.io>

### What type of PR is this?

/kind feature

In the console, the snapshotclass list is required to display the number of snapshots.

Fixes #

```
None
```

```release-note
None
```

```docs
None
```

### DIFF
--- a/pkg/models/resources/v1alpha3/volumesnapshotclass/volumesnapshotclass.go
+++ b/pkg/models/resources/v1alpha3/volumesnapshotclass/volumesnapshotclass.go
@@ -17,7 +17,10 @@ limitations under the License.
 package volumesnapshotclass
 
 import (
+	"strconv"
 	"strings"
+
+	"k8s.io/apimachinery/pkg/labels"
 
 	v1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
 	"github.com/kubernetes-csi/external-snapshotter/client/v4/informers/externalversions"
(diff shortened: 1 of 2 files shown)


pkg/models/resources/v1alpha3/volumesnapshotclass/volumesnapshotclass.go
```diff
@@ -53,6 +56,11 @@ func (v *volumeSnapshotClassGetter) List(namespace string, query *query.Query) (
 
 	var result []runtime.Object
 	for _, snapshotClass := range all {
+		count := v.countVolumeSnapshots(snapshotClass.Name)
+		if snapshotClass.Annotations == nil {
+			snapshotClass.Annotations = make(map[string]string)
```
dkeven, line 61: Do a deepcopy before mutating the SnapshotClass, as in:
https://github.com/kubesphere/kubesphere/pull/4599

dkeven (changes requested):

pkg/models/resources/v1alpha3/volumesnapshotclass/volumesnapshotclass.go
f10atin9: Okay, I've added it.

pkg/models/resources/v1alpha3/volumesnapshotclass/volumesnapshotclass.go
```diff
@@ -86,3 +95,18 @@ func (v *volumeSnapshotClassGetter) filter(object runtime.Object, filter query.F
 		return v1alpha3.DefaultObjectMetaFilter(snapshotClass.ObjectMeta, filter)
 	}
 }
+
+func (v *volumeSnapshotClassGetter) countVolumeSnapshots(name string) int {
+	snapshots, err := v.informers.Snapshot().V1().VolumeSnapshots().Lister().VolumeSnapshots("").List(labels.Everything())
```
stoneshi-yunify, line 100: is this correct? shouldn't be `v.informers.Snapshot().V1().VolumeSnapshots().Lister().List(labels.Everything())`?

f10atin9: My mistake, it has been changed.

dkeven:
The unit tests should be changed accordingly, too. As it used objects from the altered cache to do assertions before, which is wrong.

PR #4734, opened 2022-03-17
Check if the cluster is the same when updating kubeconfig

iawia002:
### What type of PR is this?

/kind feature

See https://github.com/kubesphere/kubesphere/issues/4426#issuecomment-1068844831 for more info.

Fixes #

```
```

```release-note
NONE
```

```docs

```

/cc @zryfish

### DIFF
--- a/pkg/kapis/cluster/v1alpha1/handler.go
+++ b/pkg/kapis/cluster/v1alpha1/handler.go
@@ -311,6 +311,20 @@ func (h *handler) updateKubeConfig(request *restful.Request, response *restful.R
 		api.HandleBadRequest(response, request, fmt.Errorf("failed to validate member cluster configuration, err: %v", err))
 	}
 
+	// Check if the cluster is the same
+	kubeSystem, err := clientSet.CoreV1().Namespaces().Get(context.TODO(), metav1.NamespaceSystem, metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


pkg/kapis/cluster/v1alpha1/handler.go
```diff
@@ -311,6 +311,20 @@ func (h *handler) updateKubeConfig(request *restful.Request, response *restful.R
 		api.HandleBadRequest(response, request, fmt.Errorf("failed to validate member cluster configuration, err: %v", err))
 	}
 
+	// Check if the cluster is the same
+	kubeSystem, err := clientSet.CoreV1().Namespaces().Get(context.TODO(), metav1.NamespaceSystem, metav1.GetOptions{})
+	if err != nil {
+		api.HandleBadRequest(response, request, err)
+		return
+	}
+	if kubeSystem.UID != cluster.Status.UID {
```
zryfish, line 320: Be careful here, since `cluster.Status.UID` only added after v3.2.1, it may not work in previous versions.

iawia002: No need to worry since this `updateKubeConfig` function is also introduced after v3.2.1, I'm just perfecting it.

PR #4770, opened 2022-04-06
Add storageclass accessor to ks

f10atin9:
### What type of PR is this?

/kind feature

Installing StorageClass-accessor by helm-charts is not the most convenient way, after discussion we decided to add it to the KS.

Fixes #

```
None
```

```release-note
Because validation is required when creating a pvc, it will take a little more time than before.
```

<!--
This section can be blank if this pull request does not require a release not


stoneshi-yunify (commented):
/LGTM

stoneshi-yunify (commented):
/lgtm

PR #4845, opened 2022-05-05
Fix: e2e env setup failed

xyz-li:
### What type of PR is this?

/kind bug

The ks-installer label has changed to `ks-installer`. 

```
```

```release-note
NONE

```

```docs

```
/cc @wansir 
/cc @zheng1 
/cc @pixiake


xyz-li:
ks-installer does not change the `app` label now.

PR #4861, opened 2022-05-09
fix tcp match error

StevenBrown008:
### What type of PR is this?

/kind bug

The tcp match  has several kind, such as port, sourceLabels, gateway, destinationSubnets . We should not just put one port match element in the L4MatchAttributes match array, this may overwrite the original match element, and  cause an error.
So I modify the code ,like the above http match code , just insert the port match in every match element to fix this problem.

Fixes #

```
```

None

```docs

```


iawia002:
The code itself LGTM, this part of the code was last modified in the PR #4560, I don't know why they didn't handle the TCP rules this way in the first place, and I'm not sure if they did it on purpose, this needs to be double-checked by the person responsible for the module.

StevenBrown008:
please /cc @zackzhangkai @zryfish , thanks !

Issue #4807, opened 2022-04-20
Failed to add cluster(Direct Connection)


wenxin-01:
**Describe the Bug**
![image](https://user-images.githubusercontent.com/68640256/164131280-1ee2c3f2-ca1a-495c-a71f-acf657704fa7.png)

`kubectl logs -f kubefed-controller-manager-xxxxx -n kube-federation-system`
![image](https://user-images.githubusercontent.com/68640256/164131376-ff12233a-f3cf-4075-ab6b-3497c1b8ee7d.png)

**Versions Used**
KubeSphere:   `v3.3.0-alpha.1`

/priority high
/assign @iawia002

iawia002:
The description is not accurate, the cluster is successfully added and the status is normal, the problem is caused by the failure of the `ClusterClient` component to cache the cluster:

https://github.com/kubesphere/kubesphere/blob/0086e98288e9761bc3dd8d27007e9640fd967bd5/pkg/utils/clusterclient/clusterclient.go#L202-L210

https://github.com/kubesphere/kubesphere/blob/7d9563dca1e5d4f81615b7a1e650f7c991fe31cd/pkg/apiserver/dispatch/dispatch.go#L64

For now, I do not know why the cache was not successful, the `ks-apiserver` Pod was restarted, the previous logs were not found, we need more logs to locate this problem.

iawia002:
When this happens, `ks-apiserver` has no associated error logs and I don't know what the problem is.

@xyz-li since you are the original author of the `ClusterClient` component, can you help to locate this problem?

/assign @xyz-li

xyz-li:
Interesting. 
After I deleted and added the cluster two times, the bug still exists.
I have no idea about this bug now.

shihaoH:
After I directly connect the ACK cluster, there is also a situation where the cluster does not exist
my issue https://kubesphere.com.cn/forum/d/7122-hostack

wenxin-01:
issue fix in `kubespheredev/ks-apiserver:nightly-20220427`

PR #4873, opened 2022-05-12
Check whether the resources are synchronized successfully after WaitForCacheSync

iawia002:
### What type of PR is this?

/kind bug

The problem in #4807 (#4831) may still exist, this is because `ks-apiserver` started normally, but the k8s lister cache could not query the cluster data (the cluster exists), after restarting the `ks-apiserver` component everything is back to normal, this PR adds more checks for cache, if it fails apiserver will be refused to start (after that it will be restarted automatically by kubelet).

Fixes #

```
```

```release-note
NONE
```

```docs

```

/cc @wansir

### DIFF
--- a/pkg/apiserver/apiserver.go
+++ b/pkg/apiserver/apiserver.go
@@ -406,7 +406,12 @@ func waitForCacheSync(discoveryClient discovery.DiscoveryInterface, sharedInform
 		}
 	}
 	sharedInformerFactory.Start(stopCh)
-	sharedInformerFactory.WaitForCacheSync(stopCh)
+	syncMap := sharedInformerFactory.WaitForCacheSync(stopCh)
+	for informType, synced := range syncMap {
+		if !synced {
+			return fmt.Errorf("resouce %s cache sync failed", informType.String())
+		}
+	}
 	return nil
 }
 


wansir:
It's unnecessary to check whether the resource is synced. WaitForCacheSync is a synchronized method.

> It returns true if it was successful, false if the controller should shutdown

https://github.com/kubesphere/kubesphere/blob/1c96f99072fc12f2c169770675629c89557d3ec3/vendor/k8s.io/client-go/tools/cache/shared_informer.go#L251-L272

iawia002:
I know what `WaitForCacheSync` does, the thing is that we now completely ignore the sync result, and even if a resource fails to sync, we start the entire apiserver as usual, as the comments say:

> false if the controller should shutdown

We should shut it down if some resources fail to sync.

wansir:
> I know what `WaitForCacheSync` does, the thing is that we now completely ignore the sync result, and even if a resource fails to sync, we start the entire apiserver as usual, as the comments say:
> 
> > false if the controller should shutdown
> 
> We should shut it down if some resources fail to sync.

Until `stopCh` is closed, `wait.Pollimmediateuntil `will wait for all resources to be synchronized, your code is duplicated.

Issue #4859, opened 2022-05-09
Ks-controller-manager in KS3.2.1 has been reading hard disks, causing ETCD to hang


yinmin2020:
**Describe the Bug**
The KS-Controller-manager of the master node keeps reading hard disks, causing the ETCD to hang. As a result, I/OS of the host are full, causing the cluster to freeze

For UI issues please also add a screenshot that shows the issue.
![image](https://user-images.githubusercontent.com/65068467/167333898-743387e5-cf5c-4013-9072-4e1fb834195c.png)

**Versions Used**
KubeSphere: 3.2.1
Kubernetes: (If KubeSphere installer used, you can skip this)

**Environment**
How many nodes and their hardware configuration:

For example: CentOS 7.5 / 3 masters:  8cpu/8g; 3 nodes: 8cpu/16g
(and other info are welcomed to help us debugging)

**How To Reproduce**
1. Received an alarm from ali Yunju mainframe
2. Log in to the normal master node and check the fault. The etCD has been reading disks
3. It is found that the component kS-controller-manager has been reading disks, causing the etCD load to be high and the whole host machine to hang
4.View the Ks - controller - manager storage volume information, only hostpath type/var/lib/kubelet/plugins/were being read, enter this folder and any file does not exist

**Expected behavior**
Do you have any troubleshooting ideas for this problem? The current stability can only be maintained by restarting the host to automatically repair the components of View the KS-Controller-Manager

PR #4928, opened 2022-06-01
Add agent to report additional information.

xyz-li:
### What type of PR is this?

/kind feature

If something goes wrong in ks-apiserver or controller-manger, we can only get information from the log.
There is no way to inspect the internal information for the running ks-apiserver and controller-manager now. This PR import gops to help us report additional information such as the current stack trace, Go version, memory stats, etc. This may help us to find ther root cause of #4859.
https://github.com/google/gops

Fixes #

```
NONE
```

```release-note
NONE
```

```docs

```


xyz-li:
If we don't specify the address, gops agent will bind to a random port on address 127.0.0.1.

wansir:
@xyz-li  Debug mode should be disabled by default. `--debug` flag is already provided.

PR #4932, opened 2022-06-01
add some unit tests for pkg/api

qiuming520:
Signed-off-by: qiuming520 <qiuming_yewu@cmss.chinamobile.com>

add some unit tests for pkg/api

Fixes #

```

```

```
None
```

```

```

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,5 +1,6 @@
 ---
 name: Bug report
+labels: ["kind/bug"]
 about: Create a report to help us improve
 ---
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,50 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 3.2.x   | :white_check_mark: |
+| 3.1.x   | :white_check_mark: |
+| 3.0.x   | :white_check_mark: |
+| 2.1.x   | :white_check_mark: |
+| < 2.1.x   | :x:                |
+
+## Reporting a Vulnerability
+
+# Security Vulnerability Disclosure and Response Process
+
+To ensure KubeSphere security, a security vulnerability disclosure and response process is adopted. And the security team is set up in KubeSphere community, also any issue and PR is welcome for every contributors.
+
+The primary goal of this process is to reduce the total exposure time of users to publicly known vulnerabilities. To quickly fix vulnerabilities of KubeSphere, the security team is responsible for the entire vulnerability management process, including internal communication and external disclosure.
+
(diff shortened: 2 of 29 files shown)


pkg/api/alerting/v2alpha1/types_test.go
```diff
@@ -0,0 +1,360 @@
...
+		Items: []*BulkItemResponse{
+			&succBulkItemResponse,
+			&errBulkItemResponse,
+		},
+	}
+)
+
+func TestParseAlertingRuleQueryParams(t *testing.T) {
+	queryParam := "/alerting.kubesphere.io/v2alpha1/rules?name=test-cpu&state=firing&health=ok&sort_field=lastEvaluation&sort_type=desc&label_filters=name~test"
+	req, err := http.NewRequest("GET", fmt.Sprintf("http://localhost?%s", queryParam), nil)
```
xyz-li, line 314: Is this a redundant `?`?

qiuming520: Obtain the http.Request object required by the restful.NewRequest function through the http.NewRequest function (various RULs) . The restful.NewRequest function returns restful.Request , which is the parameter required by the ParseQueryParameter function.  it's necessary.

pkg/api/alerting/v2alpha1/types_test.go
```diff
@@ -0,0 +1,360 @@
...
+	bulkResponse = BulkResponse{
+		Items: []*BulkItemResponse{
+			&succBulkItemResponse,
+			&errBulkItemResponse,
+		},
+	}
+)
+
+func TestParseAlertingRuleQueryParams(t *testing.T) {
+	queryParam := "/alerting.kubesphere.io/v2alpha1/rules?name=test-cpu&state=firing&health=ok&sort_field=lastEvaluation&sort_type=desc&label_filters=name~test"
```
wansir, line 313: `/alerting.kubesphere.io/v2alpha1/rules` is the request path (relative paths may omit leading slash).

> An HTTP URL takes the form:
> http://<host>:<port>/<path>?<searchpart>
> where <host> and <port> are as described in Section 3.1. If :<port>
is omitted, the port defaults to 80.  No user name or password is
allowed.  <path> is an HTTP selector, and <searchpart> is a query
string. The <path> is optional, as is the <searchpart> and its
preceding "?". If neither <path> nor <searchpart> is present, the "/"
may also be omitted.

https://www.rfc-editor.org/rfc/rfc1738#section-3.3

qiuming520: Right! Thanks for your review, I have fixed it 。

pkg/api/alerting/v2alpha1/types_test.go
wansir, line 314:
```suggestion
	queryParam := "name=test-cpu&state=firing&health=ok&sort_field=lastEvaluation&sort_type=desc&label_filters=name~test"
	req, err := http.NewRequest("GET", fmt.Sprintf("http://localhost/alerting.kubesphere.io/v2alpha1/rules?%s", queryParam), nil)
```

qiuming520: Thanks for suggestion ！ fixed and committed， help review again plz @wansir

Issue #4934, opened 2022-06-01
Enabling servicemesh breaks ks-apiserver and prevents kubeshphere from functioning


brandan-schmitz:
### Describe the Bug
When setting enabling the service mesh on a new install (environment details below) the pod created by ks-apiserver fails to start and goes into a crash loop with a runtime error. It should be noted that installation occurs fine and the apiserver runs without issue if the servicemesh is set to false. Installing the servicemesh after the cluster is already running as described in the 
`Enable Pluggable Components` section of the documentation also results in this issue which then breaks the installation.

```
W0601 18:15:22.577233       1 client_config.go:615] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
W0601 18:15:22.581954       1 client_config.go:615] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I0601 18:15:22.881396       1 interface.go:60] start helm repo informer
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
kubesphere.io/kubesphere/pkg/kapis/servicemesh/metrics/v1alpha2.NewHandler(0xc0009c1d80, 0x41eea38, 0xc0008f2160, 0x41ae580, 0xc000556358, 0xd)
	/workspace/pkg/kapis/servicemesh/metrics/v1alpha2/handler.go:51 +0x57b
kubesphere.io/kubesphere/pkg/kapis/servicemesh/metrics/v1alpha2.AddToContainer(0xc0009c1d80, 0xc000c81320, 0x41eea38, 0xc0008f2160, 0x41ae580, 0xc000556358, 0xc000847ea0, 0x412d800)
	/workspace/pkg/kapis/servicemesh/metrics/v1alpha2/register.go:42 +0xfa
kubesphere.io/kubesphere/pkg/apiserver.(*APIServer).installKubeSphereAPIs(0xc000814000)
	/workspace/pkg/apiserver/apiserver.go:247 +0x12d4
kubesphere.io/kubesphere/pkg/apiserver.(*APIServer).PrepareRun(0xc000814000, 0xc0008ca000, 0xc000814000, 0x0)
	/workspace/pkg/apiserver/apiserver.go:172 +0x171
kubesphere.io/kubesphere/cmd/ks-apiserver/app.Run(0xc000aaf410, 0x4199eb8, 0xc000ac65c0, 0x0, 0x0)
[...]

PR #4940, opened 2022-06-06
create default token for service account

xyz-li:
### What type of PR is this?

/kind bug

After kubernetes v1.24, kube-conteoller-manager will not create the default token for service account.
This pr will create a default token for cluster service account.

#4934

```
NONE
```

```release-note
NONE
```

```docs

```
/cc @wansir 
/cc @zheng1 
/cc @iawia002

### DIFF
--- a/pkg/controller/cluster/join.go
+++ b/pkg/controller/cluster/join.go
@@ -18,6 +18,7 @@ package cluster
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"time"
 
@@ -268,7 +269,7 @@ func createAuthorizedServiceAccount(joiningClusterClientset kubeclient.Interface
 
 	klog.V(2).Infof("Creating service account in joining cluster: %s", joiningClusterName)
 
-	saName, err := createServiceAccount(joiningClusterClientset, namespace,
+	saName, err := createServiceAccountWithSecret(joiningClusterClientset, namespace,
 		joiningClusterName, hostClusterName, dryRun, errorOnExisting)
 	if err != nil {
 		klog.V(2).Infof("Error creating service account: %s in joining cluster: %s due to: %v",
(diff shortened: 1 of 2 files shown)


pkg/controller/cluster/join.go
```diff
@@ -320,25 +321,54 @@ func createAuthorizedServiceAccount(joiningClusterClientset kubeclient.Interface
...
+	if dryRun {
+		return saName, nil
+	}
+
+	secretName := fmt.Sprintf("%s-sa-token", saName)
+
+	secret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      saName,
+			Name:      secretName,
```
wansir, line 339:
```suggestion
			GenerateName:     fmt.Sprintf("%s-token-", saName),
```

pkg/controller/cluster/join.go
```diff
@@ -320,25 +321,54 @@ func createAuthorizedServiceAccount(joiningClusterClientset kubeclient.Interface
...
+		return "", err
+	}
+
+	sa := &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      saName,
+			Namespace: namespace,
+		},
+		Secrets: []corev1.ObjectReference{
+			{Name: secretName},
```
wansir, line 366:
```suggestion
			{Name: secret.Name},
```

pkg/controller/cluster/join.go
```diff
@@ -320,31 +321,79 @@ func createAuthorizedServiceAccount(joiningClusterClientset kubeclient.Interface
...
+			return "", err
+		}
+	}
+
+	if len(sa.Secrets) > 0 {
+		if _, err := clusterClientset.CoreV1().Secrets(namespace).Get(ctx, sa.Secrets[0].Name, metav1.GetOptions{}); err == nil {
+			return saName, nil
+		} else if !apierrors.IsNotFound(err) {
+			return "", err
+		}
```
wansir, line 367:
```suggestion
		return saName, nil
```

If the secret name already exists, double-checking is unnecessary. If the secret does not exist, `populateSecretInHostCluster` will report the not found error.

PR #5018, opened 2022-06-30
[KS-OIDC] Remove special characters form sub OIDC standard claim

najib-houcine:
#### Issue
Users cannot login using an OIDC identity provider.

#### Kubesphere version
3.2.1

#### Error Message
Logs form ks-apiserver 
```json
{
  code: 422,
  statusText: 'Unprocessable Entity',
  message: `User.iam.kubesphere.io "{UNLUCKY-USER}" is invalid: metadata.labels: Invalid value: "{UNLUCK-USER_OIDC_SUB-StartOrEnd-Special-Char}": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')\n`
}
```

#### Cause
The standard claim **sub** in OIDC specs https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims is used as label ``` iam.kubesphere.io/origin-uid ```  in  https://github.com/kubesphere/kubesphere/blob/master/pkg/models/auth/authenticator.go#L82  , for the case OIDC as identity  provided referenced in https://github.com/kubesphere/kubesphere/blob/master/pkg/apiserver/authentication/identityprovider/oidc/oidc.go#L118

Kubernetes  doesn't allow special characters in start and end of labels value as for https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

#### Proposed Workaround  
Since is OIDC claim cannot  be substituted  by another claim , the proposed  solution is to replace any special characters at start or end with **0**

#### Solution
Replace the label ``` iam.kubesphere.io/origin-uid ```  with  unique generated UID  based on combining OIDC claims while respecting Kubernetes  labels standards.

### DIFF
--- a/pkg/apiserver/authentication/identityprovider/oidc/oidc_test.go
+++ b/pkg/apiserver/authentication/identityprovider/oidc/oidc_test.go
@@ -198,7 +198,7 @@ var _ = Describe("OIDC", func() {
 			req := &http.Request{URL: url}
 			identity, err := provider.IdentityExchangeCallback(req)
 			Expect(err).Should(BeNil())
-			Expect(base64.URLEncoding.DecodeString(identity.GetUserID())).Should(Equal("110169484474386276334"))
+			Expect(string(base64.URLEncoding.DecodeString(identity.GetUserID()))).Should(Equal("110169484474386276334"))
 			Expect(identity.GetUsername()).Should(Equal("test"))
 			Expect(identity.GetEmail()).Should(Equal("test@kubesphere.io"))
 		})
--- a/pkg/apiserver/authentication/identityprovider/oidc/oidc_test.go
+++ b/pkg/apiserver/authentication/identityprovider/oidc/oidc_test.go
@@ -198,7 +198,7 @@ var _ = Describe("OIDC", func() {
 			req := &http.Request{URL: url}
 			identity, err := provider.IdentityExchangeCallback(req)
 			Expect(err).Should(BeNil())
-                        Expect(identity.GetUserID()).Should(Equal(base64.RawURLEncoding.EncodeToString([]byte("110169484474386276334"))))
+			Expect(identity.GetUserID()).Should(Equal(base64.RawURLEncoding.EncodeToString([]byte("110169484474386276334"))))
 			Expect(identity.GetUsername()).Should(Equal("test"))
 			Expect(identity.GetEmail()).Should(Equal("test@kubesphere.io"))
 		})
--- a/pkg/apiserver/authentication/identityprovider/oidc/oidc_test.go
+++ b/pkg/apiserver/authentication/identityprovider/oidc/oidc_test.go
@@ -198,7 +198,7 @@ var _ = Describe("OIDC", func() {
 			req := &http.Request{URL: url}
 			identity, err := provider.IdentityExchangeCallback(req)
 			Expect(err).Should(BeNil())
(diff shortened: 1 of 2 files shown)


wansir:
@najib-houcine Thanks for your contribution. These changes seem to have some risks, removing the special characters directly would cause duplication of uid, it cannot be tolerated.  It's better to use `base64.RawURLEncoding.EncodeToString` here.

najib-houcine:
@wansir I took a look into it and base64 encoding will grantee an **==** at the end of orgin_uid label , which will fail any user with good or bad Sub.

wansir:
> @wansir I took a look into it and base64 encoding will grantee an **==** at the end of orgin_uid label , which will fail any user with good or bad Sub.

> RawURLEncoding is the unpadded alternate base64 encoding defined in [RFC 4648](https://rfc-editor.org/rfc/rfc4648.html). It is typically used in URLs and file names. This is the same as URLEncoding but omits padding characters.

https://pkg.go.dev/encoding/base64@master#pkg-variables

BTW, the length of base64-encoded uid will be limited to 63 characters, I think it can be accepted. And we should convert the uid labels after the upgrade.

najib-houcine:
Thanks @wansir , I updated the PR

wansir:
@najib-houcine Please take a look at the failed CI jobs.

>FAILED   verify-gofmt.sh	2s
Verifying verify-goimports.sh
The following files are not import formatted 
./pkg/apiserver/authentication/identityprovider/oidc/oidc.go
Please run the following command:
  make goimports

pkg/apiserver/authentication/identityprovider/oidc/oidc_test.go
```diff
@@ -198,7 +198,7 @@ var _ = Describe("OIDC", func() {
 			req := &http.Request{URL: url}
 			identity, err := provider.IdentityExchangeCallback(req)
 			Expect(err).Should(BeNil())
-			Expect(identity.GetUserID()).Should(Equal("110169484474386276334"))
+			Expect(string(base64.URLEncoding.DecodeString(identity.GetUserID()))).Should(Equal("110169484474386276334"))
```
wansir, line 201:
```suggestion
			Expect(identity.GetUserID()).Should(Equal(base64.RawURLEncoding.EncodeToString([]byte("110169484474386276334"))))
```

wansir:
> @wansir , Thanks I checked and managed to fix the verification step but unit-test keep on-failing , could you check what wrong ?

> vet: pkg/apiserver/authentication/identityprovider/oidc/oidc_test.go:201:18: 2-valued base64.URLEncoding.DecodeString(identity.GetUserID()) (value of type ([]byte, error)) where single value is expected

@najib-houcine Refer to https://prow.kubesphere.io/view/s3/prow-logs/pr-logs/pull/kubesphere_kubesphere/5018/pull-kubesphere-unit-test/1543948976712060928

najib-houcine:
Thanks @wansir  for the hint , but unit test failed also https://prow.kubesphere.io/view/s3/prow-logs/pr-logs/pull/kubesphere_kubesphere/5018/pull-kubesphere-unit-test/1544265767887835136 I will investigate more into it.

wansir:
@najib-houcine 

You can reformat the code using `make goimports`

najib-houcine:
@wansir seems good now .

PR #5044, opened 2022-07-08
intergrate opensearch v2

wenchajun:
Signed-off-by: chengdehao <dehaocheng@kubesphere.io>

### What type of PR is this?

/kind feature

Fixes #

```
```

```release-note
None
```

```docs

```


pkg/simple/client/es/client.go
```diff
@@ -34,9 +36,10 @@ import (
 )
 
 const (
-	ElasticV5 = "5"
-	ElasticV6 = "6"
-	ElasticV7 = "7"
+	ElasticV5    = "5"
+	ElasticV6    = "6"
+	ElasticV7    = "7"
+	OpenSearchV2 = "2"
```
benjaminhuo, line 42: The version can be changed like below to distinguish it with elasticsearch versions
```suggestion
	OpenSearchV2 = "OpenSearchV2"
```

wenchajun: We determine the version by the `version` field in the data returned in the response, which, if set up this way, is not well handled on the back end.
https://github.com/kubesphere/kubesphere/blob/master/pkg/simple/client/es/client.go#L125

pkg/simple/client/es/versions/opensearch/v2/v2.go
```diff
@@ -0,0 +1,128 @@
+/*
+Copyright 2020 KubeSphere Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
```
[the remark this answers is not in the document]
wenchajun: fixed

pkg/simple/client/es/client.go
benjaminhuo: Is there some field returned in the response that contains `opensearch`?

wenchajun: This field is returned, so this constant I think is better set to `2`

benjaminhuo: I mean the distribution field contains "opensearch"
So if this field is "opensearch", we can add a `opensearchv` prefix to the version 2 and that will make the version become `opensearchv2`

wenchajun: a5b793b04e2efd33257844887610184
I tested the return fields of elasticsearch and there is no `distribution` field in es, which should be the difference between elasticsearch and OpenSearch.

benjaminhuo: So when there is distribution field and the value is opensearch, we can add `opensearchv` prefix to the v

PR #5045, opened 2022-07-08
Upgrade go version to 1.18

wansir:
### What type of PR is this?

Upgrade go version to 1.18.

Fixes #

```
```

```release-note
None
```

```docs

```


LICENSES/LICENSE
```diff
@@ -0,0 +1,207 @@
+================================================================================
+= KubeSphere licensed under: =
```
iawia002, line 2: This format looks weird, why is there a `:`

wansir: The script is copied from https://github.com/kubernetes/kubernetes/blob/master/hack/update-vendor-licenses.sh. The license directory is helpful for us.

iawia002: Got it.

hack/update-vendor-licenses.sh
```diff
@@ -155,55 +155,80 @@ fi
...
+mkdir -p "${LICENSES_DIR}"
+
+# The tmp directory to save all the LICENSE files, will move to LICENSES_DIR
+TMP_LICENSES_DIR="${KUBE_TEMP}/LICENSES.DIR.$$"
+mkdir -p "${TMP_LICENSES_DIR}"
+
 DEPS_DIR="vendor"
 declare -Ag CONTENT
 
 # Put the K8S LICENSE on top
```
xyz-li: `Put the KubeSphere LICENSE on top`

wansir: Updated

PR #5075, opened 2022-07-20
Adjust container terminal priority: bash, sh

### LINKED ISSUE #5067: Adjust the priority of bash and sh in the container terminal
**What's it about?**

Adjust the priority of bash and sh in the container terminal.

Bash should have a higher priority than other shells.

**Area Suggestion**

/area apiserver
/kind feature-request
/good-first-issue

tal66:
Hi, this fixes #5067 : Adjust the priority of bash and sh in the container terminal.

### DIFF
--- a/pkg/models/terminal/terminal.go
+++ b/pkg/models/terminal/terminal.go
@@ -328,7 +328,7 @@ func isValidShell(validShells []string, shell string) bool {
 
 func (t *terminaler) HandleSession(shell, namespace, podName, containerName string, conn *websocket.Conn) {
 	var err error
-	validShells := []string{"sh", "bash"}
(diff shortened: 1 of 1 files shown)


wansir:
@tal66 Thanks for your contribution!

/lgtm
/approve
/cherry-pick release-3.3

Gerrylinux:
Front end code file：（console/src/stores/terminal.js）
When entering the terminal, access the websocket interface, and the shell seems to be set to sh,
The back-end code only adjusts and modifies the data priority, which should not be realized
The retry logic of the back-end terminal. When testing the first bash, the failure to enter the container terminal will cause the websocket link to close. The SH terminal environment will use the closed websocket link
image

PR #5126, opened 2022-08-09
update helm release annotation inject key

### LINKED ISSUE #4911: The annotation of openelb error
**Describe the Bug**

Request URL:  `/kapis/iam.kubesphere.io/v1alpha2/clusterroles?sortBy=createTime&limit=10&annotation=kubesphere.io%2Fcreator`

openelb should not have this annotation.

image

**Versions Used**
KubeSphere:   `kubespheredev/ks-installer:nightly-20220524`

x893675:

Fixes #4911

```
Not only openelb chart , the helm chart that contains the clusterrole template will have the same problem
```

```release-note
None
```

```docs

```

### DIFF
--- a/pkg/controller/openpitrix/helmrelease/helm_release_controller.go
+++ b/pkg/controller/openpitrix/helmrelease/helm_release_controller.go
@@ -21,6 +21,8 @@ import (
 	"errors"
 	"time"
 
+	"kubesphere.io/kubesphere/pkg/constants"
+
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/flowcontrol"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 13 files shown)


pkg/controller/openpitrix/helmrelease/helm_release_controller.go
```diff
@@ -339,7 +338,6 @@ func (r *ReconcileHelmRelease) createOrUpgradeHelmRelease(rls *v1alpha1.HelmRele
 
 	// If clusterConfig is empty, this application will be installed in current host.
 	hw := helmwrapper.NewHelmWrapper(clusterConfig, rls.GetRlsNamespace(), rls.Spec.Name,
-		helmwrapper.SetAnnotations(map[string]string{constants.CreatorAnnotationKey: rls.GetCreator()}),
```
xyz-li, line 342: We can't simply remove this field, front end needs this value to display the creator of the app, you can change the key to apps.kubesphere.io/creator.

x893675:
fixed.. @xyz-li

xyz-li:
Please change the key in the function `GetCreator` too.
https://github.com/kubesphere/kubesphere/blob/587dbe723bf80f5c96ab1d49aa92c4c2311a8bef/staging/src/kubesphere.io/api/application/v1alpha1/helmrelease_types.go#L119-L121

xyz-li:
There 5 `GetCreator` functions in dir staging/src/kubesphere.io/api/application/v1alpha1. You need to change them all.

x893675:
QQ20220811-131004@2x

front end needs update too

x893675:
@xyz-li

xyz-li:
> QQ20220811-131004@2x
> 
> when resource is rendered by helm release, front end needs use new annotation key

This is a dilemma.

xyz-li:
/hold

We need to find a better way to deal with this annotation.

PR #5131, opened 2022-08-10
Set ready status to false when the cluster join federation failed

### LINKED ISSUE #5125: When member cluster disconnect from host, kubeSphere still show the cluster's Ready status as "True"
**Describe the Bug**
after the member cluster has been disconnected from host cluster for a long time,  kubesphere still show the cluster's Ready status as "True". 

I haven't get this skill.

**Versions Used**
KubeSphere: v3.3.0
Kubernetes: (If KubeSphere installer used, you can skip this) v1.21.0

**Environment**
2 nodes for the host cluster, and 2 nodes for the member cluster.

For example: CentOS 7.5 / 3 masters:  8cpu/8g; 3 nodes: 8cpu/16g
(and other info are welcomed to help us debugging)
CentOS 7.9

**How To Reproduce**
1、Add a member cluster normally，make sure everything is ok；
2、Disconnect member cluster from KubeSphere by network unreachable or shutdown nodes...
3. Go to KubeSphere 'Platform-Manage' -> 'Cluster-Manage'. The displayed page shows that the member cluster's status is OK.

5.  The response of url '/kapis/resources.kubesphere.io/v1alpha3/clusters?limit=10&labelSelector=%21cluster-role.kubesphere.io%2Fhost&sortBy=createTime'  is also 'True'
image

7. When click into the member cluster. 'connect: no route to host' error was sent, and this is reasonable.

**Expected behavior**
fix it

iawia002:
### What type of PR is this?

/kind bug

Currently, when a cluster is disconnected, we only update the `Federated` status, but not the `Ready` status, which causes the Ready status to remain true, this patch fixes this issue.

Fixes #5125

```
```

```release-note
Set ready status to false when the cluster join federation failed.
```

```docs

```

/cc @wansir

### DIFF
--- a/pkg/controller/cluster/cluster_controller.go
+++ b/pkg/controller/cluster/cluster_controller.go
@@ -421,6 +421,15 @@ func (c *clusterController) syncCluster(key string) error {
 				Message:            "Cluster can not join federation control plane",
 			}
 			c.updateClusterCondition(cluster, federationNotReadyCondition)
+			notReadyCondition := clusterv1alpha1.ClusterCondition{
+				Type:               clusterv1alpha1.ClusterReady,
+				Status:             v1.ConditionFalse,
+				LastUpdateTime:     metav1.Now(),
+				LastTransitionTime: metav1.Now(),
+				Reason:             err.Error(),
+				Message:            "Cluster is not available now",
+			}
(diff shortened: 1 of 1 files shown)


iawia002:
Closing this PR in favor of #5130

PR #5134, opened 2022-08-12
enhance globalrulegroups

junotx:
### What type of PR is this?

/kind feature

new supports for globalrulegroups as follows:

- configure template rules (by `exprBuilder`)
- limit rule scope (by `clusterSelector`, `namespaceSelector`)
- filter built-in rule groups and their alerts

Fixes #

```
```

```release-note
Support template rule and scope limit for globalrulegroups
```

```docs

```


pkg/api/alerting/v2beta1/types.go
```diff
@@ -69,6 +71,7 @@ type RuleGroupStatus struct {
 }
 
 type RuleStatus struct {
+	Expr           string     `json:"expr,omitempty" description:"expression evaluated actually, provided only for global rules"`
```
benjaminhuo, line 74:
```suggestion
	Expr           string     `json:"expr,omitempty" description:"expression evaluated, for global rules only"`
```

benjaminhuo (changes requested):

PR #5182, opened 2022-08-30
fix bug helm repo paging query

### LINKED ISSUE #5181: HelmRepo paging query error
**Describe the Bug**

You don't need to remove this comment section, it's invisible on the issues page.

**When the added HelmRepo is more than 10, the paging query error.**

**Versions Used**
KubeSphere: v3.2.1

image

image

yongxingMa:
### What type of PR is this?
/kind bug

Fixes #5181 

```
```

```release-note
fix bug helm repo paging query
```

```docs

```

### DIFF
--- a/pkg/models/openpitrix/repos.go
+++ b/pkg/models/openpitrix/repos.go
@@ -298,8 +298,10 @@ func (c *repoOperator) ListRepos(conditions *params.Conditions, orderBy string, 
 	}
 
 	totalCount := len(repos)
+	start, end := (&query.Pagination{Limit: limit, Offset: offset}).GetValidPagination(totalCount)
+	repos = repos[start:end]
 	items := make([]interface{}, 0, len(repos))
-	for i, j := offset, 0; i < len(repos) && j < limit; i, j = i+1, j+1 {
+	for i := range repos {
 		items = append(items, convertRepo(repos[i]))
 	}
 	return &models.PageableResponse{Items: items, TotalCount: totalCount}, nil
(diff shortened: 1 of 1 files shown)


pkg/models/openpitrix/repos.go
```diff
@@ -298,8 +298,6 @@ func (c *repoOperator) ListRepos(conditions *params.Conditions, orderBy string,
 	}
 
 	totalCount := len(repos)
-	start, end := (&query.Pagination{Limit: limit, Offset: offset}).GetValidPagination(totalCount)
-	repos = repos[start:end]
 	items := make([]interface{}, 0, len(repos))
 	for i, j := offset, 0; i < len(repos) && j < limit; i, j = i+1, j+1 {
```
mangoGoForward, line 302: I think we should not to remove those line, but should fix the for loop.
when we want to get the second page, the `start` is 10, `end` is 11, `len(repos)` is 1 and `i` is 10, so if `i < len(repos)` is false

mangoGoForward (changes requested):

pkg/models/openpitrix/repos.go
xyz-li: How about this?
```
	start, end := (&query.Pagination{Limit: limit, Offset: offset}).GetValidPagination(totalCount)
	repos = repos[start:end]
	items := make([]interface{}, 0, len(repos))
	for i := range repos {
		items = append(items, convertRepo(repos[i]))
	}
```

mangoGoForward: more better

yongxingMa: > How about this?
> 
> ```
> 	start, end := (&query.Pagination{Limit: limit, Offset: offset}).GetValidPagination(totalCount)
> 	repos = repos[start:end]
> 	items := make([]interface{}, 0, len(repos))
> 	for i := range repos {
> 		items = append(items, convertRepo(repos[i]))
> 	}
> ```

Well, this way.

PR #5293, opened 2022-10-14
Check the cluster ID (kube-system UID)  when updating the cluster

yzxiu:
Signed-off-by: yzx <946666800@qq.com>

### What type of PR is this?
Add one of the following kinds:
/kind feature

When manual updating the cluster crd resource, check the cluster ID (kube-system UID)  to avoid heavy losses caused by copying wrong files.Because kubefed automatically performs a lot of irreversible operations

```release-note
None
```


wansir:
@yzxiu It is better to add an admission webhook to check cluster ID (kube-system UID) before updating.

iawia002:
> @yzxiu It is better to add an admission webhook to check cluster ID (kube-system UID) before updating.

I agree this is a more reasonable solution.

yzxiu:
@iawia002 I agree with this solution and have updated the code.
I have tested in my own cluster,like


This also requires adding a validatingwebhookconfigurations,like
```yaml
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingWebhookConfiguration
metadata:
  annotations:
    meta.helm.sh/release-name: ks-core
    meta.helm.sh/release-namespace: kubesphere-system
  labels:
    app.kubernetes.io/managed-by: Helm
  name: cluster.kubesphere.io
webhooks:
- admissionReviewVersions:
  - v1beta1
  clientConfig:
    caBundle: <data>
    service:
      name: ks-controller-manager
      namespace: kubesphere-system
      path: /validate-cluster-kubesphere-io-v1alpha1
      port: 443
  failurePolicy: Fail
  matchPolicy: Exact
  name: validating-cluster.kubesphere.io
  namespaceSelector:
    matchExpressions:
    - key: control-plane
      operator: DoesNotExist
  objectSelector: {}
  rules:
  - apiGroups:
    - cluster.kubesphere.io
    apiVersions:
    - v1alpha1
    operations:
    - UPDATE
    resources:
    - clusters
    scope: '*'
  sideEffects: None
  timeoutSeconds: 30
```
I also submitted it in ks-installer https://github.com/kubesphere/ks-installer/pull/2131

pkg/controller/cluster/cluster_controller.go
```diff
@@ -142,6 +144,28 @@ type clusterController struct {
 	hostClusterName string
 }
 
+func (c *clusterController) ValidateUpdate(old runtime.Object, new runtime.Object) error {
```
iawia002, line 147: I didn't get it, why does the webhook need to be related to the controller, it should be independent.

pkg/controller/cluster/cluster_controller.go
```diff
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
+	"k8s.io/apimachinery/pkg/runtime"
```
iawia002, line 26: This should be moved to the next group.

pkg/controller/cluster/cluster_controller.go
yzxiu: It references existing webhook code
https://github.com/kubesphere/kubesphere/blob/master/pkg/controller/network/ippool/ippool_controller.go

yzxiu:
@iawia002 @wansir Thanks for your patient review, I have simplified the code

pkg/controller/cluster/cluster_controller.go
yzxiu: Thanks for your patient review, I have simplified the code

pkg/controller/cluster/cluster_webhook.go
```diff
@@ -0,0 +1,78 @@
...
+// InjectDecoder injects the decoder into a ValidatingHandler.
+func (h *ValidatingHandler) InjectDecoder(d *admission.Decoder) error {
+	h.decoder = d
+	return nil
+}
+
+// Handle handles admission requests.
+func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
+
+	if req.Operation == v1.Update {
```
iawia002, line 48: We can do things like this:

```go
if req.Operation != v1.Update {
    return admission.Allowed("")
}

...

```

we can move the main logic out of the `if` condition.

iawia002 (commented):
Thanks @yzxiu for doing this, basically LGTM

iawia002:
Thanks @yzxiu 

/lgtm

PR #5300, opened 2022-10-18
support fieldselector filter query secrets

nuclearwu:
Signed-off-by: wuzhongjian <wuzhongjian_yewu@cmss.chinamobile.com>

### What type of PR is this?

/kind feature

You can query the secret list according to the secret type by supporting fieldSelector filtering

Fixes #

```
```

```release-note
None
```

```docs

```

### DIFF
--- a/pkg/models/resources/v1alpha3/secret/secrets.go
+++ b/pkg/models/resources/v1alpha3/secret/secrets.go
@@ -18,6 +18,7 @@ package secret
 
 import (
 	"encoding/json"
+	"fmt"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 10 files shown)


hongzhouzi:
There seems to be some problem with the import format, you need to run `make goimports` command.

nuclearwu:
> There seems to be some problem with the import format, you need to run `make goimports` command.

ok, done

pkg/models/resources/v1alpha3/secret/secrets.go
```diff
@@ -74,5 +76,10 @@ func (s *secretSearcher) filter(object runtime.Object, filter query.Filter) bool
 		return false
 	}
 
+	if filter.Field == query.ParameterFieldSelector &&
+		strings.Contains(string(filter.Value), string(secret.Type)) {
```
wansir, line 80: Why not use `jsonpath.JsonPathLookup` instead? It can be a generic query filter, just go ahead. As we know, the field selector is still not working well with custom resources, https://github.com/kubernetes/kubernetes/issues/51046.

nuclearwu: Is this what you mean? https://github.com/oliveagle/jsonpath

nuclearwu: > Why not use `jsonpath.JsonPathLookup` instead? It can be a generic query filter, just go ahead. As we know, the field selector is still not working well with custom resources, [kubernetes/kubernetes#51046](https://github.com/kubernetes/kubernetes/issues/51046).

ok, done

pkg/models/resources/v1alpha3/secret/secrets.go
```diff
@@ -74,5 +79,35 @@ func (s *secretSearcher) filter(object runtime.Object, filter query.Filter) bool
...
+	strValues := strings.Split(string(value), "=")
+	res, err := jsonpath.JsonPathLookup(jsonData, "$."+strValues[0])
+	if err != nil {
+		klog.Error(err)
+		return false
+	}
+	if res != strValues[1] {
+		return false
+	}
+	return true
```
wansir, line 112:
```suggestion
selectors := strings.Split(string(fieldSelector), ",")
	for _, selector := range selectors {
		var negative bool
		var key string
		var value string
		// supports '=', and '!='.(e.g. ?fieldSelector=key1=value1,key2=value2)
		if strings.Contains(selector, "!=") {
			negative = true
			keyValues := strings.Split(selector, "!=")
			key = keyValues[0]
			value = keyValues[1]
		} else {
			negative = false
			keyValues := strings.Split(selector, "=")
			key = keyValues[0]
			value = keyValues[1]
		}
		var input map[string]interface{}
		data, err := json.Marshal(secret)
		if err != nil {
			klog.V(4).Infof("failed marshal to JSON string: %s", err)
			return false
		}
		if err = json.Unmarshal(data, &input); err != nil {
			klog.V(4).Infof("failed unmarshal to map object: %s", err)
			return false
		}
		rawValue, err := jsonpath.JsonPathLookup(input, "$."+key)
		if err != nil {
			klog.V(4).Infof("failed to lookup jsonpath: %s", err)
			return false
		}
		if (negative && fmt.Sprintf("%v", rawValue) != value) || (!negative && fmt.Sprintf("%v", rawValue) == value) {
			continue
		} else {
			return false
		}
	}
```

nuclearwu: Done

pkg/models/resources/v1alpha3/secret/secrets.go
```diff
@@ -74,5 +80,53 @@ func (s *secretSearcher) filter(object runtime.Object, filter query.Filter) bool
...
+			negative = true
+			keyValues := strings.Split(selector, "!=")
+			key = keyValues[0]
+			value = keyValues[1]
+		} else {
+			negative = false
+			keyValues := strings.Split(selector, "=")
+			key = keyValues[0]
+			value = keyValues[1]
+		}
```
wansir, line 109: Out of range exception should be caught.

nuclearwu: > @nuclearwu Out of range exception should be caught.

@wansir Done, review please.

PR #5416, opened 2022-12-14
ResourceGetter v1beta1

zhou1203:
### What type of PR is this?

/kind feature
/kind api-change

1. This PR helps us reduce the resource-related code which is redundant, we no longer need to write a separate `ResourcesGetter` for each Resource. so that I add the new api group /resources/v1beta1 can get any register or not register CRD. Moreover, the CRD created during KubeSphere running can also be cached and getted.

2. Adding the fieldSelector for filtering resources by any field. I used the benchmark to test the performance of fieldSelector. There is a detailed testing report.

Fixes #

```
```

```release-note
Add resources v1beta1 for getting/cacheing unregistered resources，add fieldSelector to filter resources by any field
```

```docs
Get/List Resource with module resources v1beta1:
Cluster Resources: `/kapis/{group}/{version}/{resources}` 
Namespaced Resources: `/kapis/{group}/{version}/namespaces/{namespace}/{resources}`

List Resources using fieldSelector:
List /kapis/{group}/{version}/{resources}?fieldSelector={field path}={field name}
```


pkg/models/resources/v1beta1/cache.go
```diff
@@ -0,0 +1,106 @@
...
+			return false
+		}
+		if (negative && fmt.Sprintf("%v", rawValue) != value) || (!negative && fmt.Sprintf("%v", rawValue) == value) {
+			continue
+		} else {
+			return false
+		}
+	}
+	return true
+}
```
wansir, line 106: Performance needs to be considered.

PR #5532, opened 2023-02-17
Use golang-jwt/jwt instead of form3tech-oss/jwt-go

wansir:
### What type of PR is this?
/kind cleanup

Related to https://github.com/kubernetes/kubernetes/issues/104007

```release-note
None
```


zhou1203 (approved):
/igtm

zhou1203 (approved):
/approve

PR #5564, opened 2023-03-03
call validate when config update coz apiserver restart

Fritzmomoto:
### What type of PR is this?
/kind bug

I made apiserver restart by adding keycloak-related configuration to clusterconfiguration. However, I found that the keycloak-related configuration was not read correctly as expected. This was because oidcProviderFactory.Create() was not called. I traced back the specific call chain: ServerRunOptions.Validate -> AuthenticationOptions.Validate -> identityprovider.SetupWithOptions -> factory.Create. Therefore, I added s.Validate() before run(s, ictx).

Fixes #

```release-note
Validate the configuration after reloading.
```


hongzhouzi:
I think this is a temporary solution. In this preliminary validation phase it makes sense to do only validation operations, and `factory.Create` in `identityprovider.SetupWithOptions` should be put in the run phase.

PR #5603, opened 2023-03-22
fix: Function NewTerms in ES DSL builder: panic if the value is string

### LINKED ISSUE #5500: Function NewTerms in ES-DSL-Builder: It will panic when the val is string.
https://github.com/kubesphere/kubesphere/blob/c8e131fc136c360a3a33ed28752f418016e5f9e5/pkg/simple/client/es/query/builder.go#L460

jinqinn:
### What type of PR is this?
/kind bug

Fix the bug of NewTerms() in ES DSL builder: panic if the value is string

Fixes #5500 

```
```

```release-note
None
```

```docs

```

### DIFF
--- a/pkg/simple/client/es/query/builder.go
+++ b/pkg/simple/client/es/query/builder.go
@@ -460,8 +460,15 @@ func (m *Terms) IsValid() bool {
 
 func NewTerms(key string, val interface{}) *Terms {
 
-	if reflect.ValueOf(val).IsNil() {
+	if val == nil {
 		return nil
+	}
+
+	switch reflect.TypeOf(val).Kind() {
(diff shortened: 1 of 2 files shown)


pkg/simple/client/es/query/builder.go
```diff
@@ -460,10 +460,17 @@ func (m *Terms) IsValid() bool {
 
 func NewTerms(key string, val interface{}) *Terms {
 
-	if reflect.ValueOf(val).IsNil() {
+	if val == nil {
```
wanjunlei, line 463: The `terms` query only accepts slice parameters, so here it is only necessary to check if `val` is nil or an empty slice. The `switch` statement below can also be omitted.

PR #5615, opened 2023-03-30
Update utils.go

chuan-you:
All build alerting is not working

### What type of PR is this?
/kind bug

For buildin alerting rules, fileshort path has a postfix that will never match the rule resourcename. Therefore, all buildin alerting rules will not fire since no rule group is matched.  Following are logs for debugging:

E0330 16:17:15.056762       1 utils.go:247] GetAlertingRuleStatus fileshortkubesphere-monitoring-system-prometheus-k8s-rules-46d5900b-1045-4e1a-b4a0-f5e28d4f8972
E0330 16:17:15.056766       1 utils.go:252] GetAlertingRuleStatus resourcename prometheus-k8s-rules

```
```

```release-note
None
```

```docs

```

### DIFF
--- a/pkg/models/alerting/rules/utils.go
+++ b/pkg/models/alerting/rules/utils.go
@@ -173,8 +173,13 @@ func GetAlertingRulesStatus(ruleNamespace string, ruleChunk *ResourceRuleChunk, 
 		if !strings.HasPrefix(fileShort, ruleNamespace+"-") {
 			continue
 		}
-		resourceRules, ok := ruleChunk.ResourceRulesMap[strings.TrimPrefix(fileShort, ruleNamespace+"-")]
-		if !ok {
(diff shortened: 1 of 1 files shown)


junotx:
@chuan-you Thanks for your fix.
The prometheus rule files have been with a uid postfix since https://github.com/prometheus-operator/prometheus-operator/pull/4347.
Maybe you use fileshort prefix matching instead of contains to be more accurate

junotx:
@chuan-you Thanks for your fix. 
The contains maybe not accurate because resource name may have a same substring.

The prometheus rule files have been with the PrometheusRule resource UID postfix since prometheus-operator/prometheus-operator#4347. Maybe you compare `fileShort` with `<ruleNamespace>-<ruleResourceName>-<ruleResourceUID>` to be more accurate.

Issue #5602, opened 2023-03-22
open api 'upgrade application' 中的“{application}”这个path variable的取值有问题


jieqian:
**Describe the Bug**
pkg/kapis/openpitrix/v1/register.go
![image](https://user-images.githubusercontent.com/4167422/226794406-8c9fbe6e-e3dc-433c-8ab2-50548e823dc9.png)
pkg/kapis/openpitrix/v1/handler.go
![image](https://user-images.githubusercontent.com/4167422/226794499-ab00bdaf-b96d-4276-acbd-f2cd4e0387f2.png)
applicationId不应该赋值给ClusterId，导致用户传的path variable是正确的app_id，但一直报错说“helmrelease.application.kubesphere.io ${YOUR_APP_ID} not found”

For UI issues please also add a screenshot that shows the issue.

**Versions Used**
KubeSphere: release-3.1
Kubernetes: (If KubeSphere installer used, you can skip this)

**Environment**
How many nodes and their hardware configuration:

For example: CentOS 7.5 / 3 masters:  8cpu/8g; 3 nodes: 8cpu/16g
(and other info are welcomed to help us debugging)

**How To Reproduce**

**Expected behavior**

zhou1203:
是的，这确实是个bug，如果你愿意的话可以修复一下提一个PR给我们

wtw1994:
@sologgfun bug是applicationId不应该赋值给ClusterId，修改此处即可， applicationId符合appstore应用商店命名规范

sologgfun:
This issue has been fix, can be closed.

PR #5666, opened 2023-05-05
Fix wrong params naming in openpitrix

sologgfun:
### What type of PR is this?
Add one of the following kinds:
/kind bug

Optionally add one or more of the following kinds if applicable:
/kind api-change

In the following issue, I found the applicationId in openpitrix api is misunderstanding.
https://github.com/kubesphere/kubesphere/issues/5602
In the API, the "applicationId" parameter is actually the name of the helmRelease, but it uses "applicationId" as the parameter name, as shown in the following figure.

image

Fixes # https://github.com/kubesphere/kubesphere/issues/5602

This PR will cause changes to the OpenAPI, so we need to be cautious. Additionally, the commit only made changes to the upgrade API, while in reality, the get API and others also need modification. I would like to confirm the opinions of the members. Is it necessary to make these changes?

```release-note

```

```docs

```

### DIFF
--- a/pkg/kapis/openpitrix/v1/handler.go
+++ b/pkg/kapis/openpitrix/v1/handler.go
@@ -767,13 +767,12 @@ func (h *openpitrixHandler) UpgradeApplication(req *restful.Request, resp *restf
 	}
 
 	upgradeClusterRequest.Namespace = namespace
-	upgradeClusterRequest.ClusterId = applicationId
 	user, _ := request.UserFrom(req.Request.Context())
 	if user != nil {
 		upgradeClusterRequest.Username = user.GetName()
 	}
 
-	err = h.openpitrix.UpgradeApplication(upgradeClusterRequest)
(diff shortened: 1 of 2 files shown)


pkg/kapis/openpitrix/v1/handler.go
wtw1994, line 1: fix bug using   upgradeClusterRequest.AppId = applicationId  , please do not modify the naming convention for the appstore API

sologgfun:
AppId already exist in body, and it`s different from the params in path.They have 
So `upgradeClusterRequest.AppId = applicationId` maybe is not suitable.

image

These two parameters have different meanings. Clearly, the AppId in the body refers to the AppId in the helmRelease.

image

At the same time, in order to avoid modifying the naming convention for the appstore API, I submitted a new commit.
Can you check if it complies with the design principles? @wtw1994

wtw1994 (approved):

PR #5678, opened 2023-05-11
docs: modified section about style shields display issues in readme-zh

cubxxw:
### What type of PR is this?
/kind documentation

Modification part about style shields display problem, reasonable modification


Bettygogo2021:
@cubxxw Hi, thx for your contribution. However, when i checked the file, I found that one more logo had been added. Is it expected?
image

PR #6244, opened 2024-10-29
Refactor FormatVersion function to handle invalid characters

inksnw:

Fixes #https://github.com/kubesphere/kubesphere/issues/6243

```
```

```release-note
None
```

```docs

```

### DIFF
--- a/pkg/simple/client/application/helper.go
+++ b/pkg/simple/client/application/helper.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/md5"
+	"crypto/sha1"
 	"crypto/tls"
 	"encoding/hex"
 	"encoding/json"
@@ -13,7 +14,6 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/simple/client/application/helper.go
```diff
@@ -571,14 +571,16 @@ func GenerateShortNameMD5Hash(input string) string {
...
-		input = re.ReplaceAllStringFunc(input, func(s string) string {
-			return "-"
-		})
-	}
-	return input
+	if len(validation.IsDNS1123Subdomain(input)) == 0 {
+		return input
+	}
+	invalidCharsStr := strings.Join(re.FindAllString(input, -1), "")
+	hash := md5.Sum([]byte(invalidCharsStr))
```
wansir, line 578: Although security is usually not the primary concern when generating Kubernetes resource names, MD5 has known weaknesses and is susceptible to collision attacks (i.e., generating the same MD5 hash for different inputs). SHA-1 and SHA-256 are more collision-resistant, which is why MD5 is generally not recommended in most modern application scenarios.

pkg/simple/client/application/helper.go
```diff
@@ -571,14 +571,16 @@ func GenerateShortNameMD5Hash(input string) string {
...
+	md5Str := hex.EncodeToString(hash[:])[:8]
+	validPart := re.ReplaceAllString(input, "")
```
wansir, line 580: If the version number contains special characters, the entire version number should be hashed, not just the special characters.

PR #6308, opened 2024-12-09
feat: Enhance JSBundle Api to declare auxiliary asset files required for extending components

lingbohome:
### What type of PR is this?

/kind feature

After the extended component is packaged, some auxiliary style files and assets files may be generated.
These files are usually generated by the following two usage scenarios:
1. The third-party libraries that the extension component depends on need to import additional css and js.
2. The extension component itself directly imports css files and other static asset files.
Therefore, we need to enhance the JSBundle Api to support this complex usage scenario
Like this:
api

Fixes #

```
```

```release-note
 add support for auxiliary asset files to jsBundle
```

```docs

```


wansir:
@lingbohome  Thanks for your contribution!

The JSBundle is used exclusively for loading extension component code, and ks-console will watch this resource in real time. It differs from ReverseProxy in functionality.

You can configure static resource proxying using ReverseProxy, which offers advanced features. Learn more at: [ReverseProxy Documentation](https://dev-guide.kubesphere.io/extension-dev-guide/en/feature-customization/extending-api/#reverseproxy).

In my opinion, a better approach would be to enable ReverseProxy to support `/proxy/*` and `/dist/*` as path prefixes.

wansir:
> > @lingbohome Thanks for your contribution!感谢您的贡献！
> > The JSBundle is used exclusively for loading extension component code, and ks-console will watch this resource in real time. It differs from ReverseProxy in functionality.JSBundle 专门用于加载扩展组件代码，ks-console 会实时监视该资源。它在功能上与 ReverseProxy 不同。
> > You can configure static resource proxying using ReverseProxy, which offers advanced features. Learn more at: [ReverseProxy Documentation](https://dev-guide.kubesphere.io/extension-dev-guide/en/feature-customization/extending-api/#reverseproxy).您可以使用 ReverseProxy 配置静态资源代理，它提供高级功能。了解更多信息： [ReverseProxy 文档](https://dev-guide.kubesphere.io/extension-dev-guide/en/feature-customization/extending-api/#reverseproxy)。
> > In my opinion, a better approach would be to enable ReverseProxy to support `/proxy/*` and `/dist/*` as path prefixes.在我看来，更好的方法是启用 ReverseProxy 支持`/proxy/*`和`/dist/*`作为路径前缀。
> 
> I have tried this solution before, but it does not seem to support CSS loading well, as well as scenarios that are sensitive to MIMETYPE of some static files.我之前尝试过这个解决方案，但似乎不能很好地支持CSS加载，以及对某些静态文件的MIMETYPE敏感的场景。

The `Content-Type` is typically provided by the upstream, and the ReverseProxy supports specific headerDown configurations, which may help resolve this issue.

wansir:
> > @lingbohome Thanks for your contribution!
> > The JSBundle is used exclusively for loading extension component code, and ks-console will watch this resource in real time. It differs from ReverseProxy in functionality.
> > You can configure static resource proxying using ReverseProxy, which offers advanced features. Learn more at: [ReverseProxy Documentation](https://dev-guide.kubesphere.io/extension-dev-guide/en/feature-customization/extending-api/#reverseproxy).
> > In my opinion, a better approach would be to enable ReverseProxy to support `/proxy/*` and `/dist/*` as path prefixes.
> 
> I have tried this solution before, but it does not seem to support CSS loading well, as well as scenarios that are sensitive to MIMETYPE of some static files.

The `Content-Type` is typically provided by the upstream, and the ReverseProxy supports specific headerDown configurations, which may help resolve this issue.

wansir:
> > > @lingbohome Thanks for your contribution!
> > > The JSBundle is used exclusively for loading extension component code, and ks-console will watch this resource in real time. It differs from ReverseProxy in functionality.
> > > You can configure static resource proxying using ReverseProxy, which offers advanced features. Learn more at: [ReverseProxy Documentation](https://dev-guide.kubesphere.io/extension-dev-guide/en/feature-customization/extending-api/#reverseproxy).
> > > In my opinion, a better approach would be to enable ReverseProxy to support `/proxy/*` and `/dist/*` as path prefixes.
> > 
> > 
> > I have tried this solution before, but it does not seem to support CSS loading well, as well as scenarios that are sensitive to MIMETYPE of some static files.
> 
> In addition, it may be better to use ReverseProxy for static files that the extension component itself can freely control, because it can be wildcarded. However, some dependent third-party libraries cannot well control the static resources they depend on. They may be loaded from the /pstatic prefix.

We need to resolve the reverse proxy issues through ReverseProxy, rather than relying on the JSBundle. The /pstatic path is specifically for loading JSBundles. Other static resources should be placed under paths like `/proxy/`, `/dist`, or any other planned public paths. Using stripPathPrefix can help alleviate issues caused by hardcoded prefixes.

wansir:
> > > > > @lingbohome Thanks for your contribution!
> > > > > The JSBundle is used exclusively for loading extension component code, and ks-console will watch this resource in real time. It differs from ReverseProxy in functionality.
> > > > > You can configure static resource proxying using ReverseProxy, which offers advanced features. Learn more at: [ReverseProxy Documentation](https://dev-guide.kubesphere.io/extension-dev-guide/en/feature-customization/extending-api/#reverseproxy).
> > > > > In my opinion, a better approach would be to enable ReverseProxy to support `/proxy/*` and `/dist/*` as path prefixes.
> > > > 
> > > > 
> > > > I have tried this solution before, but it does not seem to support CSS loading well, as well as scenarios that are sensitive to MIMETYPE of some static files.
> > > 
> > > 
> > > In addition, it may be better to use ReverseProxy for static files that the extension component itself can freely control, because it can be wildcarded. However, some dependent third-party libraries cannot well control the static resources they depend on. They may be loaded from the /pstatic prefix.
> > 
> > 
> > We need to resolve the reverse proxy issues through ReverseProxy, rather than relying on the JSBundle. The /pstatic path is specifically for loading JSBundles. Other static resources should be placed under paths like `/proxy/`, `/dist`, or any other planned public paths. Using stripPathPrefix can help alleviate issues caused by hardcoded prefixes.
> 
> First of all, put aside whether ReverseProxy can solve this scenario, Based on my current understanding of JSBundle Api, let me talk about my opinion. Since it is named js bundle, from the user's perspective, it should be cohesive enough and it should contain all the JavaScript code of the application. , dependent libraries, static resources, etc., and some may be an integral part, rather than forcibly dividing them with other solutions to solve the problem. Unless the JSBundle api itself has a better implementation solution, everything can be packaged into an index and js file, so that there is no need to declare other resources that the application depends on in the JSBundle.

I’m with you on this. However, we need a more comprehensive design to implement this, which requires collaboration from the front-end team, including path adjustments and proper planning.

PR #6370, opened 2025-02-19
fix: pod list with owner filter logic

smartcat999:

Fixes #

```
```

```release-note
Fixed the issue that deployments and statefulsets use the same label, which causes abnormal pod list data.
```

```docs

```


donniean:
https://github.com/kubesphere/kse-console/issues/6131

